PR #6, opened 2015-07-29
Moved all viewport adapters to subfolder

WardBenjamin:
This pull request moves all viewport adapters to their own subfolder inside of `Graphics` to improve organization a bit. 

I was getting a bit bothered by the folder structure :wink:


craftworkgames:
2 things:

1. I don't like having spaces in folder names because..
2. I think the folder structure should match the namespaces.

However, I agree that it's annoying having the viewport adapters mixed in with the other classes. It's definitely more organized having them in a separate folder. The question is, do we also want to change the namespace? 

I've given this some thought before and these are the candidates in my head:

`MonoGame.Extended.Graphics.ViewportAdapters`
`MonoGame.Extended.ViewportAdapters`

What I don't like about both of these options is that viewport adapters are closely related to the camera class. It seems odd to have them in separate namespaces. Not sure.

Perhaps I'm just being too strict on the namespace = folder thing. If MonoGame.Extended where being used by lots of people I certainly wouldn't want to change the namespace (breaking change) so maybe I should just relax the rules a bit.

I guess we don't need to figure all of these things out right away. Take the spaces out of the folder names and I'll merge the request.

WardBenjamin:
Sounds good!

WardBenjamin:
Huh, look at that. For some reason I merged master into this branch. Whatever, I guess.

WardBenjamin:
@craftworkgames after some testing this should be ready to merge

craftworkgames:
@WardBenjamin Sorry I've had a really busy weekend. I've barely had time to sit down in front of a computer. I'll merge it now.

craftworkgames:
Not sure what happened but the merge [broke the build](http://build.craftworkgames.com/viewType.html?buildTypeId=MonoGameExtended_CI&guest=1). I've fixed it up now.

WardBenjamin:
Huh, that's really odd.... I wonder why that happened. Don't worry about it though, I haven't had much time either (and I soon leave for my last vacation of the year, so.... yeah :wink:).

PR #26, opened 2015-08-06
Removing MG nuget package, adding updated PCL

WardBenjamin:
This pull request updates the MonoGame PCL that the MonoGame.Extended PCL compiles against to version 3.4.


craftworkgames:
Okay. So with this PR we will be temporarily using our own DLL. 

This will work for a while, but we'll need another issue raised to go back to a NuGet package in the future.

Issue #27, opened 2015-08-06
Platform-based build configurations


WardBenjamin:
Hey Dylan,

In some of my projects, I have created different build configurations for each platform. This means that multiple versions of MonoGame.Framework are referenced via Nuget, and only the platform-specific one that is required by the build configuration is actually used.

Would you be interested in this for the sandbox project? It's not too much work for me, and it would save me hacking together a new csproj file and editing references every time I pull.

craftworkgames:
I think it's a very good thing to have everything work out of the box for everyone. I don't want you to be hacking together csproj files and editing references all the time. That sux.

On the other hand, I'm not entirely convinced that having multiple people working in the sandbox project is going to scale very well. At this stage I'm not sure how to solve that problem. We definitely need a way to make changes to the library and test them in some sort of "game" like project. Maybe that's a sandbox, maybe it isn't.

> I have created different build configurations for each platform.

Do you have a repo I can look at to understand how this works? It sounds good on paper but I'm not sure what to expect from the PR.

If it makes the project better and won't break anything, by all means.

I'm

WardBenjamin:
I don't have a repo that you can look at, since the project isn't open-source, but I can paste some info in here. Here's what the references inside of `foobar.csproj` might look like:

    <ItemGroup>
      <Reference Include="System" />
      <Reference Include="System.Data" />
      <Reference Include="System.Data.DataSetExtensions" />
      <Reference Include="System.Xml" />
      <Reference Include="System.Core" />
      <Reference Include="Tao.Sdl" Condition="'$(Configuration)'=='LinuxDebug' Or '$(Configuration)'=='LinuxRelease'">
        <HintPath>..\packages\MonoGame.Framework.Linux.3.4.0.459\lib\net40\Tao.Sdl.dll</HintPath>
      </Reference>
      <Reference Include="MonoGame.Framework" Condition="'$(Configuration)'=='WinDebug' Or '$(Configuration)'=='WinRelease'">
        <HintPath>..\packages\MonoGame.Framework.WindowsGL.3.4.0.459\lib\net40\MonoGame.Framework.dll</HintPath>
      </Reference>
      <Reference Include="MonoGame.Framework" Condition="'$(Configuration)'=='LinuxDebug' Or '$(Configuration)'=='LinuxRelease'">
        <HintPath>..\packages\MonoGame.Framework.Linux.3.4.0.459\lib\net40\MonoGame.Framework.dll</HintPath>
[...]

PR #29, opened 2015-08-14
Platform-based build configurations

WardBenjamin:
Alright! Here we go!

This should fulfill https://github.com/craftworkgames/MonoGame.Extended/issues/27, adding the following configurations:

`WindowsDebug, WindowsRelease, MacDebug, MacRelease, LinuxDebug, LinuxRelease`

These are all compiled towards `x86`.

I'd appreciate testing of this more than I have done, it seems to work fine on my systems but I want a double-check.


craftworkgames:
It's completely messed up in Visual Studio 2015 on Windows 10. I'm going to roll back the merge until we come up with a solution.

PR #30, opened 2015-08-14
Revert "Platform-based build configurations"

craftworkgames:
Reverts craftworkgames/MonoGame.Extended#29


WardBenjamin:
Hmm yeah this is odd. I'm not yet running Windows 10, but it seemed to work fine in VS 2013 Community. I'll investigate a bit if I can.

craftworkgames:
I doubt Windows 10 is the issue. VS2015 might be though.

WardBenjamin:
Yeah, that's what I was thinking. Sorry for the ambiguity.

craftworkgames:
If we can't get this approach to work it might be worth considering alternatives. Another option might be to have 2 Sandbox projects (Sandbox.Windows and Sandbox.Linux) that share code between them.

Issue #43, opened 2015-08-31
Can I bring my Circle over?


killerrin:
Awhile ago I wrote a Pull Request to get a circle into MonoGame, though I think general concensus is that it shouldn't be in the core library. 

That said, since it looks like we are trying to create an official Monogame Extension Library, I was wondering if this would be something we would be interested in adding to the Extended Library?

Heres the Pull Request where you can see the files
https://github.com/mono/MonoGame/pull/3419

WardBenjamin:
Tentatively, I feel that this is some very good work and it should be included (perhaps in MonoGame.Extended.Utilities? I don't know). CC @craftworkgames

craftworkgames:
Sorry guys, I've been extra busy today.

> I was wondering if this would be something we would be interested in adding to the Extended Library?

Yes, I think this could be a good addition to the library.

> perhaps in MonoGame.Extended.Utilities?

I'd rather use more descriptive namespaces like `MonoGame.Extended.Shapes` or `MonoGame.Extended.Geometry` perhaps.

Pretty much everything in the PR looks really good. We just need to be careful to keep the API as  consistent as possible. For example, in our Sprite class there's a method called  [`GetBoundingRectangle`](https://github.com/craftworkgames/MonoGame.Extended/blob/master/Source/MonoGame.Extended/Sprites/Sprite.cs#L60) and in the Circle class there's a method called [`ToRectangle`](https://github.com/mono/MonoGame/pull/3419/files#diff-c5dede4989ab48ad181f4a16c05f6ae7R473). It's only a subtle difference, but I'd like to try an keep things like this consistent across the API. Which do you guys prefer?

If we are going to introduce a `MonoGame.Extended.Shapes` namespace it would make sense to include other kinds of shapes. Triangle, Polygon and Line maybe. I noticed some of this has already been discussed in the [MonoGame PR discussion](https://github.com/mono/MonoGame/pull/3419#issuecomment-70435158)

Lastly, I think it's important to take a hard line on what we are including and what we are not. @KonajuGames put it nicely when he said:

> It can be a slippery slope.  Personally I think MonoGame is not a collision
detection library, and that is best handled by more dedicated libraries
such as Farseer and Bepu.  That said, Circle is a simple primitive and is
often used in larger games for 2D culling purposes.

[...]

PR #46, opened 2015-09-04
Added Circle

killerrin:
Added the circle as discussed in https://github.com/craftworkgames/MonoGame.Extended/issues/43

### DIFF
--- a/Source/MonoGame.Extended/Shapes/Circle.cs
+++ b/Source/MonoGame.Extended/Shapes/Circle.cs
@@ -1,0 +1,506 @@
+﻿using Microsoft.Xna.Framework;
+using System;
+using System.Diagnostics;
+using System.Runtime.Serialization;
+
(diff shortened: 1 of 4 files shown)


Source/MonoGame.Extended/Shapes/Circle.cs
```diff
@@ -0,0 +1,506 @@
...
+
+			if (distance.X > (value.Width / 2.0f + Radius))
+				result = false;
+			if (distance.Y > (value.Height / 2.0f + Radius))
+				result = false;
+
+			if (distance.X <= (value.Width / 2.0f))
+				result = true;
+			if (distance.Y <= (value.Height / 2.0f))
+				result = true;
```
craftworkgames: These `result = true` lines look like bugs? The result *always* gets overridden at the end of the method so the if statements are not doing anything useful.

Issue #15, opened 2015-08-03
Sprite animation using sprite sheets


craftworkgames:
We've all seen 2D animation using the [frames of a sprite sheet](https://github.com/libgdx/libgdx/wiki/2D-Animation). 

This is reasonably easy to implement but we want to design a clean and consistent API that works well with other types of animation (e.g. skeletal and transitions). Also, be careful not to leak the animation code into the base `Sprite` classes.

cra0zy:
I'll be adding an Update method to Sprite class for setting animation speed and current frame.

craftworkgames:
Hey @cra0zy 

Thanks for taking on this job. I've been thinking about it for a while but haven't started anything.

> I'll be adding an Update method to Sprite class for setting animation speed and current frame.

Please don't modify the `Sprite` class. I know it sounds silly, but I'd really rather keep the Sprite class as simple as possible and use it as a building block for other functionality. There's 2 ways to achieve this:

### Option 1: Create a derived class called `AnimatedSprite`

Doing it this way would allow you to add an `Update` method to the `AnimatedSprite` class and keep the animation code isolated from the `Sprite` class. Separation of concerns.

### Option 2: Create a `SpriteAnimator` class

An alternative approach is to create an animator class that takes a `Sprite` and a `TextureAtlas` in it's constructor as an animation target. This is similar to [how LibGDX does it](https://github.com/libgdx/libgdx/wiki/2D-Animation).

The advantage of this approach is that we could create different types of animators and isolate the code in a MonoGame.Extended.Animations namespace.

My current preference is option 2. Let's discuss.

PR #48, opened 2015-09-11
Added Animation class

cra0zy:
#15

### DIFF
--- a/Source/MonoGame.Extended/MonoGame.Extended.csproj
+++ b/Source/MonoGame.Extended/MonoGame.Extended.csproj
@@ -82,6 +82,8 @@
     <Compile Include="ViewportAdapters\DefaultViewportAdapter.cs" />
     <Compile Include="ViewportAdapters\ScalingViewportAdapter.cs" />
     <Compile Include="ViewportAdapters\ViewportAdapter.cs" />
+    <Compile Include="Sprites\Animation.cs" />
+    <Compile Include="Sprites\AnimationExtension.cs" />
   </ItemGroup>
   <ItemGroup>
     <Reference Include="Newtonsoft.Json, Version=7.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed">
--- a/Source/MonoGame.Extended/Sprites/Animation.cs
+++ b/Source/MonoGame.Extended/Sprites/Animation.cs
@@ -113,7 +113,7 @@ namespace MonoGame.Extended.Sprites
                         }
                     }
 
-                    _timeElapsed = 0;
+                    _timeElapsed -= Speed;
                 }
             }
         }
--- a/Source/MonoGame.Extended/Sprites/Animation.cs
+++ b/Source/MonoGame.Extended/Sprites/Animation.cs
@@ -1,0 +1,142 @@
+﻿using System;
+using System.Collections.Generic;
+
+using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
+
+using MonoGame.Extended.TextureAtlases;
+
+namespace MonoGame.Extended.Sprites
+{
+    public class Animation
+    {
+        private void Initalize()
+        {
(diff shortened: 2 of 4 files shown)


cra0zy:
@craftworkgames I changed the default origin of Sprite to 0, 0 because everything in XNA has that as it's default origin.

craftworkgames:
> I changed the default origin of Sprite to 0, 0 because XNA uses that as default origin.
It would be confusing if some classes were using 0, 0 as default, and others were centering.

Which classes specifically? I have a list of the classes [here](http://www.dylanwilson.net/list-of-xna-classes-grouped-by-dll) for reference.

Personally I still think it should default to the center because I think that's a more useful value when rotating the sprite. Rotating around the top left (or any other corner) is rarely useful. However, I'm always open to change it if there's a good reason.

I posted [the question on twitter](https://twitter.com/craftworkgames/status/642883019171172352) to see if I could get some outside opinions. I got two different opinions so far:

> Dylan Wilson ‏@craftworkgames
Opinion poll: should a Sprite class have a default Origin of top left or centre? #monogame #gamedev

> Scott Sebelius ‏@scottsebelius
@craftworkgames the obvious leaning for a sprite would be center since you think of center body as local and world position
> @scottsebelius that was my thinking too. I'm hoping to get a few more opinions on the matter.

> Patrick Weigel ‏@PatrickW462
@craftworkgames Bottom right. So we can finally correctly animate someone walking.
> @PatrickW462 wouldn't that be bottom middle?
> @craftworkgames More like wherever the foot is, so that animations can keep that point fixed. Bottom middle is a good place.
> @PatrickW462 seems kind of arbitrary but I guess so is centre and top left.
> @PatrickW462 rotation around the feet doesn't make sense though.


Changes like this can seem tiny on the surface but if they are not done carefully they can be far reaching and create frustrating bugs. I'd really prefer that we get into a habit of discussing them before making them please.

craftworkgames:
Btw.. now some tests are failing.
http://build.craftworkgames.com/viewLog.html?buildId=379&tab=buildResultsDiv&buildTypeId=MonoGameExtended_CI

cra0zy:
I know why the tests are failing I'll fix it a bit later today, I didn't know you've setup a teambuild server, nice.

As for what classes that use top left, spritebatch and texture2d, spritebatch and spritefont. I do understand wanting to center sprites(considering a lot of game engines do it), but for XNA you usually get used to thinking of 0, 0 as default origin for everything.

If we were building our own game engine I would go for centering, but since we are doing an extension on XNA, it makes a lot more sense to use 0, 0.

craftworkgames:
> spritebatch and texture2d, spritebatch and spritefont.

So basically the *only* class in XNA that has anything to do with the origin is SpriteBatch. Yes, it is [documented](https://msdn.microsoft.com/en-us/library/ff433992.aspx) to have a default origin of (0, 0) but I hardly think that's a good reason to follow suit.

> I do understand wanting to center sprites(considering a lot of game engines do it),

Yes, a lot of game engines do it because it makes sense. Our goal is to make it easier to make games with MonoGame. Setting the default origin to the top left corner is rarely useful when making games and I think having the center as the default is adding value to the Sprite class. In the few cases when it does make sense to use the top left corner (e.g. rendering a tile based map) you'll probably not be using the Sprite class anyway.

> but for XNA you usually get used to thinking of 0, 0 as default origin for everything.

Even if you *think* the default origin is 0, 0 for the Sprite class the first time you use it, you'll not likely want to leave it that way. The *most* likely scenario is that you'll want to set the origin to something other than the top left corner in almost all cases. It just doesn't make any sense.

cra0zy:
@WardBenjamin What do you think?

WardBenjamin:
@cra0zy Personally, I'd prefer the center of the sprite for the default origin since most often (always?) that's how I've used it. I mean, yes, nothing in the MonoGame API uses it, but our goal is to make this more intuitive and useful than the MG API alone, and I think that's a part of it.

Issue #40, opened 2015-08-27
Frames Per Second Counter


craftworkgames:
I while ago I wrote an FPS counter and posted as [an answer to a question on stack overflow](http://stackoverflow.com/questions/20676185/xna-monogame-getting-the-frames-per-second). It could be a good addition to the library.

craftworkgames:
This is done. Thanks @cra0zy

PR #49, opened 2015-09-11
Added FPS Counter

cra0zy:
#40

### DIFF
--- a/Source/MonoGame.Extended/FPSCounter.cs
+++ b/Source/MonoGame.Extended/FPSCounter.cs
@@ -1,0 +1,58 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+using Microsoft.Xna.Framework;
+
+namespace MonoGame.Extended
+{
+    public static class FPSCounter
+    {
+        #region Private Fields
+
+        private static int _maximumSamples = 100;
+        private static readonly Queue<float> _sampleBuffer = new Queue<float>();
+
+        #endregion
+
+        #region Public Fields
+
+        public static float AverageFramesPerSecond { get; private set; }
+        public static float CurrentFramesPerSecond { get; private set; } 
+
+        public static int MaximumSamples
+        {
+            get { return _maximumSamples; }
+            set { _maximumSamples = value; }
+        }
+
+        #endregion
+
+        #region Public Methods
+
+        public static void ResetCounter()
+        {
+            _sampleBuffer.Clear();
+        }
+
+        public static void Update(GameTime gameTime)
+        {
+            var deltaTime = (float)gameTime.ElapsedGameTime.TotalSeconds;
+            CurrentFramesPerSecond = 1.0f / deltaTime;
+
+            _sampleBuffer.Enqueue(CurrentFramesPerSecond);
+
+            if (_sampleBuffer.Count > MaximumSamples)
+            {
+                _sampleBuffer.Dequeue();
+                AverageFramesPerSecond = _sampleBuffer.Average(i => i);
+            } 
+            else
+            {
+                AverageFramesPerSecond = CurrentFramesPerSecond;
+            }
+        }
+
+        #endregion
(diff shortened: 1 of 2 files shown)


Source/MonoGame.Extended/FPSCounter.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+using Microsoft.Xna.Framework;
+
+namespace MonoGame.Extended
+{
+    public static class FPSCounter
+    {
+        #region Private Fields
```
craftworkgames: Please don't use regions. They just add to the clutter. If the class is small enough, there's absolutely no benefit. If the class is big, it probably should be broken up into smaller classes

Source/MonoGame.Extended/FPSCounter.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+using Microsoft.Xna.Framework;
+
+namespace MonoGame.Extended
+{
+    public static class FPSCounter
```
craftworkgames: Why is the class static? What if someone wants to have more than one instance of it?

craftworkgames:
I'm going to do the merge and fix some things it up. Thanks.

cra0zy:
Want me to setup a PR modifying stuff you requested?

craftworkgames:
Yeah. I haven't had a lot of time today.

Also, with the Animation PR, I couldn't see any actual changes to the Sprite and SpriteBatch extension aside from moving code around. Can we just leave as it is now.

Thanks.

cra0zy:
Currently this library is still young, but you want to setup some sort of code organization that all the classes will follow. If you don't do it now, you are gonna have big troubles later on when the code becomes messy.

And one more thing, yes I only reorganized the code in Sprite, I was sleepy when I wrote the commot message, I should have made it more specific, but just look at classes and tell me what looks nicer, https://github.com/craftworkgames/MonoGame.Extended/blob/master/Source/MonoGame.Extended/Sprites/Sprite.cs or https://github.com/cra0zy/MonoGame.Extended/blob/animatedsprite/Source/MonoGame.Extended/Sprites/Sprite.cs

I'll do what you asked for anyway.

craftworkgames:
> but just look at classes and tell me what looks nicer,

Obviously my opinion is biased and I'm going to pick the way I've organized the code over the way you've done it. I'm not saying there's anything wrong with your way, it's a perfectly reasonable way to do it.. but it's not consistent with the rest of the project.

The way I do it follows these rules generally:
 - constructors always go first
 - properties
 - properties with logic
 - methods
 - no regions
 - backing fields go directly above the property (e.g. _textureReigon)

The way you do it appears to be:
 - private fields including backing fields at the top
 - properties
 - constructors
 - methods
 - regions around everything

I totally respect that you have a clean and consistent coding style. If I'm ever making a PR on one of your repositories I promise to keep the style consistent with the rest of the project. I just ask that you try do the same with mine.

 > Currently this library is still young, but you want to setup some sort of code organization that all the classes will follow. If you don't do it now, you are gonna have big troubles later on when the code becomes messy.

I agree. A library should feel like it all fits together as one cohesive consistent whole. I'm always thinking about this stuff but maybe I've failed in a few areas. What do you think could be done to improve this situation?  Coding guidelines in the readme?  An outline of the project vision?  Other ideas?

cra0zy:
> The way I do it follows these rules generally:

Copy paste it somewhere in readme, I just didn't know the pattern.

> regions around everything

Funny thing, I actually very rarely use them :P

> Coding guidelines in the readme?

I think that would be a good idea, look at how MonoGame has done it: https://github.com/mono/MonoGame/wiki/Coding-Guidelines also I'll add a MonoDevelop coding policy file once you've decided on coding style.

Issue #18, opened 2015-08-03
Transition animations using a fluent syntax


craftworkgames:
A while ago I worked on a project that had a fluent syntax for short animations in code. They are not as powerful as Spine #16 or Spriter #17 but they are very handy for creating quick animations and effects.

They syntax looked something like this:

    var transitionParameters = new TransitionParameters(1.0f, EasingFunctions.CubicEaseInOut);
    Animations
        .CreateSequence(playButton)
        .MoveTo(new Vector2(50, 50), transitionParameters)
        .MoveTo(new Vector2(700, 400), transitionParameters)
        .MoveTo(new Vector2(700, 230), transitionParameters)
        .MoveTo(new Vector2(400, 250), transitionParameters)
        .Delay(1.0f)
        .RotateTo(0, transitionParameters)
        .Delay(1.0f)
        .RotateTo(MathHelper.TwoPi, transitionParameters)
        .FadeOut(transitionParameters)
        .FadeIn(transitionParameters)
        .Play();

Of course, it doesn't have to work exactly the same way. Hopefully this will be a good starting point though. It might be useful to consider adding Timelines so that animations like this can be created in parallel.

ClassicThunder:
I ran across this a while back and it is under the MIT license: [Glide by jacobalbano](https://bitbucket.org/jacobalbano/glide/src/4f1648dcf2cb428ebdd0bef1a43b08e37bc3ee6d?at=default)

craftworkgames:
I ran across a couple more related libraries for reference.

[XnaTweener](https://xnatweener.codeplex.com/)
[GoKit](https://github.com/prime31/GoKit/wiki/1.-GoKit-Code-Basics)

PR #51, opened 2015-09-15
Updated Animation stuff

cra0zy:
Stuff done:
 - removed initalizrs for fields that are gonna get those values regardless
 - added the ability to modify sprites in animation as well as to add / remove them
 - added animation position
 - added ability to read spritesheet up-down instead of just being able to read left-right
   - should I maybe allow reading right-left and bottom-up? never saw anyone use those but still...

https://youtu.be/hc4_hqe87UE

### DIFF
--- a/Source/MonoGame.Extended/Sprites/Animation.cs
+++ b/Source/MonoGame.Extended/Sprites/Animation.cs
@@ -10,47 +10,70 @@ namespace MonoGame.Extended.Sprites
 {
     public class Animation
     {
-        private void Initalize()
-        {
-            _sprites = new List<Sprite>();
-            _timeElapsed = 0;
-            _frameIndex = 0;
-            _speed = 0;
-            _playing = false;
-
-            Loop = true;
-        }
-
         public Animation(Sprite[] sprites)
         {
-            Initalize();
+            Loop = true;
+
+            _sprites = new List<Sprite>();
             _sprites.AddRange(sprites);
         }
 
         public Animation(Texture2D[] textures)
         {
-            Initalize();
+            Loop = true;
+
+            _sprites = new List<Sprite>();
             foreach (var t in textures)
                 _sprites.Add(new Sprite(t));
         }
 
-        public Animation(Texture2D texture, Vector2 startpoint, float framewidth, float frameheight, int rows, int cols, float rowborder, float colborder)
+        public Animation(Texture2D texture, Vector2 startpoint, float framewidth, float frameheight, int rows, int cols, float rowborder, float colborder, bool lefttop)
         {
-            Initalize();
-            for (float x = startpoint.X; x < startpoint.X + framewidth * cols; x += framewidth + colborder)
+            Loop = true;
+
+            _sprites = new List<Sprite>();
+
+            if (lefttop)
+            {
+                for (float x = startpoint.X; x < startpoint.X + framewidth * cols; x += framewidth + colborder)
+                    for (float y = startpoint.Y; y < startpoint.Y + frameheight * rows; y += frameheight + rowborder)
+                        _sprites.Add(new Sprite(new TextureRegion2D(texture, (int)x, (int)y, (int)framewidth, (int)frameheight)));
+            }
+            else
+            {
                 for (float y = startpoint.Y; y < startpoint.Y + frameheight * rows; y += frameheight + rowborder)
-                    _sprites.Add(new Sprite(new TextureRegion2D(texture, (int)x, (int)y, (int)framewidth, (int)frameheight)));
+                    for (float x = startpoint.X; x < startpoint.X + framewidth * cols; x += framewidth + colborder)
+                        _sprites.Add(new Sprite(new TextureRegion2D(texture, (int)x, (int)y, (int)framewidth, (int)frameheight)));
+            }
         }
 
-        internal List<Sprite> _sprites;
         private float _timeElapsed;
         public bool Loop { get; set; }
+
+        public Vector2 Position { get; set; }
 
         public Sprite CurrentSprite
         {
             get
             {
                 return _sprites[_frameIndex];
+            }
+        }
+
+        public int FrameCount
+        {
+            get
+            {
+                return _sprites.Count;
+            }
+        }
+
+        private List<Sprite> _sprites;
+        public List<Sprite> Sprites
+        {
+            get
+            {
+                return _sprites;
             }
         }
 
@@ -64,7 +87,7 @@ namespace MonoGame.Extended.Sprites
             set
             {
                 if (_speed < 0)
-                    throw new ArgumentOutOfRangeException("value", "AnimationSpeed has to be equal or higher than 0");
+                    throw new ArgumentOutOfRangeException("value", "Speed has to be equal or higher than 0.");
 
                 _speed = value;
             }
@@ -77,7 +100,7 @@ namespace MonoGame.Extended.Sprites
             set
             {
                 if (value < 0)
-                    throw new ArgumentOutOfRangeException("value", "FrameIndex has to be equal or higher than 0");
+                    throw new ArgumentOutOfRangeException("value", "FrameIndex has to be equal or higher than 0.");
 
                 if (value >= _sprites.Count)
                     throw new ArgumentOutOfRangeException("value", "FrameIndex cannot be higher than the ammount of frames that the sprite contains.");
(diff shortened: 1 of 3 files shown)


craftworkgames:
Looks cool. I like that you posted a youtube video.

Probably the single most important thing when we develop new features is the API to use it. I noticed in the background of the youtube video you've got some test code in the `Initialize` method. That's the code we should be talking about here on github. The hidden implementation details are not so important, but getting the API right is.

Also, once the feature is finished I'd like to have a sample in the [Samples repository](https://github.com/craftworkgames/Samples.Extended) and a blog post about it to make it complete. I don't mind doing this stuff, but it would be really helpful to get a starting point.

Please post any test code here on github (in the [original issue](https://github.com/craftworkgames/MonoGame.Extended/issues/15) preferably) for discussion.

cra0zy:
I just noticed that I've put the code into Initalize instead of LoadContent(in the video).... silly me....

Anyway I was planning on posting the code to the samples repo later, I just wanted this PR done first because of position.

craftworkgames:
> Anyway I was planning on posting the code to the samples repo later

We need to discuss the API before putting it into the samples. There's a number of things that need attention.

cra0zy:
> We need to discuss the API before putting it into the samples.

Discuss!!!

craftworkgames:
> Discuss!!!

Maybe I didn't make it clear.

I want *you* to post *the code* from your `AnimationTest` here in the *github discussion* please.

cra0zy:
```c#
#region Using Statements
using System;

using Microsoft.Xna.Framework.Graphics;
using Microsoft.Xna.Framework.Storage;
using Microsoft.Xna.Framework.Input;

using MonoGame.Extended;
using MonoGame.Extended.Sprites;

#endregion

namespace AnimationTest
{
    /// <summary>
    /// This is the main type for your game.
    /// </summary>
    public class Game1 : Game
    {
        GraphicsDeviceManager graphics;
        SpriteBatch spriteBatch;

        Animation s1, s2;
        SpriteFont font;

        public Game1()
        {
            graphics = new GraphicsDeviceManager(this);
            Content.RootDirectory = "Content";

            this.Window.Title = "Animation Test";
        }

        /// <summary>
        /// Allows the game to perform any initialization it needs to before starting to run.
        /// This is where it can query for any required services and load any non-graphic
        /// related content.  Calling base.Initialize will enumerate through any components
        /// and initialize them as well.
        /// </summary>
        protected override void Initialize()
        {
            // TODO: Add your initialization logic here
            base.Initialize();
        }

        /// <summary>
        /// LoadContent will be called once per game and is the place to load
        /// all of your content.
        /// </summary>
        protected override void LoadContent()
        {
            // Create a new SpriteBatch, which can be used to draw textures.
            spriteBatch = new SpriteBatch(GraphicsDevice);

            Texture2D t1 = Content.Load<Texture2D>("1");
            Texture2D t2 = Content.Load<Texture2D>("2");
            Texture2D t3 = Content.Load<Texture2D>("3");
            Texture2D t4 = Content.Load<Texture2D>("4");
            Texture2D t5 = Content.Load<Texture2D>("5");
            Texture2D t6 = Content.Load<Texture2D>("6");
            Texture2D ss = Content.Load<Texture2D>("spritesheet-fireball");

            font = Content.Load<SpriteFont>("arial");

            s1 = new Animation(new [] { t1, t2, t3, t4, t5, t6 });
            s1.Position = new Vector2(360, 100);
            s1.Speed = 100;
            s1.Play();

            s2 = new Animation(ss, Vector2.Zero, ss.Width / 3, ss.Height / 2, 2, 3, 0, 0, false);
            s2.Position = new Vector2(360, 300);
            s2.Speed = 100;
            s2.Play();
        }

        /// <summary>
        /// Allows the game to run logic such as updating the world,
        /// checking for collisions, gathering input, and playing audio.
        /// </summary>
        /// <param name="gameTime">Provides a snapshot of timing values.</param>
        protected override void Update(GameTime gameTime)
        {
            s1.Update(gameTime);
            s2.Update(gameTime);

            base.Update(gameTime);
        }

        /// <summary>
        /// This is called when the game should draw itself.
        /// </summary>
        /// <param name="gameTime">Provides a snapshot of timing values.</param>
        protected override void Draw(GameTime gameTime)
        {
            graphics.GraphicsDevice.Clear(Color.MidnightBlue);

            spriteBatch.Begin();
            spriteBatch.Draw(s1);
            spriteBatch.Draw(s2);
            spriteBatch.End();
            
            base.Draw(gameTime);
        }
    }
}


```

craftworkgames:
Awesome. So I'm going to do a code review to discuss some of my thoughts about the API. 

### The class name `Animation` is a little bit too generic. 

The animation you've created is only one type of animation we are going to support. The goal is to eventually have animations for [Spine](http://esotericsoftware.com/), [Spriter](http://www.brashmonkey.com/spriter.htm), [transition style animations](https://github.com/craftworkgames/MonoGame.Extended/issues/18) and possibly some others I haven't thought of yet. The point is, the classes need to be named more specifically.

### We shouldn't be encouraging people to load each animation frame as a separate texture
 
Code like this will not scale very well. Rendering sprites with a `SpriteBatch` is most efficient when sprites are packed into a single texture. Rendering 6 different textures for each animation is going to bring reasonable sized game to a halt fairly quickly.

    Texture2D t1 = Content.Load<Texture2D>("1");
    Texture2D t2 = Content.Load<Texture2D>("2");
    Texture2D t3 = Content.Load<Texture2D>("3");
    Texture2D t4 = Content.Load<Texture2D>("4");
    Texture2D t5 = Content.Load<Texture2D>("5");
    Texture2D t6 = Content.Load<Texture2D>("6");

### Loading a sprite sheet is the correct thing to do, but it's not the responsibility of the animation code

What you've done here is more efficient than loading individual textures, but the `Animation` class shouldn't be breaking the sprite sheet up into texture regions in it's constructor.

	Texture2D ss = Content.Load<Texture2D>("spritesheet-fireball");
	s2 = new Animation(ss, Vector2.Zero, ss.Width / 3, ss.Height / 2, 2, 3, 0, 0, false);

We already have a class in the library called [`TextureAtlas`](https://github.com/craftworkgames/MonoGame.Extended/blob/master/Source/MonoGame.Extended/TextureAtlases/TextureAtlas.cs) that's good for this type of thing. It would be better to do something like this:

	var texture = Content.Load<Texture2D>("spritesheet-fireball");
	var textureAtlas = TextureAtlas.Create(texture, Vector2.Zero, ss.Width / 3, ss.Height / 2, 2, 3, 0, 0, false);
	var s2 = new Animation(textureAtlas);

The parameters passed into the `TextureAtlas.Create` method is also an issue, but that's a discussion for another time.

### An animation *frame* should not be a Sprite

There's a bit of a disconnect between what goes into the constructor (textures) and what is stored internally in the animation class (sprites).  This issue also leaks out into the methods (`AddSprite`, `RemoveSprite`). 

What are we really animating here?  If you look at the code inside the constructor there's a hint..

	_sprites.Add(new Sprite(new TextureRegion2D(texture, (int)x, (int)y, (int)framewidth, (int)frameheight)));

There doesn't appear to be any benefit of storing a collection of sprites here. It would be perfectly reasonable to store a collection of texture regions instead:

	_regions.Add(new TextureRegion2D(texture, (int)x, (int)y, (int)framewidth, (int)frameheight));

Of course, this has implications on how the test of the class is implemented, but I think it make more sense this way.


### An animation is not something that you `Draw`

This might seem like a subtle difference at first but it could be a problem as we add different types of animations. Some animations might not be "drawable". 

You've created an extension method on the `SpriteBatch` for animations, which seems reasonable on the surface...

    spriteBatch.Begin();
    spriteBatch.Draw(s1);
    spriteBatch.Draw(s2);
    spriteBatch.End();

but looking at the implementation, it's rendering a sprite and overriding the position.

    public static void Draw(this SpriteBatch spriteBatch, Animation animation)
    {
        spriteBatch.Draw(animation.CurrentSprite, animation.Position);
    }

This means there's a disconnect between the `sprite.Position` and the `animation.Position`. They both have a position property but it's unclear (from the outside) which one is going to be used. 

### So what's the solution to all this?

To resolve the above issues I think it would be better if the `Animation` class became a `SpriteAnimator` class instead. The idea is that an animator has an `Update` method but does *not* have the ability to `Draw`. Instead, you pass in a `Sprite` as an animation target like this:


	var texture = Content.Load<Texture2D>("spritesheet-fireball");
	var textureAtlas = TextureAtlas.Create(texture, Vector2.Zero, ss.Width / 3, ss.Height / 2, 2, 3, 0, 0, false);
	
	_sprite = new Sprite(textureAtlas.Regions.First());		
	_animator = new SpriteAnimator(sprite, textureAtlas);

The update method would update the animator

    protected override void Update(GameTime gameTime)
    {
        _animator.Update(gameTime);

        base.Update(gameTime);
    }

And the draw method would draw the sprite

    protected override void Draw(GameTime gameTime)
    {
        graphics.GraphicsDevice.Clear(Color.MidnightBlue);

        spriteBatch.Begin();
        spriteBatch.Draw(_sprite);
        spriteBatch.End();

        base.Draw(gameTime);
    }

Internally a `SpriteAnimator` is simply changing the `TextureRegion` of the target sprite over time:

	_targetSprite.TextureRegion = _regions[_frameIndex];

The target sprite is not *owned* by the animator. It can move around and change its other properties independently of the animator. For example, if you wanted to move or rotate the sprite you still have control over those properties independently of the animation code.

This is how I would solve these particular issues, but I'm sure it's got flaws. If you disagree anything I've said here I welcome an open discussion. 

I just hope you realize now why I want to discuss these things before getting too deep into the implementation. There's lots of different ways to solve a problem and most of them cut right into the design of the API. If we don't sort this stuff out early it's going come back and bite later.

cra0zy:
You know, I started working on this, and just noticed you started adding some code for this as well so now merge conflicts!!!!!

talking about TextureAtlas Create method

craftworkgames:
> talking about TextureAtlas Create method

Sorry, my bad. I was just thinking about it a lot today after writing about it here. I added a few different things to the TextureAtlas class you should find useful. There's also lots of unit tests you can refer to for example usage.

I'm finished with my changes now (about to go to bed) so there won't be any more merge conflicts.

Issue #54, opened 2015-09-17
Timers


killerrin:
How would we feel about adding in a Timer Class? Its something that MonoGame doesn't have, and something that every Game Dev ends up having to create in all of their projects. 

I have two timers which I would be willing to add to the project, I made it a couple months ago for a MonoGame 'Game Engine' which I used in my Capstone Project which I think could fit this bill perfectly

Timer - https://github.com/KillerrinStudios/Anarian-Game-Engine-MonoGame/blob/master/Anarian%20Game%20Engine.Shared/DataStructures/Timer.cs
Clock - https://github.com/KillerrinStudios/Anarian-Game-Engine-MonoGame/blob/master/Anarian%20Game%20Engine.Shared/DataStructures/ContinuousClock.cs

The Timers themselves are event based, so once you set them up all you have to do is call the Update on them to update the time and once the Timer hits its goal (Timer counts down to 0 and Clock reports every x seconds), it fires off the Completed Event to let you know

craftworkgames:
I've used similar timers before in previous projects. They could be helpful.

I'd be hesitant about calling them `Timer` though. There's already `System.Timers.Timer` and `System.Threading.Timer` in the BCL. Another `Timer` class could get confusing.

As a side note: I also think there might be some bugs in the way MonoGame calculates frame times for the Update method. It's intended to run at 60 fps and always seems to return the same value, but I don't think they are accumulating the time properly. There's an [old artice](http://gafferongames.com/game-physics/fix-your-timestep/) about time-stepping that did the rounds a few years ago. Not sure how true it is though.

killerrin:
Alright, I'll submit a pull request

killerrin:
Alright, I'll submit a pull request

PR #58, opened 2015-09-18
Timers

killerrin:
https://github.com/craftworkgames/MonoGame.Extended/issues/54

Here are the timers

### DIFF
--- a/Source/MonoGame.Extended/Timers/ContinuousClock.cs
+++ b/Source/MonoGame.Extended/Timers/ContinuousClock.cs
@@ -1,0 +1,50 @@
+﻿using Microsoft.Xna.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MonoGame.Extended.Timers
+{
+    public class ContinuousClock : IGameComponent
+    {
+        public event EventHandler ReportTimePassed;
+
+        public TimerState TimerState;
+        public double ReportEveryXSeconds;
+
+        public TimeSpan CurrentTime { get; protected set; } = TimeSpan.Zero;
+        public TimeSpan NextReport { get; protected set; } = TimeSpan.Zero;
+
+        public ContinuousClock(double reportEveryXSeconds)
+        {
+            ReportEveryXSeconds = reportEveryXSeconds;
+            Reset();
+        }
+
+        void IGameComponent.Initialize() { Reset(); }
+        public void Reset()
+        {
+            TimerState = TimerState.Stopped;
+            CurrentTime = TimeSpan.Zero;
+            NextReport = CurrentTime + TimeSpan.FromSeconds(ReportEveryXSeconds);
+        }
+
+        public virtual void Update(GameTime gameTime)
+        {
+            if (TimerState == TimerState.Stopped ||
+                TimerState == TimerState.Paused)
+                return;
+
+            CurrentTime += gameTime.ElapsedGameTime;
+
+            if (CurrentTime >= NextReport)
+            {
+                NextReport = CurrentTime + TimeSpan.FromSeconds(ReportEveryXSeconds);
+
+                if (ReportTimePassed != null)
+                    ReportTimePassed(this, null);
+            }
+        }
+    }
+}
--- a/Source/MonoGame.Extended/Timers/CountdownTimer.cs
+++ b/Source/MonoGame.Extended/Timers/CountdownTimer.cs
@@ -1,0 +1,59 @@
+﻿using Microsoft.Xna.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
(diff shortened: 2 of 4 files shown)


Source/MonoGame.Extended/Timers/ContinuousClock.cs
```diff
@@ -0,0 +1,50 @@
+﻿using Microsoft.Xna.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MonoGame.Extended.Timers
+{
+    public class ContinuousClock : IGameComponent
```
craftworkgames: IGameComponent interface doesn't exist.

craftworkgames: What's the advantage of using IGameComponent?

Source/MonoGame.Extended/Timers/CountdownTimer.cs
```diff
@@ -0,0 +1,59 @@
...
+                TimerState == TimerState.Paused ||
+                TimerState == TimerState.Completed)
+                return;
+
+            CurrentTick += gameTime.ElapsedGameTime;
+
+            if (CurrentTick < Interval)
+            {
+                if (Tick != null)
+                    Tick(this, null);
```
craftworkgames: Probably best to get into a habit of firing events in a [thread safe way](http://stackoverflow.com/questions/3668953/raise-event-thread-safely-best-practice). It may never be needed, but it can't hurt.

craftworkgames:
Very nice PR. I really appreciate the effort you put in to make the code consistent with the MonoGame.Extended coding style.

Source/MonoGame.Extended/Timers/ContinuousClock.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        public event EventHandler ReportTimePassed;
+
+        public TimerState TimerState;
+        public double ReportEveryXSeconds;
+
+        public TimeSpan CurrentTime { get; protected set; } = TimeSpan.Zero;
+        public TimeSpan NextReport { get; protected set; } = TimeSpan.Zero;
```
craftworkgames: We are not using any C# 6.0 features in the project yet.

killerrin:
No problem, I looked at the changes you made in the previous Circle PR and decided to try and mimic that in this quick rewrite. Glad you like it :)

Source/MonoGame.Extended/Timers/ContinuousClock.cs
killerrin: Alright, wasn't sure about that, though decided to use it because I've been trying to get used to the new C# 6.0 stuf

Source/MonoGame.Extended/Timers/CountdownTimer.cs
killerrin: Didn't know that would be an issue cross thread, thanks for the read.

Source/MonoGame.Extended/Timers/ContinuousClock.cs
killerrin: I wouldn't really call it an advantage, just an alternative way of doing things within XNA/MonoGame.

From what I've read IGameComponent allows you to register the object as a component object with the Game class and as such will make the framework call the Update Method automatically on base.Update (http://stackoverflow.com/questions/15046181/what-is-base-updategametime-for-in-xna-games).

Some people prefer to do it that way, most though (I included) prefer to call Update Manually.

killerrin: That said, I just realized that in order to have IGameComponent work in that method, I would have also needed to implement IUpdateable. Right now all it would do is set it up on the Initialize.base.

So either I can Implement IUpdatable, or I can just go ahead and remove the IGameComponent

Source/MonoGame.Extended/Timers/ContinuousClock.cs
craftworkgames: I love C# 6.0 but I'm not sure how many people are using it yet. I'd rather keep the project compatible with VS2013 for a while until there's more adoption. If you're using [Resharper](https://www.jetbrains.com/resharper/) you can set the preferred language version on a project level. I realize not everyone's going to be using it though.

Source/MonoGame.Extended/Timers/ContinuousClock.cs
craftworkgames: >  I can just go ahead and remove the IGameComponent altogether.

I'd rather not use IGameComponent at all. At least until there's an obvious reason to do so. I'm a strong believer in keeping code as simple as possible. If we decide to support XNA game components we'll do it across the entire project as a separate issue.

killerrin: Sounds good

PR #70, opened 2015-10-05
Tilset importer fix for external .tsx tilesets

Ravenheart:
The content pipeline extension was throwing an exception when one tried to build a .tmx Tiled map file with a tileset that pointed to an external .tsx file. The fix checks for such tilesets, deserializes the external file and replaces the one in the main map instance.


craftworkgames:
@Ravenheart Thanks. I've had this one on the TODO list for a while. I had no idea it was only going to be a few more lines of code :smile:

PR #84, opened 2016-01-05
TiledMap background color now being rendered

jakethornton47:
Here are the changes! :)


craftworkgames:
Great work. Thanks @jakethornton47

Issue #81, opened 2015-12-12
Camera center


Sejsel:
I would love to have the ability to set the center of the camera.
This would allow simple implementation of a variable game resolution.

If the center was at the real center, increasing the game window would reveal content on all sides at the same rate, staying centered.

An example of what I mean I quickly put together (I apologize for poor quality):
http://gfycat.com/FaroffShrillDeviltasmanian

craftworkgames:
Good idea.

craftworkgames:
How would you imagine the API working for this?

Sejsel:
I have been thinking about this for some time and still haven't figured out what the best way to do this would be.

What seems most sensible for me right now is to have a Center Vector2 property which would default to (0, 0). (0.5, 0.5) would act like the example I have linked above. (1, 1) would expand from the other corner.

This would not even have to be limited to the range [0, 1], having the center off-camera would work just fine.

The only way I can imagine this working is by transforming the camera by Center*ViewportSize.

This might not be the best way to do this, I just wasn't able to come up with anything better.

PR #85, opened 2016-01-06
Draw method update for TiledLayer and TiledTileLayer

cupstudio:
I use <kbd>Ctrl + K, D</kbd> a lot out of habit, so it added a little formatting to a couple of things.

TiledTileLayer:
* The obsolete method `Draw(Camera2D camera)` no longer obsolete

TiledLayer:
* The obsolete method `Draw(Camera2D camera)` no longer obsolete.
* Implemented `Draw(Camera2D camera)` so that it only renders tiles within the camera's bounds.

### DIFF
--- a/Source/MonoGame.Extended/Maps/Tiled/TiledLayer.cs
+++ b/Source/MonoGame.Extended/Maps/Tiled/TiledLayer.cs
@@ -18,7 +18,6 @@ namespace MonoGame.Extended.Maps.Tiled
 
         public abstract void Draw();
 
-        [Obsolete("The camera is no longer required for drawing Tiled layers")]
         public abstract void Draw(Camera2D camera);
     }
 }
--- a/Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
+++ b/Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
@@ -13,7 +13,7 @@ namespace MonoGame.Extended.Maps.Tiled
         {
             Width = width;
             Height = height;
-            
+
             _map = map;
             _graphicsDevice = graphicsDevice;
             _spriteBatch = new SpriteBatch(graphicsDevice);
@@ -32,7 +32,7 @@ namespace MonoGame.Extended.Maps.Tiled
         private readonly GraphicsDevice _graphicsDevice;
         private readonly TiledTile[] _tiles;
         private readonly SpriteBatch _spriteBatch;
-        
+
         public IEnumerable<TiledTile> Tiles
         {
             get { return _tiles; }
@@ -82,10 +82,36 @@ namespace MonoGame.Extended.Maps.Tiled
             _spriteBatch.End();
         }
 
-        [Obsolete("The camera is no longer required for drawing Tiled layers")]
         public override void Draw(Camera2D camera)
         {
-            Draw();
+            var area = camera.GetBoundingRectangle();
+            int firstCol = (int)Math.Floor(camera.Position.X / _map.TileWidth);
+            int firstRow = (int)Math.Floor(camera.Position.Y / _map.TileHeight);
+
+            // +3 to cover any gaps
+            int numberOfColumns = (int)area.Width / _map.TileWidth + 3;
+            int numberOfRows = (int)area.Height / _map.TileHeight + 3;
+
+            _spriteBatch.Begin(transformMatrix: camera.GetViewMatrix(), blendState: BlendState.AlphaBlend, samplerState: SamplerState.PointClamp);
+            int index;
+            for (int row = firstRow; row <= firstRow + numberOfRows; row++)
+            {
+                index = row * _map.Width + firstCol;
+                for (int x = 0; x < numberOfColumns; x++)
+                {
+                    // index must be greater than 0 and less than _tiles.Length
+                    if (index >= 0 && index < _tiles.Length)
+                    {
+                        var region = _map.GetTileRegion(_tiles[index].Id);
+                        if (region != null)
+                        {
+                            RenderLayer(_map, _tiles[index], region);
+                        }
+                    }
+                    index++;
+                }
+            }
+            _spriteBatch.End();
         }
 
         private void RenderLayer(TiledMap map, TiledTile tile, TextureRegion2D region)
@@ -93,7 +119,7 @@ namespace MonoGame.Extended.Maps.Tiled
             switch (map.Orientation)
             {
                 case TiledMapOrientation.Orthogonal:
-                    RenderOrthogonal(tile,region);
+                    RenderOrthogonal(tile, region);
                     break;
                 case TiledMapOrientation.Isometric:
                     RenderIsometric(tile, region);
@@ -125,7 +151,7 @@ namespace MonoGame.Extended.Maps.Tiled
 
         public TiledTile GetTile(int x, int y)
         {
-            return _tiles[x + y*Width];
+            return _tiles[x + y * Width];
         }
 
         private Func<IEnumerable<TiledTile>> GetRenderOrderFunction()
--- a/Source/MonoGame.Extended/MonoGame.Extended.csproj
+++ b/Source/MonoGame.Extended/MonoGame.Extended.csproj
@@ -101,6 +101,7 @@
     <Compile Include="Vector2Extensions.cs" />
     <Compile Include="ViewportAdapters\BoxingViewportAdapter.cs" />
     <Compile Include="ViewportAdapters\DefaultViewportAdapter.cs" />
+    <Compile Include="ViewportAdapters\WindowViewportAdapter.cs" />
     <Compile Include="ViewportAdapters\ScalingViewportAdapter.cs" />
     <Compile Include="ViewportAdapters\ViewportAdapter.cs" />
     <Compile Include="FramesPerSecondCounter.cs" />
--- a/Source/MonoGame.Extended/ViewportAdapters/WindowViewportAdapter.cs
+++ b/Source/MonoGame.Extended/ViewportAdapters/WindowViewportAdapter.cs
@@ -1,9 +1,6 @@
 ﻿using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
 using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 
 namespace MonoGame.Extended.ViewportAdapters
 {
--- a/Source/MonoGame.Extended/ViewportAdapters/WindowViewportAdapter.cs
+++ b/Source/MonoGame.Extended/ViewportAdapters/WindowViewportAdapter.cs
@@ -1,0 +1,55 @@
+﻿using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MonoGame.Extended.ViewportAdapters
+{
+    public class WindowViewportAdapter : ViewportAdapter
+    {
+        public WindowViewportAdapter(GameWindow window, GraphicsDevice graphicsDevice)
+        {
+            GraphicsDevice = graphicsDevice;
+            Window = window;
+            window.ClientSizeChanged += OnClientSizeChanged;
+        }
+
+        protected readonly GameWindow Window;
+        protected GraphicsDevice GraphicsDevice;
+
+        public override int ViewportWidth
+        {
+            get { return Window.ClientBounds.Width; }
+        }
+        public override int ViewportHeight
+        {
+            get { return Window.ClientBounds.Height; }
+        }
+
+        public override int VirtualWidth
+        {
+            get { return Window.ClientBounds.Width; }
+        }
+
+        public override int VirtualHeight
+        {
+            get { return Window.ClientBounds.Height; }
+        }
+
+        public override Matrix GetScaleMatrix()
+        {
+            return Matrix.Identity;
+        }
+
+        private void OnClientSizeChanged(object sender, EventArgs eventArgs)
+        {
+            var x = Window.ClientBounds.Width;
+            var y = Window.ClientBounds.Height;
+
+            GraphicsDevice.Viewport = new Viewport(0, 0, x, y);
+        }
+
+    }
+}


craftworkgames:
Looks good for the most part.

What is the purpose of the `WindowViewportAdapter`?

I can't recall why I made the `Draw` with camera method obsolete. I'm slightly concerned something is broken, but overall I really appreciate your changes. I'll merge it shortly, when I get a little time to test it.

craftworkgames:
> What is the purpose of the WindowViewportAdapter?

Nevermind. I just noticed issue #86.

cupstudio:
> I'm slightly concerned something is broken

Let me know what you think - we can revert that method to Obsolete and try something else.  

One thing I've noticed, and I will look into this more tomorrow, is that near the left edge of the map I sometimes will see lines between tiles.  I've tried drawing with and without a render target and that doesn't seem to help.  I am using the `SamplerState.PointClamp` option, and the actual rendering is handled by the existing method being passed a tile, so I'm not sure what it is yet.  They don't always appear though, as you can see in the video I posted on #86, I hadn't even come across it yet at that point even after running it many times.

[Here is a screenshot](http://imgur.com/8kIoo4V)

craftworkgames:
>  I've tried drawing with and without a render target and that doesn't seem to help.

When you tried drawing with the render target did you remember to move the transformation matrix out of the first call to `spriteBatch.Begin` and into the second call?

It should look something like this:

    _graphicsDevice.SetRenderTarget(_renderTarget); 

    _spriteBatch.Begin(sortMode: SpriteSortMode.Deferred, samplerState: SamplerState.PointClamp);

    // render map here

    _spriteBatch.End();

    _graphicsDevice.SetRenderTarget(null);

    _spriteBatch.Begin(sortMode: SpriteSortMode.Immediate, blendState: BlendState.NonPremultiplied,
        samplerState: SamplerState.PointClamp, transformMatrix: CameraManager.pInstance.pFinalTransform);
    _spriteBatch.Draw(_renderTarget, Vector2.Zero, Color.White);
    _spriteBatch.End();

Coincidentally, I answered [a similar question on stack overflow](http://stackoverflow.com/questions/34626732/seeing-wrap-texture-when-using-clamp-mode-in-monogame-pictures-incl) this morning.

cupstudio:
Yes that appears to be it.  That means a small change to the `Draw` method in TiledTileLayer.  If you agree I will file another pull request (or if there is a different method for changing an already filed pull request let me know).  Thanks for being patient with me, hopefully this will pay off for some point as I intend to continue making contributions to the community over the long term as well.  :smile: 

Originally:

    spriteBatch.Begin(transformMatrix: camera.GetViewMatrix(), blendState:     BlendState.AlphaBlend, samplerState
    : SamplerState.PointClamp);


Changed to this so that the user can set a _renderTarget.

    _spriteBatch.Begin(sortMode: SpriteSortMode.Deferred, samplerState:     SamplerState.PointClamp);


In my test game's code I have this:

            // Draw Map:
            graphics.SetRenderTarget(_renderTarget);
            foreach (var layer in _tiledMap.Layers)
            {
                layer.Draw(camera);
            }
            graphics.SetRenderTarget(null);
            SpriteBatch.Begin(sortMode: SpriteSortMode.Immediate, blendState: BlendState.NonPremultiplied,
                samplerState: SamplerState.PointClamp, transformMatrix: camera.GetViewMatrix());
            SpriteBatch.Draw(_renderTarget, Vector2.Zero, Color.White);
            SpriteBatch.End();


**So that leads me to the next thing** - should there be a way to use this `Draw` from `TileMap` or should users manually draw their own layers when using this method?

craftworkgames:
So I merged your changes and had a bit of a play with them last night. I kinda got it working but had some problems with GetBoundingRectangle on the camera that I haven't figured out yet.

The first thing you should note is that the [`TiledMap.Draw` method (on line 108)](https://github.com/craftworkgames/MonoGame.Extended/blob/master/Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs#L108) already draws to a render target. So I don't think we need to do it in the layers, although, maybe that would be a better way of doing it.

With that in mind, the [`TiledTileLayer.Draw` method](https://github.com/craftworkgames/MonoGame.Extended/blob/master/Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs#L85) doesn't really need the camera passed in; which explains why I made it obsolete. Instead of passing in the camera we could just pass in a `visibileRectangle` instead. Something like:

        public override void Draw(RectangleF visibleRectangle)
        {
            var area = visibleRectangle;
            int firstCol = (int)Math.Floor(area.X / _map.TileWidth);
            int firstRow = (int)Math.Floor(area.Y / _map.TileHeight);

            // +3 to cover any gaps
            int numberOfColumns = (int)area.Width / _map.TileWidth + 3;
            int numberOfRows = (int)area.Height / _map.TileHeight + 3;

Next, it's possible to refactor the method to take render order into account by changing the [`GetRenderOrderFunction`](https://github.com/craftworkgames/MonoGame.Extended/blob/master/Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs#L157) to return a function that takes 4 parameters.

Then each of the render order functions would take the bounds of the camera rectangle like this:

        private IEnumerable<TiledTile> GetTilesRightDown(int left, int top, int right, int bottom)
        {
            for (var y = top; y < bottom; y++)
            {
                for (var x = left; x < right; x++)
                    yield return GetTile(x, y);
            }
        }

I kinda like this because the `for` statements say "top to bottom, left to right" etc.

Anyway, as I said I've already been playing around with these changes and they are a little difficult to explain so if you want to just PR the other stuff I'm happy to handle the render order fix.

Oh one last thing. The Sandbox project already has [some test tile maps](https://github.com/craftworkgames/MonoGame.Extended/tree/master/Source/Sandbox/Content/Tilesets) for showing render order. At the moment it's kind of a manual test that I run by just having the Tiled Map editor open next to me but I'd love to automate these kinds of tests with screenshots or something.

craftworkgames:
> I kinda got it working but had some problems with GetBoundingRectangle on the camera that I haven't figured out yet.

This was really bugging me so I got to fixing it tonight. The issue turned out to be that the size of the viewport changes when you set a render target. This was causing the calculated bounding rectangle of the camera to be out by a lot. The debugger was telling me that the visible rectangle was 91000 pixels wide :dizzy_face: 

Anyway, I got it sorted and made some other improvements. I hope you haven't made too many conflicting changes. If you do get a merge conflict I apologize in advance.

I'm really happy with the way it turned out. I'd welcome any other improvements to the library you'd like to make :smile:

cupstudio:
This looks good!  I'm seeing something odd in testing - it looks like the camera is not updating the visible bounds as it gets moved.

Hard to explain - [see this](https://youtu.be/udDSGzNPOD4).  I noticed that the camera's position and the X,Y in the bounding rectangle aren't the same even though I am getting those values in the same loop.


I'm not really doing much in my game screen, just letting the camera and map do its thing.


        public override void LoadContent()
        {
            var viewportAdapter = new DefaultViewportAdapter(graphics);
            camera = new Camera2D(viewportAdapter);
            Assets.MakeActionCatalog();
            _tiledMap = Content.Load<TiledMap>("testMapLarge");
        }

The Update just moves camera around:

        public override void Update(GameTime gameTime)
        {
            float deltaTime = (float)gameTime.ElapsedGameTime.TotalSeconds;
            if (keyboard.IsKeyPressed(Keys.W))
            {
                camera.Move(new Vector2(0, -400) * deltaTime);
            }
            if (keyboard.IsKeyPressed(Keys.A))
            {
                camera.Move(new Vector2(-400, 0) * deltaTime);
            }
            if (keyboard.IsKeyPressed(Keys.S))
            {
                camera.Move(new Vector2(0, 400) * deltaTime);
            }
            if (keyboard.IsKeyPressed(Keys.D))
            {
                camera.Move(new Vector2(400, 0) * deltaTime);
            }

            var area = camera.GetBoundingRectangle();
            console.Write("Camera Pos: " + camera.Position.ToString());
            console.Write("Camera Rect: " + area.ToString());
        }

Draw just draws map:

        public override void Draw(GameTime gameTime)
        {
            _tiledMap.Draw(camera);

            base.Draw(gameTime);
        }

craftworkgames:
> I noticed that the camera's position and the X,Y in the bounding rectangle aren't the same even though I am getting those values in the same loop.

That is very odd. I can't explain it either. I'll try reproduce it later today.

craftworkgames:
Try adding the Viewport to your debugging numbers. Maybe it's a similar bug to the one I discovered last night. It could be that the viewport size is not getting set back correctly when the render target changes again?

craftworkgames:
image

cupstudio:
I changed my movement keys get fine control: 

            if (keyboard.IsKeyTriggered(Keys.W))
            {
                camera.Move(new Vector2(0, -1));
            }

[You can see the bounding rectangle is doubling the actual movement.  The camera position is correct, but the bounding rectangle is twice the value I think it should be.

craftworkgames:
Okay, any idea what the problem might be? It kinda feels like the camera is zoomed out.

cupstudio:
Looking into this further.  Still not sure why it is behaving this way.  

#### Test

I created a stripped down test using only a `DefaultViewportAdapter` and `Camera2D`.  No Tiled map, no camera zoom or rotation.

````c#
_viewport = new DefaultViewportAdapter(_graphics);
_camera = new Camera2D(_viewport);
_camera.Move(new Vector2(2,0));
_camera.Move(new Vector2(0, 2));

Console.Write(string.Format("Camera Position: {0} \n", _camera.Position.ToString()));

var area = _camera.GetBoundingRectangle();
Console.Write(string.Format("Camera Bounds: X:{0}   Y:{1}\n", area.X.ToString(), area.Y.ToString()));

var boundingFrustum = _camera.GetBoundingFrustum();
var corners = boundingFrustum.GetCorners();
for (int i = 0; i < corners.Length; i++)
{
    Console.Write(string.Format("Corners [{0}]: {1}\n", i.ToString(), corners[i].ToString()));
           }
````

I would expect the Camera to be at position (2, 0) now.  I skipped the Bounds and went right for the Frustum.

````c#
var boundingFrustum = _camera.GetBoundingFrustum();
var corners = boundingFrustum.GetCorners();
for (int i = 0; i < corners.Length; i++)
{
    Console.Write(string.Format("Corners [{0}]: {1}\n", i.ToString(), corners[i].ToString()));
}
````

Output:

````
Camera Position: {X:2 Y:2} 
Camera Bounds: X:3.999939   Y:3.999968
Corners [0]: {X:3.999939 Y:3.999968 Z:1}
Corners [1]: {X:1284 Y:3.999968 Z:1}
Corners [2]: {X:1284 Y:723.9999 Z:1}
Corners [3]: {X:3.999939 Y:723.9999 Z:1}
Corners [4]: {X:3.999939 Y:3.999968 Z:0}
Corners [5]: {X:1284 Y:3.999968 Z:0}
Corners [6]: {X:1284 Y:723.9999 Z:0}
Corners [7]: {X:3.999939 Y:723.9999 Z:0}

````

#### Observation

It is pretty clear that moving the camera by 2 changes the actual position to 4 (though it is not a strict doubling based on the floats above, probably due to the transformation process).  This applies to both X and Y coordinates. 

I'm not sure if this is actually expected behavior and I'm missing something.  Please let me know if you think this is the case.


GetBoundingRectangle:
    GetBoundingFrustum:
        GetViewMatrix(Vector2.One)

cupstudio:
FYI still looking at this.  Right now I am down to this method in Camera2D:

````c#
        public Matrix GetViewMatrix(Vector2 parallaxFactor)
        {
            return 
                Matrix.CreateTranslation(new Vector3(-Position * parallaxFactor, 0.0f)) *
                Matrix.CreateTranslation(new Vector3(-Origin, 0.0f)) *
                Matrix.CreateRotationZ(Rotation) *
                Matrix.CreateScale(Zoom, Zoom, 1) *
                Matrix.CreateTranslation(new Vector3(Origin, 0.0f)) *
                _viewportAdapter.GetScaleMatrix(); 
        }
````

craftworkgames:
Thanks for looking into this. I spent some time on it a couple of days ago but I haven't had a chance to get back to it.

I like your test, we should turn that into an actual unit test.

At this stage I'm not entirely certain the camera is behaving as expected. I thought it was once, but now I'm not so sure. I think we need to really consider what the behavior of the camera should actually be. Maybe it's worth looking at [a camera from another framework](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/graphics/OrthographicCamera.html) to come up with a set of requirements.

cupstudio:
Thanks!  I think I found the issue.

The GetBoundingFrustrum multiplies viewMatrix and ProjectionMatrix, but GetProjectionMatrix already does that.

````c#
        public BoundingFrustum GetBoundingFrustum() 
        {
            var viewMatrix = GetViewMatrix();
            return new BoundingFrustum(viewMatrix * GetProjectionMatrix(viewMatrix));
        }
````

````c#
        private Matrix GetProjectionMatrix(Matrix viewMatrix)
        {
            var projection = Matrix.CreateOrthographicOffCenter(0, _viewportAdapter.VirtualWidth, _viewportAdapter.VirtualHeight, 0, -1, 0);
            Matrix.Multiply(ref viewMatrix, ref projection, out projection);    // <--------------
            return projection;
        }
````

I definitely think we should get some input from anyone using or testing the framework if possible.  Maybe create a new issue calling for suggestions.

Here are my thoughts:

* Position:  is public, so when changing the position it doesn't apply the Transform.  I would say either use methods to move the position and apply the transform, or update the `set { }` logic to apply the transform.  Unless there is another use-case I am not thinking of here.
* Origin: I think there should be some sort of option of controlling the camera via the Origin (a different camera class maybe).  I think all that amounts to is that the position is offset by -Origin.  
* Anchor: Possibility to bind Camera position based on an object that has a Vector2 Position?  It is easy enough to do manually but could be nice?

Right now in one of my tests I have the following to achieve this:

````c#
// creating player
player = new Player(knight);
player.Position = new Vector2(viewportAdapter.VirtualWidth / 2, viewportAdapter.VirtualHeight / 2);
player.OnPositionChanged += _moveCamera;
// ...
private void _moveCamera(object sender, PositionArgs e)
{
    camera.Move(e.DistanceMoved);
}
````

WardBenjamin:
I knew it!

I've been working on a camera system for my own game framework, (C#, currently built on top of OpenTK but moving to SDL probably), and I was reading through this code and wondered why the view matrix is multiplied twice. 

The usual way that it works is the Projection and View matrices are either referenced seperately and multiplied later by the user, or the camera just offers one combined "camera" matrix. Either way, those would both probably fix the issue.

craftworkgames:
So if we just change this:

        public BoundingFrustum GetBoundingFrustum() 
        {
            var viewMatrix = GetViewMatrix();
            return new BoundingFrustum(viewMatrix * GetProjectionMatrix(viewMatrix));
        }

To this:

        public BoundingFrustum GetBoundingFrustum() 
        {
            var viewMatrix = GetViewMatrix();
            return new BoundingFrustum(GetProjectionMatrix(viewMatrix));
        }

The bug is fixed?

@cupstudio 
> Position: is public, so when changing the position it doesn't apply the Transform. I would say either use methods to move the position and apply the transform, or update the set { } logic to apply the transform. Unless there is another use-case I am not thinking of here.

It doesn't apply the transform to what? Unless I'm misunderstanding something, there's no need to transform anything when the Position is set because the transform is applied when calling `GetViewMatrix()`

Of course, there is another pattern that might be appropriate here. I implemented something similar in the [PolygonF](https://github.com/craftworkgames/MonoGame.Extended/blob/master/Source/MonoGame.Extended/Shapes/PolygonF.cs) class recently. It's basically setting an `_isDirty` flag when a property changes to cause the transform to be recalculated on get operations. This pattern is more complicated to implement but probably provides the best performance in both situations (reading and writing properties).

> Origin: I think there should be some sort of option of controlling the camera via the Origin (a different camera class maybe). I think all that amounts to is that the position is offset by -Origin.

Sorry, I'm not sure what you mean? The camera already has an `Origin` property, although I do think it could do with some improvement.

> Anchor: Possibility to bind Camera position based on an object that has a Vector2 Position? It is easy enough to do manually but could be nice?

Maybe. How would it work? I don't really like the idea of having to call an `Update` method on the camera.

@WardBenjamin 
> The usual way that it works is the Projection and View matrices are either referenced seperately and multiplied later by the user, or the camera just offers one combined "camera" matrix.

Do you have an example (from another framework perhaps?)

The [camera in LibGDX](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/Camera.java) is probably a good reference.

craftworkgames:
Okay. I [committed](https://github.com/craftworkgames/MonoGame.Extended/commit/943479a1b22355399465204c309e860d0095c990) a fix for the camera and wrote a unit test around `GetBoundingRectangle`

cupstudio:
Ok I've tested the change and it looks good!  The unit test ran find, and I re-tested my game screen and the map was drawing exactly as expected.

> It doesn't apply the transform to what?

I saw that Camera.Move applies a transform when updating the Position, but the position can also be changed just by directly setting a value.  Is this intended?

````c#
public Vector2 Position { get; set; }
// ...
public void Move(Vector2 direction)
{
    Position += Vector2.Transform(direction, Matrix.CreateRotationZ(-Rotation));
}
````


> Sorry, I'm not sure what you mean? The camera already has an Origin property, although I do think it could do with some improvement.

I didn't explain it well.  The camera "location" is really always 0,0 in relation to the screen.  A subclass could treat that as viewportWidth / 2, viewportHeight / 2 so that it is essentially always centered and you work from there.  This would be related to #81.  If no one else gets to it, I will work up an example at some point.  I don't think its a burning issue at all - this camera is working quite well for me now.


> Maybe. How would it work? I don't really like the idea of having to call an Update method on the camera.

No update necessary.  I've implemented anchoring using events.  Again this is not something I see as a critical issue and things are working for me, just throwing stuff out there.

Here is an example of my anchoring system (feel free to comment!)

##### Interface:
````c#
    public interface IAnchorable
    {
        Rectangle Bounds { get; }
        event EventHandler OnLocationChanged;
    }
````

##### Implementation: 
````c#
public event EventHandler OnLocationChanged;
private IAnchorable _anchorTarget;
public IAnchorable AnchorTarget { get { return _anchorTarget; } }
private PositionType _anchorStyle;
private Vector2 _anchorOffset;

public void AnchorTo(IAnchorable target, PositionType anchorStyle, Vector2 offset)
{
    _anchorTarget = target;
    _anchorStyle = anchorStyle;
    _anchorOffset = offset;
    _anchorTarget.OnLocationChanged += _getAnchorUpdate;
    Position = Helper.GetPosition(_anchorTarget.Bounds, Bounds, _anchorStyle, _anchorOffset);
}

private void _getAnchorUpdate(object sender, EventArgs e)
{
    // here you can either just update position or use a dirty bit
    // _positionDirty = true;
    Position = Helper.GetPosition(_anchorTarget.Bounds, Bounds, _anchorStyle, _anchorOffset);
}

public void RemoveAnchor()
{
    if (_anchorTarget != null)
        {
             _anchorTarget.OnLocationChanged -= _getAnchorUpdate;
            _anchorTarget = null;
        }
}
````

##### Static GetPosition

````c#
public static Vector2 GetPosition(Rectangle master, Rectangle minion, PositionType pos, Vector2 offset)
        {
            Rectangle a = master;
            Rectangle b = minion;

            // check parameters
            if (offset == null)
            {
                offset = Vector2.Zero;
            }
            if (master == minion)
            {
                // who you foolin'
                return new Vector2(a.X, a.Y);
            }
            switch (pos)
            {
                case PositionType.InsideTopLeft:
                    return new Vector2(a.X + offset.X, a.Y + offset.Y);

                case PositionType.InsideTopCenter:
                    return new Vector2((int)((a.X + a.Width / 2) - (b.Width / 2) + offset.X), a.Y + offset.Y);

                case PositionType.InsideTopRight:
                    return new Vector2(a.X + offset.X + a.Width - b.Width, a.Y + offset.Y);

                case PositionType.InsideMiddleLeft:
                    return new Vector2(a.X + offset.X, (int)(a.Y + (a.Height / 2) - (b.Height / 2) + offset.Y));

                case PositionType.InsideMiddleCenter:
                    return new Vector2((int)(a.X + (a.Width / 2) - (b.Width / 2) + offset.X), (int)(a.Y + (a.Height / 2) - (b.Height / 2) + offset.Y));

                case PositionType.InsideMiddleRight:
                    return new Vector2(a.X + a.Width - b.Width + offset.X, (int)(a.Y + (a.Height / 2) - (b.Height / 2) + offset.Y));

                case PositionType.InsideBottomLeft:
                    return new Vector2(a.X + offset.X, a.Y + a.Height - b.Height + offset.Y);

                case PositionType.InsideBottomCenter:
                    return new Vector2((int)(a.X + (a.Width / 2) - (b.Width / 2) + offset.X), a.Y + a.Height - b.Height + offset.Y);

                case PositionType.InsideBottomRight:
                    return new Vector2(a.X + a.Width - b.Width + offset.X, a.Y + a.Height - b.Height + offset.Y);

                case PositionType.OutsideTopLeft:
                    return new Vector2(a.X - b.Width + offset.X, a.Y + offset.Y);

                case PositionType.OutsideTopCenter:
                    return new Vector2((int)(a.X + (a.Width / 2) - (b.Width / 2) + offset.X), a.Y - b.Height + offset.Y);

                case PositionType.OutsideTopRight:
                    return new Vector2(a.X + a.Width + offset.X, a.Y + offset.Y);

                case PositionType.OutsideMiddleLeft:
                    return new Vector2(a.X - b.Width + offset.X, (int)(a.Y + (a.Width / 2) - (b.Width / 2) + offset.Y));

                case PositionType.OutsideMiddleRight:
                    return new Vector2(a.X + a.Width + offset.X, (int)(a.Y + (a.Height / 2) - (b.Height / 2) + offset.Y));

                case PositionType.OutsideBottomLeft:
                    return new Vector2(a.X + offset.X, a.Y + a.Height + offset.Y);

                case PositionType.OutsideBottomCenter:
                    return new Vector2((int)(a.X + (a.Width / 2) - (b.Width / 2) + offset.X), a.Y + a.Height + offset.Y);

                case PositionType.OutsideBottomRight:
                    return new Vector2(a.X + a.Width + offset.X, a.Y + a.Height + offset.Y);

                default:
                    throw new NotImplementedException("Position type supplied is not implemented.");
            }
        }
````

craftworkgames:
> I saw that Camera.Move applies a transform when updating the Position, but the position can also be changed just by directly setting a value. Is this intended?

Yes, it's intended. Although, I'm always open to debate. The idea is to provide two different ways to move the camera. To illustrate the point, imagine this cross as the camera, and it's rotated slightly clockwise to the right.

image

What do you do if you wanted to move the camera up?  If the camera is rotated this can mean one of two things. 

Either you are moving it up relative to the camera's up vector accounting for rotation (blue arrow) 

    camera.Move(new Vector2(0, 10));

or relative to the screens up vector (red arrow).

    camera.Position += new Vector2(0, 10);

I realize the distinction isn't immediately clear. It could probably do with some documentation or some refactoring of the code. But anyway, I hope that makes it clear for you at least.

> I've implemented anchoring using events. Again this is not something I see as a critical issue and things are working for me, just throwing stuff out there.

Anchoring is an interesting idea but I don't want to overload the camera class with too many responsibilities. Maybe we'll implement a more generic way to anchor two movable things together at some stage, but I don't think it's a priority right now.

cupstudio:
Ok that does make sense now!  Thanks for clarifying that!

PR #92, opened 2016-01-20
New functionality: GamePadListener

detuur:
Hey guys.

I've been using Monogame.Extended in a school project, and I'm very happy with it. However, one feature was missing in particular to me, which was a `GamePadListener`. So I decided to implement one myself.

The functionality is very much complete. It has several events for buttons going up, down, being held down, thumbsticks moving, triggers moving, controllers connecting and disconnecting.
There is an extensive associated settings class that allows users to fine-tune the behaviour of the listener. It even has a very simple public method to set a vibration that is very much "fire and forget" (the listener brings the vibration down for you, using a duration parameter, instead of having to manually set both vibes to 0). About the only thing that doesn't work is multiple buttons firing `ButtonRepeated` at the same time. Doing this would require computationally heavy code which I consider not worth it. Besides, this mimics most games' behaviour.

There is a change to the `InputListenerManager` to call the static `GamePadListener.CheckConnections()` method every update. This is so that it can fire the static `ControllerConnectionChanged` event. It only does this when `GamePadListener.CheckControllerConnections` is set to true, however.

One thing that I feel that I do have to mention is that Monogame's GamePad **only works** on particular platforms. More specifically, those which use Xinput (afaik). I developed against Monogame.Windows with a PS3 controller hooked up through [SCP Toolkit](https://github.com/nefarius/ScpToolkit) and everything worked perfectly.

The code is extensively documented for the convenience of both users and future maintainers.

(First contribution to an open source project ever, by the way)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ MonoGame.Extended is designed to build on top of MonoGame, but may work with XNA
 - **Tile based maps** using [Tiled](http://www.mapeditor.org/) (orthogonal only in v0.3, isometric in master)
 - **Bitmap fonts** using [BMFont](http://www.angelcode.com/products/bmfont/)
 - **[Sprites](http://dylanwilson.net/sprites-and-spritebatch-extensions-in-monogame-extended)** (with SpriteBatch extensions!)
-- **Input listeners** for event driven input handling (Keyboard, Mouse, Touch)
+- **Input listeners** for event driven input handling (Keyboard, Mouse, Touch) (GamePad in master)
 - **Texture Atlases** using the JSON format in [TexturePacker](https://www.codeandweb.com/texturepacker)
 - **2D Camera** with pan, zoom, and rotation
 - **Viewport Adapters** for resolution independent rendering
--- a/Source/MonoGame.Extended/InputListeners/GamePadEventArgs.cs
+++ b/Source/MonoGame.Extended/InputListeners/GamePadEventArgs.cs
@@ -1,0 +1,62 @@
+﻿using System;
+using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Input;
+
+namespace MonoGame.Extended.InputListeners
+{
+    /// <summary>
+    /// This class contains all information resulting from events fired by
+    /// <see cref="GamePadListener"/>.
+    /// </summary>
+    public class GamePadEventArgs : EventArgs
+    {
+        internal GamePadEventArgs(GamePadState previousState, GamePadState currentState,
+            TimeSpan elapsedTime, PlayerIndex playerIndex, Buttons? button = null,
+            float triggerState = 0, Vector2? thumbStickState = null)
+        {
+            PlayerIndex = playerIndex;
+            PreviousState = previousState;
+            CurrentState = currentState;
+            ElapsedTime = elapsedTime;
+            if (button != null)
+                Button = button.Value;
+            TriggerState = triggerState;
+            ThumbStickState = thumbStickState ?? Vector2.Zero;
+        }
+
+        /// <summary>
+        /// The index of the controller.
+        /// </summary>
+        public PlayerIndex PlayerIndex { get; private set; }
+
+        /// <summary>
+        /// The state of the controller in the previous update.
+        /// </summary>
+        public GamePadState PreviousState { get; private set; }
+
+        /// <summary>
+        /// The state of the controller in this update.
+        /// </summary>
+        public GamePadState CurrentState { get; private set; }
+
+        /// <summary>
+        /// The button that triggered this event, if appliable.
+        /// </summary>
+        public Buttons Button { get; private set; }
+
+        /// <summary>
+        /// The time elapsed since last event.
+        /// </summary>
+        public TimeSpan ElapsedTime { get; private set; }
+
+        /// <summary>
+        /// If a TriggerMoved event, displays the responsible trigger's position.
+        /// </summary>
+        public float TriggerState { get; private set; }
+
+        /// <summary>
+        /// If a ThumbStickMoved event, displays the responsible stick's position.
+        /// </summary>
+        public Vector2 ThumbStickState { get; private set; }
+    }
+}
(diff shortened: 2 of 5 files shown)


craftworkgames:
Excellent work. Thanks for the contribution! :smile: 

> There is a change to the InputListenerManager to call the static GamePadListener.CheckConnections() method every update.

This is the only change that irks me a little. Is there any reason why this can't live in the `GamePadListener.Update` method?

> #region events

To be consistent with the coding style in the project please don't use `#region`'s.

> - **Input listeners** for event driven input handling (Keyboard, Mouse, Touch) (GamePad in master)

I think it's excellent that you updated the README. Documentation is often forgotten, even by myself :disappointed: However, since this functionality wasn't delivered in version 0.2 we should probably make a new heading. Something like "Features in v0.5 (in development)" perhaps.

None of these changes are a big deal. If you like I'm happy to merge the pull request and fix these things up myself. The only difficult thing is that I don't have a game pad to test it on.

cupstudio:
Only out of curiosity, why no regions?  Preference only or are there other concerns?

craftworkgames:
> Only out of curiosity, why no regions? Preference only or are there other concerns?

It's just a preference. Here are my reasons.
 - when they are collapsed you can easily scroll paste important code
 - when they are expanded they take up extra lines.
 - they usually just state the obvious (#region properties). yep, I can see that already.
 - sometimes they create an excuse to make really, really big classes.
 - but most of all, i just hate expanding and collapsing them all the time.

Obviously, it doesn't make a lot of difference to the actual output of the code. But it can lead to bad design.

I'm not the only one that feels this way either. There's a pretty good discussion on [programmer's stack exchange](http://programmers.stackexchange.com/questions/53086/are-regions-an-antipattern-or-code-smell) and also [this](http://blog.codinghorror.com/the-problem-with-code-folding/) :wink: 

@detuur Are we good to go with the merge?

detuur:
> Is there any reason why this can't live in the `GamePadListener.Update` method?

Yes actually. This is a static method that fires whenever any controller connects/disconnects. The idea is to allow people to listen for controllers connecting, and then attach listeners to them. If you were to put this in `GamePadListener.Update`, it would get called anywhere from 0 to 4 times per update.

Honestly, I don't like how it's hardcoded into the manager either, but this was the only sensible way out I could come up with.

The alternative is to can the method as it is now, and instead make it an instance method that only updates on the currently assigned controller. The utility of this will obviously be a bit less, but at the end of the day, if your game supports up to 4 controllers (or more*), then it shouldn't be a problem to keep 4 dormant gamepadlisteners.

I'll leave it up to you, you're the maintainer and you have the best idea of where you want to go with the functionality.

>Documentation is often forgotten

Honestly in my experience the state of the documentation did indeed work against me when I was developing my game with Monogame.Extended. There's only so much information you can derive from the function names before you have to resort to reading the sample code or even M.E's source. Having either an up-to-date external doc or IntelliSense documentation would make this a lot easier.

*: If I ever have too much time, I'll try to see if casting higher integers to the PlayerIndex enum also allows you to access more controllers. My bet is it will. This would make interesting functionality to neatly expose in Monogame.Extended.

detuur:
Sorry about taking my time, I had my CCNA exam at college today.

Weighing in myself on the `#regions` thing, I too think that a class of pure code shouldn't need these directives in order to stay readable/navigable, but the fact of the matter is that once you document your code with XML comments, it suddenly gains a huge amount of clunk. And there's no way around this in VS(2015). There isn't any option that automatically collapses all XML comments, so the way I do it right now is to `#region` out the heavily documented public properties at the start, so that I can get to the meat of my code more quickly.

Ideally, VS should allow you to close the XML docs by default.


So yes, @craftworkgames , you can merge however you see fit.

craftworkgames:
> Yes actually. This is a static method that fires whenever any controller connects/disconnects. The idea is to allow people to listen for controllers connecting, and then attach listeners to them.

Good to know. I'll keep that in mind if I do any refactoring. Unfortunately I don't have a game pad to test it on so I'm going to be very careful not to break anything.

> If you were to put this in GamePadListener.Update, it would get called anywhere from 0 to 4 times per update.

I thought about that. I think we could prevent it from being called multiple times fairly easily simply by keeping a count of the number of listeners attached.

> Honestly, I don't like how it's hardcoded into the manager either, but this was the only sensible way out I could come up with.

There's 2 downsides. The first is more of a code smell because it's leaking responsibilities into the manager. The second is that the method is going to get called over and over for no reason in games that have nothing to do with game pads. The performance penalty is probably next to nothing though so it doesn't really matter.

> The alternative is to can the method as it is now, and instead make it an instance method that only updates on the currently assigned controller. 

Yep, that's a reasonable approach. Another alternative would be to pull the call into it's own listener (something like `GamePadConnectionListener`). The downside is that it's not obvious someone would need to add one of these to the collection but it would be a cleaner design.

> Honestly in my experience the state of the documentation did indeed work against me when I was developing my game with Monogame.Extended.

Yep, I understand. Documentation is king when it comes to open source projects in my experience. The problem is, it takes a lot of work to write and maintain it. 

That said, I don't want to make excuses so let's declare now that we will start sorting it out. I'd like to start using [the wiki](https://github.com/craftworkgames/MonoGame.Extended/wiki) more to create some documentation like [LibGDX](https://github.com/libgdx/libgdx/wiki). I was also reading some nice documentation today on [knockoutjs](http://knockoutjs.com/documentation/introduction.html).

> the fact of the matter is that once you document your code with XML comments, it suddenly gains a huge amount of clunk. 

Yeah, I hate the clunk too but I guess it's a necessary evil to make a great open source project. At some point I'm just going to have to bite the bullet and start documenting everything. 

> And there's no way around this in VS(2015). There isn't any option that automatically collapses all XML comments

I did use the [HideShow Comments](https://visualstudiogallery.msdn.microsoft.com/03141ea7-c35e-4533-b05b-9e60545e93eb) for a while but at the time it was a bit buggy. Maybe things have improved.

The thing is, when I started MonoGame.Extended I had no idea if people would actually use it or not. So I just focused on the code and making it clear what the project was about. Now that there's a whopping 50+ stars on github things are starting to take off and the project is getting more serious :smile: 

> when I was developing my game with Monogame.Extended.

Last but not least I would *love* to see the game you've make with MonoGame.Extended!  I'd love to start a showcase of games made with the library.

cupstudio:
> Now that there's a whopping 50+ stars on github things are starting to take off and the project is getting more serious

Before long you will only have time to manage pull requests!!  :grinning:

PR #95, opened 2016-01-23
MouseListener Update for new Drag events - Issue 94

cupstudio:
Sorry about the whitespace changes, I must have had a copy of TiledTileLayer.cs open when I hit <kbd>Ctrl + K, D</kbd>.

* Added a convenience function in MouseEventArgs for distance moved.
* Added three events: `MouseDragStart` , `MouseDrag` , `MouseDragEnd`
* Added convenience function for `MouseHasMoved()`
* `CheckMouseDragged` implemented in similar fashion to CheckButton methods.
* `MouseDragged` is now `MouseDragEnd` and only triggers if currently dragging

Drag Start relies on the DragThreshold, moving after that is based only on `MouseHasMoved()`

I implemented for all mouse buttons because I could see cases where you may want to drag based on other buttons (moving a map around vs a panel).  User would need to track the mouse button via the MouseEventArgs, otherwise any button will be able to initiate a drag (this is the same for clicks, too).

Example Usage:
In this scenario we have a list of things we are going to check and see what is on top for movement.  The resulting object is stored while dragging in _draggedItem.  Note that you also need to perform movement on DragStart and DragEnd.

Hookup events in your class:

````c#
_mouse.MouseDragStart += DragStart;
_mouse.MouseDrag += Drag;
_mouse.MouseDragEnd += DragEnd;
````

Something to drag

````c#
private IDraggable _draggedItem;
````

Implement

````c# 
        private void Drag(object sender, MouseEventArgs e)
        {
            if (_draggedItem != null)
            {
                _draggedItem.Drag(e);
            }
        }


        private void DragEnd(object sender, MouseEventArgs e)
        {
            if (_draggedItem != null)
            {
                _draggedItem.Drag(e);
                Console.WriteLine("Drag End: " + e.DistanceMoved.ToString());
                _draggedItem = null;
            }
        }


        private void DragStart(object sender, MouseEventArgs e)
        {
            Control c = GetControlAtPoint(e.Position);
            if (c is IDraggable)
            {
                _draggedItem = (IDraggable)c;
                _draggedItem.Drag(e);
                Console.WriteLine("Drag Start: " + e.DistanceMoved.ToString());
            }
        }
````


craftworkgames:
Thanks. These changes look great :smile:

PR #96, opened 2016-01-23
Fixed BoxingViewportAdapter for DirectX Rendering

Mikescher:
Hi,

I managed to find a bug in the `BoxingViewportAdapter` with DirectX rendering, annoyingly it works with MonoGame.Linux and MonoGame.OpenGL without problems.
But with MonoGame.DirectX the PrefferedBackBufferSize isn't correctly updated when the window is resized, which results in a distortion of the rendering.
 *(see: http://gfycat.com/BasicLegalHerculesbeetle)*

To fix this the PrefferedBackBufferSize has to be manually updated.
*(http://gamedev.stackexchange.com/questions/68914/issue-with-monogame-resizing)*

The `BoxingViewportAdapter` now needs to get a `GraphicsDeviceManager` in its constructor instead of an `GraphicsDevice`. Perhaps that should be changed in the other Adapters too. So that the interface stays symmetrical? If you want I can change that in this pull request too.

&nbsp;

Greetings,
Mike

PS: Nice library you're doing here. I'm planning to do a few games for android and looking forward to use it :D

### DIFF
--- a/Source/MonoGame.Extended/ViewportAdapters/BoxingViewportAdapter.cs
+++ b/Source/MonoGame.Extended/ViewportAdapters/BoxingViewportAdapter.cs
@@ -12,14 +12,20 @@ namespace MonoGame.Extended.ViewportAdapters
 
     public class BoxingViewportAdapter : ScalingViewportAdapter
     {
-        public BoxingViewportAdapter(GameWindow window, GraphicsDevice graphicsDevice, int virtualWidth, int virtualHeight)
-            : base(graphicsDevice, virtualWidth, virtualHeight)
+        private readonly GraphicsDeviceManager _graphicsDeviceManager;
+        private readonly GameWindow _window;
+
+        public BoxingViewportAdapter(GameWindow window, GraphicsDeviceManager graphicsDeviceManager, int virtualWidth, int virtualHeight)
+            : base(graphicsDeviceManager.GraphicsDevice, virtualWidth, virtualHeight)
         {
+            _graphicsDeviceManager = graphicsDeviceManager;
+            _window = window;
+
             window.ClientSizeChanged += OnClientSizeChanged;
         }
 
         public BoxingMode BoxingMode { get; private set; }
-        
+
         private void OnClientSizeChanged(object sender, EventArgs eventArgs)
         {
             var viewport = GraphicsDevice.Viewport;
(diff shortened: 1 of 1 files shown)


craftworkgames:
Thanks for the fix! I remember discussing this issue somewhere before but I can't seem to find it. Anyway, we originally came to the conclusion that nothing could be done about it until the next release of MonoGame. I'm glad you found a workaround :smile: 

As this is hopefully only a temporary workaround for people using DirectX I'm thinking we should take care not to break the existing constructor. The easiest way to do that is to introduce a second constructor and document them clearly. Something like this:

    /// <summary>
    /// Initializes a new instance of the <see cref="BoxingViewportAdapter"/>. 
    /// Use this constructor only if you're using DirectX due to a bug in MonoGame.
    /// This constructor will be made obsolete and eventually removed once the bug has been fixed.
    /// </summary>
    public BoxingViewportAdapter(GameWindow window, GraphicsDeviceManager graphicsDeviceManager, int virtualWidth, int virtualHeight)

    /// <summary>
    /// Initializes a new instance of the <see cref="BoxingViewportAdapter"/>. 
    /// Note: If you're using DirectX please use the other constructor due to a bug in MonoGame.
    /// </summary>
    public BoxingViewportAdapter(GameWindow window, GraphicsDevice graphicsDevice, int virtualWidth, int virtualHeight)

If you don't have any objections I'll go ahead and do this?

> Perhaps that should be changed in the other Adapters too. So that the interface stays symmetrical?

Constructors are not really part of the "interface". Ideally they should only take the dependencies required. It's not unusual to have different dependencies in related constructors. As you can see, we're already doing that.

    public DefaultViewportAdapter(GraphicsDevice graphicsDevice)
    public ScalingViewportAdapter(GraphicsDevice graphicsDevice, int virtualWidth, int virtualHeight) 
    public BoxingViewportAdapter(GameWindow window, GraphicsDevice graphicsDevice, int virtualWidth, int virtualHeight)

Changing them to all take a `GraphicsDeviceManager` instead of `GraphicsDevice` would only confuse the situation more be a bigger breaking change for no benefit.

> PS: Nice library you're doing here. I'm planning to do a few games for android and looking forward to use it :D

Fantastic! Be sure to [let us know](http://community.monogame.net/category/extended) what you're working on and post some screenshots :grinning: I'd like to start putting together a showcase of what people are using the library for.

craftworkgames:
I found [the issue](https://github.com/mono/MonoGame/issues/4018) where I was discussing this bug a while back.

wmansfield:
I have tried both the original and the updated on ios. Neither work. It appears that it has a similar problem with a different work around.  So far I have only been able to get it to work when forcing a .Reset() a single time in the Update() method.  Any ideas which event can be hooked to ensure it stays?

PR #97, opened 2016-01-24
Add class for common math operations on floats

Mikescher:
Currently you have to cast pretty often to `float` because everything in the [System.Math](https://msdn.microsoft.com/de-de/library/system.math%28v=vs.110%29.aspx) class returns a `double` but most stuff in MonoGame expects `float`.

This implements a util class with all the useful mathematical functions on float values.
There is something similiar in libGDX under [com/badlogic/gdx/math/MathUtils](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/MathUtils.java).

Currently most methods just call `System.Math` and then cast to `float`. But in the libGDX code are a few faster algorithms for the trigonometric functions. It could be worth the consideration to include them here too?
(Would need to take a look at the license, if they can be freely copied)

Suggestions for improvement are welcome :smile:

### DIFF
--- a/Source/MonoGame.Extended/FloatMath.cs
+++ b/Source/MonoGame.Extended/FloatMath.cs
@@ -1,0 +1,209 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MonoGame.Extended
+{
+    public static class FloatMath
+    {
+        public const float PI  = 3.1415927f;
+        public const float TAU = PI*2;
+        public const float E = 2.7182818f;
+
+        public const float RadiansToDegrees = 180f / PI;
+        public const float RadDeg = RadiansToDegrees;
+
+        public const float DegreesToRadians = PI / 180;
+        public const float DegRad = DegreesToRadians;
+
+        public static float Asin(float value)
+        {
+            return (float) Math.Asin(value);
+        }
+
+        public static float Sin(float value)
+        {
+            return (float)Math.Sin(value);
+        }
+
+        public static float Sinh(float value)
+        {
+            return (float)Math.Sinh(value);
+        }
+
+        public static float Acos(float value)
+        {
+            return (float)Math.Acos(value);
+        }
+
+        public static float Cos(float value)
+        {
+            return (float)Math.Cos(value);
+        }
+
+        public static float Cosh(float value)
+        {
+            return (float)Math.Cosh(value);
+        }
+
+        public static float Atan(float value)
+        {
+            return (float)Math.Atan(value);
+        }
+
+        public static float Atan2(float y, float x)
+        {
+            return (float)Math.Atan2(y, x);
+        }
+
+        public static float Tan(float value)
+        {
+            return (float)Math.Tan(value);
+        }
+
+        public static float Tanh(float value)
+        {
+            return (float)Math.Tanh(value);
+        }
+
+        public static float Abs(float value)
+        {
+            return (float)Math.Abs(value);
+        }
+
+        public static float Ceiling(float value)
+        {
+            return (float)Math.Ceiling(value);
+        }
+
+        public static float Floor(float value)
+        {
+            return (float)Math.Floor(value);
+        }
+
+        public static float Round(float value)
+        {
+            return (float)Math.Round(value);
+        }
+
+        public static float Round(float value, int digits)
+        {
+            return (float)Math.Round(value, digits);
+        }
+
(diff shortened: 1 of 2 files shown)


craftworkgames:
Hmmm... The implementation looks reasonable.

I'm not sure how much value it adds though if it's mostly just casting things to floats. Compared with the cost of maintaining it in the library, potential bugs, etc.

Plus, now there's 3 math helpers to think about. MonoGame already has a [MathHelper](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/MathHelper.cs) class.

That said, I'm not overly bothered either way if we include it. People don't have to use it if they don't want to I guess. Maybe we should get some more feedback from the community before including this?

detuur:
Like Dylan said, it adds very little functionality.It basically amounts to (very simple) syntactical sugar. This is not a problem for any developer to implement themselves. See if you can implement (and benchmark?) some of those faster algorithms. As it stands, this sacrifices maintainability for a too small gain in functionality.

So I can only rationalise it if it brings actual reimplementation with it. If that's done, I'm a fan.

Mikescher:
> MonoGame already has a [MathHelper](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/MathHelper.cs) class.

Ah, totally missed that class. Only looked in this lib and the .Net Framework...

> See if you can implement (and benchmark?) some of those faster algorithms.

Yeah I think, I will do this this afternoon and see if they provide a good improvement over the standard implementation

Mikescher:
Ok that was dissapointing. It seems like the standard Math functions are way better - even if you have to do the roundtrib over double values. (I even ignored the initial creation time of the SINUS_TABLE)

Method     | Iterations    |  System.Math  |  Ported code from libGDX
-----------|---------------|---------------|-----------------------
Sinus      | 2.000.000.000 | 645 ms        | 1.211 ms
Cosinus    | 2.000.000.000 | 1.760 ms      | 1.827 ms
Atan2      |    10.000.000 | 200 ms        | 577 ms

Looking further into this, the System.Math methods are apparently [implemented](http://referencesource.microsoft.com/#mscorlib/system/math.cs,a4407e67b9a5afad) directly in the CLR.
So its implemented in the [underlying C Code](https://msdn.microsoft.com/de-de/library/system.runtime.compilerservices.methodimploptions%28v=vs.110%29.aspx), which explains that it's faster.

So we can probably close this pull request and I move the class directly into my project (because I still like the convenience methods :smile:)

craftworkgames:
Hehe. Well at least it was an interesting experience in measuring things.

tstavrianos:
Apologies for resurrecting this.
@Mikescher how did you benchmark this because with nearly identical code to that of libgdx (I use a 2nd table for cosine instead of using a single one for both sine and cosine, faster at the expense of some extra memory) I am getting significantly different results.

My results:

Method|Iterations|System.Math|LibGDX
--------|----------|-------------|--------
Sinus|2,000,000,000|60,462.00 ms|7,963,90 ms
Cosinus|2,000,000,000|71,154.40 ms|7,486.00 ms
Atan2|10,000,000|482.30 ms|64.20 ms

Not a typo, the LibGDX functions are 10x faster in my test.

This is my bench function:

>         
        static void Bench(string description, Action act, int Iterations)
        {
            Process.GetCurrentProcess().PriorityClass = ProcessPriorityClass.High;
            Thread.CurrentThread.Priority = ThreadPriority.Highest;

            // Warm up.
            act();
            var sw = new Stopwatch();

            GC.Collect();
            GC.WaitForPendingFinalizers();
            GC.Collect();

            Console.WriteLine("Benching {0}", description);
            sw.Start();
            for (var i = 0; i < Iterations; i++)
                act();
            sw.Stop();
            Console.WriteLine("Took {0:0.0} ms for {1} iterations", sw.Elapsed.TotalMilliseconds, Iterations);
        }

craftworkgames:
@tstavrianos A couple of ideas spring to mind.

1. It's probably super obvious, but did you run it in Release mode?
2. What result do you get if `act` does absolutely nothing?
3. What result do you get if you replace `act` with the actual method call rather than using a lambda expression?

My gut feeling is that most of the time is being taken up by something other than what you're trying to benchmark.

tstavrianos:
1. Yes, all tests were run in release.
2. Takes about 3300 - 3400ms.
3. Good catch. Replacing the lambda with the actual function makes my function 2x faster but makes the native function 100x faster. Time to re-check all my math functions again :)

PR #98, opened 2016-01-28
Add cast operators to and from Rectangle

stromkos:
Allows RectangleF to be explicitly cast to Rectangle.


craftworkgames:
Excellent. Thanks :)

Issue #105, opened 2016-02-02
Release early, release often


craftworkgames:
One of motto's is ["Release early, release often"](https://en.wikipedia.org/wiki/Release_early,_release_often). My plan has always been to do a release every month or so.

So far we've had [4ish releases](https://www.nuget.org/packages/MonoGame.Extended/) (not counting minor bug fixes).
    
    Version	                                    Downloads	Last updated
    MonoGame.Extended   0.4.64 (this version)	294	        13 December 2015
    MonoGame.Extended   0.3.44-alpha	        271	        23 September 2015
    MonoGame.Extended   0.2.7-alpha	            126	        26 August 2015
    MonoGame.Extended   0.1.1-alpha	            34	        02 August 2015

This is okay. It's certainly a lot better than MonoGame's release cycle :stuck_out_tongue_winking_eye: 

But I think we can do a lot better. 

Pushing a release is easy. I've got an automated build to push a NuGet package from `master` in one click. The real problem is that it's difficult to get the code ready for release. To make sure that nothing makes it into the release that shouldn't be there.

I'm guilty of checking all my stuff into `master`. It seemed like the right thing to do at the time, but it's not working very well now. `master` is getting dirty, with half finished features.

So I propose we start doing [feature branching](http://martinfowler.com/bliki/FeatureBranch.html) and push a release *every* time we pull a completed* feature into `master`.

* Of course, the definition of *completed* is going to vary. We're not looking for perfection, just something that looks pretty good from an API point of view.

craftworkgames:
Before we can do this properly I think we need to cleanup `master` and remove anything that shouldn't be there. Fortunately, the project namespaces organized nicely into features.

![image](https://cloud.githubusercontent.com/assets/3201643/12768613/8c24d7e6-ca5c-11e5-8a4f-76c050bb4a33.png)

The features marked with a red dot are my candidates for removal. Here's why:

 **Animations** - the animations feature was kind of an experiment to write a [sprite sheet animation tool](http://dylanwilson.net/introducing-astrid-animator). It kinda works but the code is a mess and probably shouldn't have been released as is. At the very least this feature needs a cleanup before the 0.5 release.

**Collisions** - the collision detection code is really rough and should have also existed in a feature branch before it made it into the release. @LithiumToast is planning to do some work on #74 to make this a lot better.

**Gui** & **Screens**  - is what I'm currently working on. There's still a lot to do so I'm going to move it into a feature branch.

[...]

PR #107, opened 2016-02-03
Add alpha feature to Sprites

rafaelalmeidatk:
Now you can change the opacity of a sprite.


craftworkgames:
Good idea.

Just to let you know, we have been discussing using feature branches #105 instead of checking everything into `master`. This change is so minor it probably doesn't make any difference but if you work on something bigger please put it in a feature branch.

Thanks.

PR #110, opened 2016-02-04
Allow drawing tiled map with SpriteBatch

cra0zy:
Very basic, I just needed those few params, but I would recommend you setup maybe some extension method for spriteBatch, and adding some sort of range for rendering for spriteBatch drawing.

### DIFF
--- a/Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
+++ b/Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
@@ -92,6 +92,29 @@ namespace MonoGame.Extended.Maps.Tiled
             _spriteBatch.End();
         }
 
+        public override void Draw(SpriteBatch spriteBatch, Vector2 position)
+        {
+            var renderOrderFunction = GetRenderOrderFunction();
+            var tileLocationFunction = GetTileLocationFunction();
+            var firstCol = 0;
+            var firstRow = 0;
+
+            var columns = _map.Width + 3;
+            var rows = _map.Height + 3;
+
+            foreach (var tile in renderOrderFunction(firstCol, firstRow, firstCol + columns, firstRow + rows))
+            {
+                var region = tile != null ? _map.GetTileRegion(tile.Id) : null;
+
+                if (region != null)
+                {
+                    var point = tileLocationFunction(tile);
+                    var destinationRectangle = new Rectangle(point.X + (int)position.X, point.Y + (int)position.Y, region.Width, region.Height);
+                    spriteBatch.Draw(region, destinationRectangle, Color.White);
+                }
+            }
+        }
+
         private Func<TiledTile, Point> GetTileLocationFunction()
         {
             switch (_map.Orientation)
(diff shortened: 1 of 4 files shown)


Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
```diff
@@ -92,6 +92,29 @@ public override void Draw(RectangleF visibleRectangle)
             _spriteBatch.End();
         }
```
craftworkgames: Copying and pasting the whole method is not a very maintainable solution. Also, by removing the visible rectangle you've removed an important optimization for drawing large maps.

craftworkgames:
What's the purpose of these changes?  Can you provide an example usage?

Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
cra0zy: Read the description, this is just a quick fix for me, plus I didn't bother setting up the rendering code since what I'm working on is using 1 map per screen that's always fully visible.

cra0zy:
> What's the purpose of these changes? Can you provide an example usage?

Drawing using my own camera system.

craftworkgames:
> Drawing using my own camera system.

Okay, but why would I accept these changes into the library. You haven't really done anything. I'm just going to have to refactor the code and change almost all of it anyway.

What I'm trying to understand is what the goal is? Do you want to be able to use your own `SpriteBatch` instead of having the layer create on for you?

cra0zy:
> What I'm trying to understand is what the goal is? Do you want to be able to use your own SpriteBatch instead of having the layer create on for you?

Yes, plus I don't want to use MG Extended camera system, it would suck for what I'm doing.

craftworkgames:
Okay well let's raise an issue about it and get the implementation right before it gets merged into the library.

Btw.. I won't be accepting pull requests into `master` unless they are very minor bug fixes or complete features. We need to keep `master` in a releasable state #105

Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
craftworkgames: > Read the description, this is just a quick fix for me

I can't accept pull requests like this. If I did, I'd spend all my time fixing up other people's code. I get that it's just a quick fix for you but quick fixes for you means more work for me. I can't leave code like this in the library.

I'm not saying this isn't a good idea. I think adding a couple more `TiledMap.Draw` overloads is a good thing to do but submitting a quick and dirty copy and paste job isn't the way to go about it. 

I mean, I could write this code myself in less time than I'm spending talking about it. I've raised [an issue](https://github.com/craftworkgames/MonoGame.Extended/issues/111) so we can get it done properly.

LithiumToast:
> Okay well let's raise an issue about it and get the implementation right before it gets merged into the library.

I just want to say that I love the commitment being taken for communication.

craftworkgames:
> I just want to say that I love the commitment being taken for communication.

Thanks mate. I'm doing my best. I quickly learned that running an open source project is a lot more than shoving some code into a repo. If I could I would work on it all day every. There's still loads we can do to make the project better. Small steps.

PR #113, opened 2016-02-05
[Tiled] Use transparent color if no color set

cra0zy:


craftworkgames:
What option in the Tiled Map editor does this align with?

Also, I've created a new branch called `tiled-maps`. Can you please submit PR's into that branch instead of `master`.

craftworkgames:
Unfortunately it seems there's no simple way of [changing the target branch](https://github.com/isaacs/github/issues/18)

cra0zy:
> What option in the Tiled Map editor does this align with?

You have a bool that asks if it should use a background color, if it's false, it should choose the Transparent color since no color has been specified: https://github.com/cra0zy/MonoGame.Extended/blob/tiletransp/Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs#L108

> Also, I've created a new branch called tiled-maps. Can you please submit PR's into that branch instead of master.

Shouldn't this go to master since it's just a minor bugfix?

craftworkgames:
>  it should choose the Transparent color since no color has been specified

It used to work without setting the color to transparent. What does it do now? I assume it goes black or something? The bug was probably introduced when I started rendering the map to a `RenderTarget2D`. 

> Shouldn't this go to master since it's just a minor bugfix?

I'll accept it into `master` this time but I'm trying to work out a way to keep `master` clean. Would you be in favor of having a `development` branch as the default? Sorry, I'm still trying to work all this stuff out.

The real problem is testing. I've started adding demos for each feature so that we at least have a way to test these changes as they come in. There's only so much I can tell from looking at the code. Later tonight I'm going to work on the `tiled-maps` branch and add a proper tiled demo project to it. Eventually I want to have demos for every feature.

cra0zy:
> Would you be in favor of having a development branch as the default?

Yes, I would. If you do create it, name it `develop` since all repos name it that way.

craftworkgames:
Done. I've created the `develop` branch and set it as the default.
Thanks.

craftworkgames:
@cra0zy I tried testing this change tonight it wasn't working.
I pulled apart the code for hours until I finally figured out that I had to set the `RenderTargetUsage` to `PreserveContents` from [this question](http://gamedev.stackexchange.com/questions/83822/setrendertargetnull-clearing-previous-draw-calls)

    _graphicsDevice.PresentationParameters.RenderTargetUsage = RenderTargetUsage.PreserveContents;

Just wondering if you're also doing this somewhere in your code?

cra0zy:
This is what I was doing minimized:
```c#
spriteBatch.Begin(transformMatrix: scaleMatrix);
GameContent.Map.Map001.Draw(new MonoGame.Extended.Camera2D(GraphicsDevice));
spriteBatch.End();
```

On the question why am I calling map draw after spriteBatch draw, this was not in 1 file, map drawing was done by the level object, which was rendered by levelmanager, which was rendered my the main game so I kind of forgot...

PR #114, opened 2016-02-05
Fix ViewMatrix bug with BoxingViewportAdapter

rafaelalmeidatk:
Fixed: when using a BoxingViewportAdapter, the tiles are not drawn correctly: http://i.imgur.com/2uWX1fW.gif

### DIFF
--- a/Source/MonoGame.Extended/Camera2D.cs
+++ b/Source/MonoGame.Extended/Camera2D.cs
@@ -137,7 +137,17 @@ namespace MonoGame.Extended
                 Matrix.CreateRotationZ(Rotation) *
                 Matrix.CreateScale(Zoom, Zoom, 1) *
                 Matrix.CreateTranslation(new Vector3(Origin, 0.0f)) *
-                _viewportAdapter.GetScaleMatrix(); 
+                _viewportAdapter.GetScaleMatrix();
+        }
+
+        public Matrix GetViewMatrixWithoutViewportAdapter()
+        {
+            return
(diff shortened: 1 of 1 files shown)


craftworkgames:
Interesting bug. How did you determine that removing the viewport adapter from the view matrix was the solution?

My gut is telling me this isn't the right fix. I'd like to understand the problem better before we settle on a solution.

Rest assured though, we'll definitely fix this.

rafaelalmeidatk:
The tiles weren't draw and when I walk through the map, they were, so I thought that its related to the "visible area" to draw the tiles. I just removed the viewport adapter and it worked, no scientific base here :P

craftworkgames:
> I thought that its related to the "visible area" to draw the tiles.

Yep. I agree it's probably the visible area calculation. Maybe that's not working correctly. I'll have a think about it and take a closer look later this afternoon.

craftworkgames:
I just had an idea. I've created a branch called `tiled-maps` to pull these changes into. I don't want to pull them directly into `master` until they've been tested properly. In fact, I'm going to try keep `master` clean of any future PR's. Maybe create a `development` branch as the default.

Unfortunately, it seem github has no easy way to [change the target branch of a PR](https://github.com/isaacs/github/issues/18) despite a lot of demand for it. I'm still working through the kinks of how to manage these, sorry.

rafaelalmeidatk:
I can close this PR and do another on the `tiled-maps` branch.

craftworkgames:
> I can close this PR and do another on the tiled-maps branch.

Yes please. Sorry for the hassle.

PR #123, opened 2016-02-10
Raise a exception if the tileset file size is not multiple of the tile …

rafaelalmeidatk:
Raise an exception if the tileset file size is not multiple of the tile size


craftworkgames:
Looks good but can you change the exception to something more specific. InvalidOperationException perhaps.

I read the other day that if you change and recommit the PR gets updated automatically.

rafaelalmeidatk:
Yep, it is true. Done :P
Let me know if you think in a better sentence to the exception.

craftworkgames:
Thanks. I'm going to work on it a bit tonight so I'll fix the message up slightly.

PR #131, opened 2016-02-28
Tiled Improvements

cra0zy:
Stuff done:
 - new overrides for spriteBatch
 - added Visible property to TileLayer
 - made map drawing code for spriteBatch actually use the spriteBatch that you give them instead of creating a new one... what would the point of giving a spriteBatch if you are just gonna render a renderTarget onto it...
 - TiledLayer no longer has it's own spriteBatch, but instead uses the one given to it from TiledMap, PS. It's not smart calling spriteBatch.Begin thousands of times for no good reason, looses performance.

I haven't tested the code with the Camera, I write stuff on Linux, and your demos are Windows only... 

Please think about using the same coding policies as MonoGame, or at least use the same coding policies as Microsoft or Mono...

### DIFF
--- a/Source/MonoGame.Extended/Maps/Tiled/SpriteBatchExtensions.cs
+++ b/Source/MonoGame.Extended/Maps/Tiled/SpriteBatchExtensions.cs
@@ -10,9 +10,28 @@ namespace MonoGame.Extended.Maps.Tiled
             tiledMap.Draw(spriteBatch, camera, useMapBackgroundColor);
         }
 
+        public static void Draw(this SpriteBatch spriteBatch, TiledTileLayer tileLayer, Rectangle? visibleRectangle = null)
+        {
+            visibleRectangle = visibleRectangle ?? new Rectangle (0, 0, tileLayer.Width * tileLayer.TileWidth, tileLayer.Height * tileLayer.TileHeight);
+
+            tileLayer.Draw(spriteBatch, (Rectangle)visibleRectangle);
+        }
+
         public static void Draw(this SpriteBatch spriteBatch, TiledTileLayer tileLayer, Rectangle visibleRectangle)
         {
             tileLayer.Draw(spriteBatch, visibleRectangle);
+        }
+
+        public static void Draw(this SpriteBatch spriteBatch, TiledMap tileMap, Rectangle? visibleRectangle = null)
+        {
+            visibleRectangle = visibleRectangle ?? new Rectangle (0, 0, tileMap.Width * tileMap.TileWidth, tileMap.Height * tileMap.TileHeight);
+
+            tileMap.Draw(spriteBatch, (Rectangle)visibleRectangle);
+        }
+
+        public static void Draw(this SpriteBatch spriteBatch, TiledMap tileMap, Rectangle visibleRectangle)
+        {
+            tileMap.Draw(spriteBatch, visibleRectangle);
         }
     }
 }
--- a/Source/MonoGame.Extended/Maps/Tiled/TiledImageLayer.cs
+++ b/Source/MonoGame.Extended/Maps/Tiled/TiledImageLayer.cs
@@ -8,29 +8,23 @@ namespace MonoGame.Extended.Maps.Tiled
     public class TiledImageLayer : TiledLayer, IMovable
     {
         private readonly Texture2D _texture;
-        private readonly SpriteBatch _spriteBatch;
 
-        public TiledImageLayer(GraphicsDevice graphicsDevice, string name, Texture2D texture, Vector2 position)
+        public TiledImageLayer(string name, Texture2D texture, Vector2 position)
             : base(name)
         {
             Position = position;
 
             _texture = texture;
-            _spriteBatch = new SpriteBatch(graphicsDevice);
-        }
-
-        public override void Dispose()
-        {
-            _spriteBatch.Dispose();
         }
 
         public Vector2 Position { get; set; }
 
-        public override void Draw(Rectangle visibleRectangle)
+        public override void Draw(SpriteBatch spriteBatch, Rectangle visibleRectangle)
         {
-            _spriteBatch.Begin(SpriteSortMode.Immediate, BlendState.AlphaBlend, SamplerState.PointClamp);
-            _spriteBatch.Draw(_texture, Position, Color.White);
-            _spriteBatch.End();
+            if (!Visible)
+                return;
+
+            spriteBatch.Draw(_texture, Position, Color.White);
         }
     }
 }
--- a/Source/MonoGame.Extended/Maps/Tiled/TiledLayer.cs
+++ b/Source/MonoGame.Extended/Maps/Tiled/TiledLayer.cs
@@ -1,21 +1,23 @@
 using System;
+
 using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
 
 namespace MonoGame.Extended.Maps.Tiled
 {
-    public abstract class TiledLayer : IDisposable
+    public abstract class TiledLayer
     {
         protected TiledLayer(string name)
         {
             Name = name;
+            Visible = true;
             Properties = new TiledProperties();
         }
 
-        public abstract void Dispose();
-
         public string Name { get; private set; }
+        public bool Visible { get; set; }
         public TiledProperties Properties { get; private set; }
 
-        public abstract void Draw(Rectangle visibleRectangle);
+        public abstract void Draw(SpriteBatch spriteBatch, Rectangle visibleRectangle);
     }
 }
--- a/Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
+++ b/Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
@@ -8,7 +8,7 @@ using MonoGame.Extended.TextureAtlases;
 
 namespace MonoGame.Extended.Maps.Tiled
 {
-    public class TiledMap : IDisposable
+    public class TiledMap
     {
         public TiledMap(GraphicsDevice graphicsDevice, int width, int height, int tileWidth, int tileHeight, 
             TiledMapOrientation orientation = TiledMapOrientation.Orthogonal)
@@ -25,12 +25,6 @@ namespace MonoGame.Extended.Maps.Tiled
             TileHeight = tileHeight;
             Properties = new TiledProperties();
             Orientation = orientation;
-        }
-        
-        public void Dispose()
-        {
-            foreach (var tiledLayer in _layers)
-                tiledLayer.Dispose();
         }
 
         private readonly List<TiledTileset> _tilesets;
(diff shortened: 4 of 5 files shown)


craftworkgames:
I've already implemented most of these things. I'll have a look later and see if you did anything differently.

> Please think about using the same coding policies as MonoGame, or at least use the same coding policies as Microsoft or Mono...

What policies are those? What don't you like about my code?

craftworkgames:
> what would the point of giving a spriteBatch if you are just gonna render a renderTarget onto it...

Actually, there is a point of using the renderTarget inside the layers. Rendering the tiles individually doesn't work when you scale the scene with the camera. You end up with little gaps between the tiles. By removing the render target from the layers you've re-introduced an old bug.

> made map drawing code for spriteBatch actually use the spriteBatch that you give them instead of creating a new one...

This indeed does look like an oversight. The layer code should use the spriteBatch you give it. However, what you've done doesn't fix the problem either. It will need to be fixed without removing the render target code as mentioned above.

> TiledLayer no longer has it's own spriteBatch, but instead uses the one given to it from TiledMap, PS. It's not smart calling spriteBatch.Begin thousands of times for no good reason, looses performance.

First of all, you're exaggerating. Nobody is calling `spriteBatch.Begin` thousands of times. Even if you had 10 layers in your map (which is unlikely) you'd only be calling `spriteBatch.Begin` 10 times per frame. This is *far* from thousands and wouldn't impact performance very much.

Secondly, to render to the renderTarget correctly we need to be able to control the parameters passed into `spriteBatch.Begin`. If we don't there could be a transformation matrix applied and the tiles won't fit onto the render target surface.

Third, while I agree this code has room for a couple of optimizations, we need to make sure we don't break anything in the process. In particular, the user should be able to render the layers individually without experiencing gaps between the tiles.

> I haven't tested the code with the Camera, I write stuff on Linux, and your demos are Windows only...

I understand that running the demos on Linux is a pain. Hopefully the soon to be released MonoGame 3.5 will make that situation better. However, by not running the demos you haven't tested your code properly. Had you have run them you would have noticed that the code is broken in two ways.

 1. Running the demo throws an `InvalidOperationException`.

    > Begin cannot be called again until End has been successfully called.

 2. There are gaps between the tiles when the screen is scaled or the camera is zoomed.

image

cra0zy:
> Actually, there is a point of using the renderTarget inside the layers. Rendering the tiles individually doesn't work when you scale the scene with the camera. You end up with little gaps between the tiles. By removing the render target from the layers you've re-introduced an old bug.

What I was talking about isn't what is the use of renderTarget, what I was talking about its there is no point for MG.Extended to draw onto renderTarget, you should leave that to the user if it's getting called from spriteBatch.

You must understand that people who are using MG are using it for the low end stuff they can do.



> First of all, you're exaggerating. Nobody is calling spriteBatch.Begin thousands of times. Even if you had 10 layers in your map (which is unlikely) you'd only be calling spriteBatch.Begin 10 times per frame. This is far from thousands and wouldn't impact performance very much.
> 
> Secondly, to render to the renderTarget correctly we need to be able to control the parameters passed into spriteBatch.Begin. If we don't there could be a transformation matrix applied and the tiles won't fit onto the render target surface.
> 
> Third, while I agree this code has room for a couple of optimizations, we need to make sure we don't break anything in the process. In particular, the user should be able to render the layers individually without experiencing gaps between the tiles.

Every tiledLayer in tiledMap is getting the same position...

> I understand that running the demos on Linux is a pain. Hopefully the soon to be released MonoGame 3.5 will make that situation better. However, by not running the demos you haven't tested your code properly. Had you have run them you would have noticed that the code is broken in two ways.

Not interested in fixing those demos.

> Running the demo throws an InvalidOperationException.

Not broken, it's now required for the user to call spriteBatch.Begin before using spriteBatch method... first paragraph... 

> There are gaps between the tiles when the screen is scaled or the camera is zoomed.

In case user is drawing from the spriteBatch code, that is totally normal. In case it's getting drawn from camera code, then that shouldn't be happening.

craftworkgames:
> there is no point for MG.Extended to draw onto renderTarget, you should leave that to the user if it's getting called from spriteBatch.

If it's not drawn to a renderTarget there will be gaps between the tiles. We *could* leave it to the user, but I doubt anyone wants gaps between their tiles. Not many people will understand why the gaps are there or how to fix them.

> You must understand that people who are using MG are using it for the low end stuff they can do.

Are you kidding? I'm not stopping people from doing low end stuff. I'm helping people by giving them *one way* to render maps without having to write all the code themselves. If they don't want to use it they can do it however low end they like.

> Every tiledLayer in tiledMap is getting the same position...

Irrelevant.

> Not interested in fixing those demos.

If you're not willing to test your code properly I can't accept the PR. 

> Not broken, it's now required for the user to call spriteBatch.Begin before using spriteBatch method... first paragraph...

Oh, you mean like this:

    _spriteBatch.Begin(transformMatrix: _camera.GetViewMatrix());
    _spriteBatch.Draw(_tiledMap, _camera);
    _spriteBatch.End();

Did you even bother to look at the code in the demo? There's nothing wrong with code in the demo. As I said, I already added most of these spriteBatch overrides before you submitted this PR.

> In case user is drawing from the spriteBatch code, that is totally normal. In case it's getting drawn from camera code, then that shouldn't be happening.

Gaps are not "normal" *ever*. And yes, it is happening from camera code.

As usual, I've spent half my night testing your code only to find out it doesn't work and arguing with you about it. I really don't like wasting my time like this when I could be doing more important things.

I do like the idea of having a Visible flag on tiled layers and some ideas you had with the spriteBatch overloads are good too. I'll make sure these get added to the library but I won't be accepting this PR in it's broken state.

LithiumToast:
One idea instead of using render targets is to create the geometry of each tile layer using primitives without batching (static vertex buffer & static index buffer). This would only work best though if the tile layer data doesn't change frequently. For dynamic things in the map, even "tiles" which disappear or appear frequently, object layers could be used.

LithiumToast:
Just a note, but taking a look at [`SpriteBatch` source code](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/SpriteBatch.cs) shows what is going on in `SpriteBatch.Begin`. It's not doing anything fancy... The performance problem is likely going be in `SpriteBatch.End`.

craftworkgames:
Okay, so I spend some time tonight incorporating these changes properly. 
 - `SpriteBatch` overloads for layers
 - An `IsVisible` property on each layer
 - Both rendering methods tested with the camera at different zoom levels (no gaps)
 - The entire map or individual layers can be rendered with your own sprite batch

Internally the each layer is drawn to a render target first without scaling to prevent gaps between the tiles. When the render target is actually rendered to the `spriteBatch` passed in the transformations are applied. This gives the correct behavior in all tested situations.

            GraphicsDevice.Clear(_tiledMap.BackgroundColor ?? Color.Black);

            _spriteBatch.Begin(transformMatrix: _camera.GetViewMatrix());

            // you can draw the whole map all at once
            //_spriteBatch.Draw(_tiledMap, _camera);

            // or you can have more control over drawing each individual layer
            foreach (var layer in _tiledMap.Layers)
            {
                _spriteBatch.Draw(_sprite);
                _spriteBatch.Draw(layer, _camera);
            }

            _spriteBatch.End();

Note, I've removed the background color from the maps Draw call. If you want to use the maps background color you can do so as shown above in the `GraphicsDevice.Clear` call. This gives more control over when and how to render the background color.

@cra0zy Thanks for your efforts on this. Once I had time to think about it I realized the intention of your changes. Sorry for arguing with you yesterday. It's just extremely frustrating when PR's are not tested properly and have really obvious bugs.

cra0zy:
You still missed the point of what is needed for spriteBatch drawing... on the other hand Nez Framework has done it perfectly... except for Java style of it.

At the end, I wrote my own tiled pipeline stuff... no dependency to anyone or anything.

PS. I didn't know at the time I was writing this, but it seems that there is "visible" property for layer in the tmx file http://doc.mapeditor.org/reference/tmx-map-format/, I only figured that out while I was writing my own pipeline stuff.

craftworkgames:
> You still missed the point of what is needed for spriteBatch drawing... on the other hand Nez Framework has done it perfectly... except for Java style of it.

Maybe I did miss the point, as usual you still haven't explained it. 
 - You told me what changes you made but not why you made them.
 - You told me that you didn't bother testing your code with the camera. 
 - You told me that you didn't bother running your code on the demos. 
 - You told me that you don't like my coding policies, but you didn't tell me what you don't like about them.
 - You told me what you think MG users want (without any evidence).
 - You told me that calling spriteBatch.Begin too many times will impact performance but you never bothered to actually measure the performance impact.

At no point did you actually explain *why* you did any of these things. In my opinion, your communication skills could use some improvement.

The fact remains your PR has some significant bugs and I can't accept it into the library. I don't expect PR's to be perfect but I do expect some level of testing.

Nez certainly has some nice additions to the renderer but I suspect it still fails to draw tiles without gaps when the camera is zoomed.

> At the end, I wrote my own tiled pipeline stuff... no dependency to anyone or anything.

I don't have a problem with people writing their own stuff. I never expected my library to suit every situation. Whatever works for you is fine with me. It may not fit into MonoGame.Extended.

> it seems that there is "visible" property for layer in the tmx file

Yep. I've read the spec. Our importer reads the visible property from the TMX file. I've just never got around to connecting it to the renderer. I guess I'll do that now.

cra0zy:
> You told me that you didn't bother testing your code with the camera.

Don't know how to use your camera.

> You told me that you didn't bother running your code on the demos.

Your demos are Windows only... and I don't have a Windows installed anywhere... not interested in getting them up and running on Linux...

> You told me that you don't like my coding policies, but you didn't tell me what you don't like about them.

Arrangement of code. Just look at some repo under Microsoft or Mono...

> You told me what you think MG users want (without any evidence).

Why would be using any XNA clone if not for doing low level stuff? Why would you not provide low level stuff? It's just common sense...

> You told me that calling spriteBatch.Begin too many times will impact performance but you never bothered to actually measure the performance impact.

Well yea... it's just a common fact that calling spriteBatch Begin/End too many time is bad... I never measured it myself, I just know about it... I only told you because I thought you might be interested about it...

craftworkgames:
> Don't know how to use your camera.

That's what the demos are for. You don't need to know anything about how to use the camera to be able to copy and paste some code from the demos.

Regardless, it doesn't change the fact that your PR doesn't work with the camera. How can I possibly accept this into the library? You made changes without considering how those changes affect the rest of the library. Making the performance *possibly* marginally better (without any proof) is not a good enough reason to break the camera.

> Your demos are Windows only... and I don't have a Windows installed anywhere... not interested in getting them up and running on Linux...

The only thing that ties the demos to Windows as far as I'm aware is the `/platform:WindowsGL` flag in the `Content.mgcb` file. It's my understanding this will be fixed in the MonoGame 3.5 release. Of course I don't have a Linux machine to test them on so I could be completely wrong about this.

I am aware that it's more effort getting the demos working on Linux. One day I would like to setup a Linux machine to support it properly. Until then there's really nothing I can do about it. If I can't test them I can't be sure they work.

> Why would be using any XNA clone if not for doing low level stuff? 

As far as I'm aware, people are using XNA / MonoGame to actually make *games*.

> Well yea... it's just a common fact that calling spriteBatch Begin/End too many time is bad... I never measured it myself, I just know about it...

I have measured it myself. It's not as bad as you think. Remember, we are not calling Begin/End for every *tile*, we are calling it for every *layer*. That's not many times in a single Draw call.

cra0zy:
> As far as I'm aware, people are using XNA / MonoGame to actually make games.

I was already talking about making games... if you use Unity3D, you will use it's camera system, it's object system, etc. If you are making a game in XNA, you will write your own camera system, write your own level system, etc. What did you think I was talking about?

craftworkgames:
> If you are making a game in XNA, you will write your own camera system, write your own level system, etc. What did you think I was talking about?

MonoGame.Extended is about providing implementations of those things so that you don't *have* to write your own. The goal is to make it easier and faster to actually make games.

MonoGame is a low level framework. Unity3D is a high level game engine. MonoGame.Extended falls somewhere in the middle.

PR #133, opened 2016-03-04
Remove DotNetZip dependancy in favor of .NET 4.5 DeflateStream. 

kreynes:
Removed the DotNetZip dependancy in favor of .NET 4.5's DeflateStream. Since .NET 4.5, DeflateStream is using zlib. [MSDN / DeflateStream](https://msdn.microsoft.com/en-us/library/system.io.compression.deflatestream%28v=vs.100%29.aspx#Anchor_5).

Configured the solution to target .NET 4.5 including the portable library by switching to Profile 24 (portable-net45 + sl50).

Haven't had the chance to test it, since I upgraded the MonoGame devbuild and Xamarin Studio, and I'm missing some symlinks that prevent me from building the solution.

### DIFF
--- a/Source/MonoGame.Extended.Content.Pipeline/MonoGame.Extended.Content.Pipeline.csproj
+++ b/Source/MonoGame.Extended.Content.Pipeline/MonoGame.Extended.Content.Pipeline.csproj
@@ -31,10 +31,6 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Ionic.Zip, Version=1.9.6.0, Culture=neutral, PublicKeyToken=6583c7c814667745, processorArchitecture=MSIL">
-      <HintPath>..\packages\DotNetZip.1.9.6\lib\net20\Ionic.Zip.dll</HintPath>
-      <Private>True</Private>
-    </Reference>
     <Reference Include="MonoGame.Framework, Version=3.1.2.0, Culture=neutral, processorArchitecture=MSIL">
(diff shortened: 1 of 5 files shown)


kreynes:
Closed the pull request at the moment. After further research it appears that the .NET 4.5 Implementation of DeflateStream is riddled with bugs that the developers don't want to fix. Should probably stick to DotNetZip.

craftworkgames:
Yep. So there's a couple of things here.

First, I'm not really clear on why you wanted to remove DotNetZip. What problem are we trying to solve.

Second, obviously splitting pull requests is a good idea if you want to be able to merge some changes separate from others. In this case it probably doesn't matter to much but just something to think about for future.

Lastly, I don't think you need to close PRs if they are not right. You can just submit another commit and it adds it to the existing PR. That way we can review and fix before it gets merged into the develop branch.

kreynes:
@craftworkgames Well, getting rid of external dependancies should be a good thing, i suppoue?
and I closed the PR because apparently, i misunderstood the way DeflateStream works. So as far as I got, It wont actually solve the problem I was hoping it will.

craftworkgames:
> It wont actually solve the problem I was hoping it will.

The problem you hoped it would solve was getting rid of external dependencies?

> I closed the PR because apparently, i misunderstood the way DeflateStream works

Fair enough but it still might be worth changing to .NET 4.5 for other reasons.

LithiumToast:
> Fair enough but it still might be worth changing to .NET 4.5 for other reasons.

@craftworkgames Yes please.

PR #140, opened 2016-03-12
3.5: update MonoGame.Framework PCL & `unsafe`

LithiumToast:
- [x] Used Piranha to build and replace the `MonoGame.Framework.dll` PCL for 3.5. (I need `GraphicsDevice.SetVertexBuffer` overload not present in the 3.4 PCL)
~~- [x] Used Piranha to build and add the `MonoGame.Framework.Content.Pipeline.dll` PCL for 3.5.~~
- [x] Changed PCL target profile to everything except Silverlight; this allows for `unsafe` code to be enabled for the PCL which will important for `PrimitiveBatch` and collision detection / response.
- [x] Fixed unit content pipeline unit tests that were having problem referencing files in the bin folder.
~~- [x] Added simple `unsafe` unit test just demonstrating that unsafe code is not that scary.~~
- [x] Built and ran all unit tests; they all succeed.
- [x] Removed DotNetZip library; MonoGame 3.5 has zlib support built in.


WardBenjamin:
Looks good! Don't you love Piranha? :gem:

LithiumToast:
@WardBenjamin Yup, I'm updating the wiki on how to build it for MonoGame and MonoGame.Extended right now.

WardBenjamin:
@LithiumToast I'd appreciate a ping when you're done so I can read it! :D

LithiumToast:
@WardBenjamin [It's done](https://github.com/craftworkgames/MonoGame.Extended/wiki/How-to-build-the-MonoGame-PCL).

WardBenjamin:
@LithiumToast Just read it, looks good! I think we should probably link to the 3.4 and 3.5 releases when the happen so that they're easy to find from the article. I'll look into that once 3.5 is released.

LithiumToast:
@WardBenjamin It's the same process to build the PCL for 3.4 and 3.5 though.

WardBenjamin:
@LithiumToast Let me revise that a little: I think we should add links to the *sources* of those releases in the #mono/MonoGame repo, aka [here](https://github.com/mono/MonoGame/releases). I'm not saying to change the rest of the article at all, I just think that those would be helpful to people reading the wiki page.

LithiumToast:
@WardBenjamin Oh, well of course. It be _wonderful_, just _awesome_, if MonoGame used Piranha with their build server too so that they could automate the NuGet packages to build a PCL for `MonoGame.Framework.dll` and `MonoGame.Framework.Content.Pipeline.dll`. Then people just need to install MonoGame for their platform and use the NuGet package to get the PCL in all their projects.

WardBenjamin:
Agreed!

LithiumToast:
@craftworkgames Don't merge this yet, there is a problem with `DebugDisplayString` on certain structs.

@WardBenjamin How did you get Piranha to not remove the DebugDisplayString method? I'm getting this in the console output:
```
RemovePrivateMembers: Removing method System.String Microsoft.Xna.Framework.Vector2::get_DebugDisplayString().
```

But when debugging, the debugger looks for that method since the attribute is applied to the struct :cry:

LithiumToast:
I tried making the MonoGame.Extended.Content.Pipeline portable but it would require a PCL third-party file storage library which adds an extra mandatory reference for every content project. So I don't see the point of forcing people to do that just to make the MonoGame.Extended.Content.Pipeline portable...

craftworkgames:
> I tried making the MonoGame.Extended.Content.Pipeline portable

I don't think that's ever going to happen. The Pipleline is always going to carry dependencies on libraries that will be difficult or impossible to build as a PCL. 

I'm also not sure if there's any benefit in having a PCL of the Pipeline DLL anyway because it should work on all platforms supported by the Pipeline tool as it is? I could be wrong about that.

Source/MonoGame.Extended.Content.Pipeline/ContentWriterExtensions.cs
```diff
@@ -0,0 +1,79 @@
+﻿using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Content.Pipeline.Serialization.Compiler;
+
+namespace MonoGame.Extended.Content.Pipeline
+{
+    public static class ContentWriterExtensions
```
craftworkgames: Good idea.

Source/MonoGame.Extended.Tests/UnsafeCodeTests.cs
```diff
@@ -0,0 +1,72 @@
...
+            var stringBuilder = new StringBuilder();
+            stringBuilder.Append("The first 10 elements are: ");
+            for (var i = 0; i < 10; ++i)
+            {
+                stringBuilder.Append(b[i] + " ");
+            }
+
+            Assert.AreEqual(stringBuilder.ToString(), "The first 10 elements are: 0 1 2 3 4 5 6 7 8 9 ");
+        }
+
```
craftworkgames: What's with the UnsafeCopy stuff? I'm not clear on why we need it?

Source/MonoGame.Extended.Content.Pipeline.Tests/TiledMapImporterProcessorTests.cs
```diff
@@ -12,14 +14,14 @@ public class TiledMapImporterProcessorTests
         [Test]
         public void TiledMapImporter_Import_Test()
         {
-            const string filename = @"TestData\level01.tmx";
+            var filePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, @"TestData\level01.tmx");
```
craftworkgames: instead of repeating Path.Comine everywhere maybe this can be a method.

Source/MonoGame.Extended.Tests/UnsafeCodeTests.cs
LithiumToast: It's just a demonstration of `unsafe` usage. We don't really need it. I can remove it if you want.

Source/MonoGame.Extended.Content.Pipeline.Tests/TiledMapImporterProcessorTests.cs
LithiumToast: Sure.

Source/MonoGame.Extended.Tests/UnsafeCodeTests.cs
craftworkgames: If we don't need it, we don't need it. When / if we do need it we'll add it then.

craftworkgames:
> if MonoGame used Piranha with their build server too so that they could automate the NuGet packages to build a PCL

If the MonoGame team doesn't do this maybe we should do it on our [build server](http://build.craftworkgames.com/)!  We could provide periodic "unofficial" MonoGame releases to people maybe.

PR #142, opened 2016-03-16
Object Pool

LithiumToast:
Initial version of object pooling. See http://gameprogrammingpatterns.com/object-pool.html object pooling pattern.

This code is based off of: http://www.codeproject.com/Articles/535735/Implementing-a-Generic-Object-Pool-in-NET

Currently, if asking for an object and there is none available (a.k.a miss) it creates a single new object and returns it. In a future version, this will not be the only way the object pool behaves. The ability to choose and change the way misses are handled will be added later.

I added a single large unit test demonstrating how to use object pooling. It passes.


LithiumToast:
Ops, I used an inappropriate branch name for this. Oh well, I'll just delete the branch after and create a new one.

Source/MonoGame.Extended/Collections/ObjectPool.cs
```diff
@@ -0,0 +1,205 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace MonoGame.Extended.Collections
+{
+    // derived from: http://www.codeproject.com/Articles/535735/Implementing-a-Generic-Object-Pool-in-NET
+    public partial class ObjectPool<T> where T : class, IPoolable
```
craftworkgames: Why do we need to restrict T to class?

Source/MonoGame.Extended/Collections/ObjectPool.Diagnostics.cs
```diff
@@ -0,0 +1,115 @@
+﻿using System.Threading;
+
+namespace MonoGame.Extended.Collections
+{
+    public partial class ObjectPool
+    {
+        public class Diagnostics
+        {
```
craftworkgames: What's the thinking behind having a separate Diagnostics object?

Source/MonoGame.Extended/Collections/ObjectPool.cs
LithiumToast: Well, why would anyone need a pool of structs?

craftworkgames: Good point.

Source/MonoGame.Extended/Collections/ObjectPool.Diagnostics.cs
LithiumToast: It's from the derived code, http://www.codeproject.com/Articles/535735/Implementing-a-Generic-Object-Pool-in-NET.
It's mostly useful for unit testing, but I can for see it being useful in a UI widget for getting a quick snapshot of the state of the pool.

Source/MonoGame.Extended/Collections/ObjectPool.cs
```diff
@@ -0,0 +1,205 @@
...
+            T newObject;
+
+            if (FactoryMethod != null)
+            {
+                newObject = FactoryMethod();
+            }
+            else
+            {
+                newObject = (T)Activator.CreateInstance(typeof(T));
+            }
```
craftworkgames: Looks like it's currently limited to objects that have a parameterless constructor? One way we could get around this is by injecting a creation function `Func<T> createInstanceFunc` maybe.

craftworkgames:
Looks good though. I'll merge it tonight.

Source/MonoGame.Extended/Collections/ObjectPool.cs
LithiumToast: Not at all. The `FactoryMethod` is used to create the object it is not null.

```
var pool = new ObjectPool<MyPoolableType>(myMinSize, myMaxSize, () => new MyPoolableType(myParameter));
```

craftworkgames: Ah I see. `Activator.CreateInstance` is just the fallback.

Source/MonoGame.Extended/Collections/Poolable.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace MonoGame.Extended.Collections
+{
+    public abstract class Poolable : IPoolable
```
craftworkgames: I think I'd rather not have this class and just provide the interface. That way people are forced to think about the `ResetState` method. If they derive from this class and forget to implement ResetState things could get hairy.

LithiumToast: How about making the `ResetState` abstract?

craftworkgames: Sure that would work but wouldn't it pretty much be an interface at that point?

LithiumToast: Well my idea was to extend the `Poolable` with a convience method to return it to the `ObjectPool` instead of doing so directly from the `ObjectPool`.

craftworkgames: The problem with using abstract classes vs interfaces is that a class can only derive from a single base class, but it can implement many interfaces. So in this case, I don't see why there's any benefit to having the abstract class (unless we plan to add to it later?)

craftworkgames: > Well my idea was to extend the Poolable with a convience method to return it to the ObjectPool instead of doing so directly from the ObjectPool.

Ah I see. Well I guess we'll leave it this way for now then.

LithiumToast: Ugh, to do that I need the `IPoolable` to have a reference to the `ObjectPool` but `ObjectPool` is a generic... Screw it, the API is ok as is with the `ObjectPool` to return the `IPoolable`.

craftworkgames: Cool. It's decided then. As I said I'll merge it in about 3 hours when I get home. If you want to make any more changes before then (e.g. delete commented code) otherwise all good.

LithiumToast:
I'll write more unit tests for this later. I have not tested destroying objects yet (not the same thing as returning objects). I also did not test concurrency yet.

craftworkgames:
No worries man. It doesn't have to be perfect right off the bat.

PR #146, opened 2016-03-21
Object pool v2

LithiumToast:
- [x] added `Deque`; double ended queue to `Collections` namespace
- [x] added `Deque` tests
- [x] renamed `ObjectPool` to `Pool`
- [x] `Pool` uses `Deque` internally
- [x] `IPoolable` is given an `Action` which returns it to the `Pool` allowing for an API where the `IPoolable` is given responsibility but still decoupled


craftworkgames:
Good enough.

PR #147, opened 2016-03-22
Object pool tests

LithiumToast:
Sorry I forgot to commit this one file :cry:

### DIFF
--- a/Source/MonoGame.Extended.Tests/Collections/PoolTests.cs
+++ b/Source/MonoGame.Extended.Tests/Collections/PoolTests.cs
@@ -8,7 +8,7 @@ namespace MonoGame.Extended.Tests.Collections
     {
         private class TestPoolable : IPoolable
         {
-            private ReturnObjectToPoolDelegate _returnFunction;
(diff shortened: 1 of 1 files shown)


LithiumToast:
So, to use a pooled object:
1. Request a free pooled object; `Pool.Create()`. This calls `IPoolable.Initialize(delegate)`, passing the internal `Pool.Return` method to `IPoolable` through the delegate.
2. Release it; `IPoolable.Return`. It's up the `IPoolable` to call the return delegate, if available. It should return false if the delegate has already been used.

```
        private class TestPoolable : IPoolable
        {
            private ReturnToPoolDelegate _returnFunction;
...
            void IPoolable.Initialize(ReturnToPoolDelegate returnFunction)
            {
                _returnFunction = returnFunction;
            }

            public bool Return()
            {
                if (_returnFunction == null)
                {
                    return false;
                }
                _returnFunction.Invoke(this);
                _returnFunction = null;
                return true;
            }
...
        }
```

All objects are referenced whether they are available or in use, so there is no worry about the garbage collector.

PR #148, opened 2016-03-22
hotfix Deque IndexOf

LithiumToast:
Check if the item is at the front or back of the `Deque` before checking the whole circular array. This is important for `Pool`.


craftworkgames:
The auto PR builds are working great :+1: 

http://build.craftworkgames.com/viewLog.html?buildId=913&tab=buildResultsDiv&buildTypeId=MonoGameExtended_CI

PR #154, opened 2016-04-05
Make FPS counter a GameComponent

LibertyLocked:
I find it confusing to call `fpsCounter.Update()` in a Draw loop.
As a GameComponent, the user doesn't have to worry about forgetting to call `fpsCounter.Update`, or calling Update in the wrong loop.

### DIFF
--- a/Source/Demos/Demo.SpriteSheetAnimations/Game1.cs
+++ b/Source/Demos/Demo.SpriteSheetAnimations/Game1.cs
@@ -28,13 +28,11 @@ namespace Demo.SpriteSheetAnimations
             _graphicsDeviceManager = new GraphicsDeviceManager(this);
             Content.RootDirectory = "Content";
             IsMouseVisible = true;
-            Window.Title = $"MonoGame.Extended - {GetType().Name}";
-            Window.Position = Point.Zero;
         }
 
         protected override void Initialize()
         {
-            _fpsCounter = new FramesPerSecondCounter();
+            Components.Add(_fpsCounter = new FramesPerSecondCounter(this));
             _viewportAdapter = new BoxingViewportAdapter(Window, GraphicsDevice, 800, 480);
             _camera = new Camera2D(_viewportAdapter)
             {
@@ -45,6 +43,8 @@ namespace Demo.SpriteSheetAnimations
                 Position = new Vector2(408, 270)
             };
 
+            Window.Title = $"MonoGame.Extended - {GetType().Name}";
+            Window.Position = Point.Zero;
             Window.AllowUserResizing = true;
 
             base.Initialize();
@@ -126,8 +126,6 @@ namespace Demo.SpriteSheetAnimations
 
         protected override void Draw(GameTime gameTime)
         {
-            _fpsCounter.Update(gameTime);
-
             GraphicsDevice.Clear(Color.CornflowerBlue);
 
             _spriteBatch.Begin(transformMatrix: _camera.GetViewMatrix());
--- a/Source/Demos/Demo.SpriteSheetAnimations/Game1.cs
+++ b/Source/Demos/Demo.SpriteSheetAnimations/Game1.cs
@@ -16,7 +16,7 @@ namespace Demo.SpriteSheetAnimations
         // ReSharper disable once NotAccessedField.Local
         private readonly GraphicsDeviceManager _graphicsDeviceManager;
         private Camera2D _camera;
-        private FramesPerSecondCounter _fpsCounter;
+        private FramesPerSecondCounterComponent _fpsCounter;
         private SpriteBatch _spriteBatch;
         private TiledMap _tiledMap;
         private ViewportAdapter _viewportAdapter;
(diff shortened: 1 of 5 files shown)


craftworkgames:
Yep. That's a good point.

We haven't really worried about GameComponent's because most people don't use them. However, it's been in the back of my mind that we might want to support them one day. You're the first person to specifically ask. I guess now is that time.

I like what you've done but I'd prefer it if we can still support doing it the old way as well.

Source/Demos/Demo.SpriteSheetAnimations/Game1.cs
```diff
@@ -28,13 +28,11 @@ public Game1()
             _graphicsDeviceManager = new GraphicsDeviceManager(this);
             Content.RootDirectory = "Content";
             IsMouseVisible = true;
-            Window.Title = $"MonoGame.Extended - {GetType().Name}";
-            Window.Position = Point.Zero;
```
craftworkgames: Please don't remove this code. It's there because Mac's have some issues with the default window position. I'm not a fan of placing the window in the top left corner either, but it's not hurting anything.

Source/Demos/Demo.SpriteSheetAnimations/Game1.cs
```diff
@@ -45,6 +43,8 @@ protected override void Initialize()
                 Position = new Vector2(408, 270)
             };
 
+            Window.Title = $"MonoGame.Extended - {GetType().Name}";
+            Window.Position = Point.Zero;
```
craftworkgames: Oh, you just moved it down here. Sorry I didn't notice.

Source/MonoGame.Extended/FramesPerSecondCounter.cs
```diff
@@ -4,9 +4,10 @@
 
 namespace MonoGame.Extended
 {
-    public class FramesPerSecondCounter : IUpdate
+    public class FramesPerSecondCounter : DrawableGameComponent
     {
-        public FramesPerSecondCounter(int maximumSamples = 100)
+        public FramesPerSecondCounter(Game game, int maximumSamples = 100)
+            :base(game)
```
craftworkgames: One thing that's always bothered me about game components is the way they take a dependency on the `Game` class for no good reason.

Source/MonoGame.Extended/FramesPerSecondCounter.cs
```diff
@@ -24,7 +25,7 @@ public void Reset()
             _sampleBuffer.Clear();
         }
 
-        public void Update(float deltaTime)
+        public void UpdateFPS(float deltaTime)
```
craftworkgames: Does this have to be called `UpdateFPS`? Can't we just have another overload of `Update`?

Source/MonoGame.Extended/FramesPerSecondCounter.cs
```diff
@@ -4,9 +4,10 @@
 
 namespace MonoGame.Extended
 {
-    public class FramesPerSecondCounter : IUpdate
+    public class FramesPerSecondCounter : DrawableGameComponent
```
craftworkgames: I get that we need to derive from `DrawableGameComponent` but why remove the `IUpdate` interface?

craftworkgames:
Supporting game components is really part of a bigger story.

An alternative to this approach would be to provide a wrapper class called `FramesPerSecondCounterComponent` and leave the existing class untouched.

Source/MonoGame.Extended/FramesPerSecondCounter.cs
LithiumToast: @craftworkgames  We can create a new `GameComponent` class which doesn't take in a `Game` as a parameter to be constructed.

craftworkgames: Cool but if we do that will it still have a `Draw` method to override?

LithiumToast: https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.idrawable.aspx

craftworkgames: Yep. Fair enough.

LibertyLocked:
How about leaving the old FPS counter as is and adding a new one that implements GameComponent?

LibertyLocked:
@craftworkgames The reason I removed `IUpdate` and renaming Update to `UpdateFPS` is that I do not want the user to call Update themselves, and they should know that it's a GameComponent. I guess I forgot to make `UpdateFPS` private. 

I did not think about compatibility.

craftworkgames:
Yeah no worries. We do have to at least consider that these kinds of changes are breaking changes to the API. Sometimes that's unavoidable, but in this case I think we can have the best of both.

Would you like to leave the old FPS counter as is and add a new one that implements GameComponent instead?  You can just update this PR.

LibertyLocked:
Now the `GameComponent` version of FPS counter just forward calls to the `IUpdate` version of FPS counter.

craftworkgames:
> Now the GameComponent version of FPS counter just forward calls to the IUpdate version of FPS counter.

Looks good. At this point you don't even need the `_fpsCounter` variable since it's not used for anything.

@LithiumToast Are you happy with this?

LithiumToast:
@craftworkgames Looks good. Game Components are pretty common for modular systems in MonoGame / XNA. People don't have to add them to `Game.Components` to use them properly either.

craftworkgames:
Okay. I'll merge this later tonight when I get home. We should probably raise an issue about what other game components we should support.

PR #155, opened 2016-04-07
Animations

ExCellRaD:
Started to work on an animator, this is just to make sure I'm heading the right direction (so don't pull yet!). The functionality is currently based on Spine, for easy implementation further down the road. 

Currently in there:

- Transformations (both with or without tweening, separate types) using the interfaces to determine what transformations can be applied (IMoveable...), will need more for stuff like mesh and texture transforms.

- Animation track groups and Animation tracks (group up the transformations per transformable type and  transform type respectively)

- Animation which holds animation tracks

- Animator, GameComponent which holds animations

- Cubic bezier easing for tweening transformations

Also added new shape (PieF), Vertex2D and some crude entity stuff. Let me know if I should leave these out.


ExCellRaD:
adding fluent api #18 support like this:
``` c#
animation.Transform(sprite)
    .Tween(s => s.Alpha, 0.0, 1f)
    .Tween(s => s.Alpha, 100.0, 0f, new CubicBezierEasing(0.5, 1, 1, 0.5))
    .Tween(s => s.Position, 0.0, Vector2.One)
    .Tween(s => s.Position, 100, Vector2.Zero)
    .Set(s => s.Color, 0, Color.Red)
    .MoveTo(0, new Vector2(0, 0))
    .MoveTo(100, new Vector2(10, 10));
```

craftworkgames:
It looks like a good start. Let's call it Tweening so we don't get it confused with other types of animations.

Also, this is a big feature, I'd rather not pull it directly into `develop`. I've created a new branch called `tweening` if you could use that instead please.

I've got a lot on my plate right now so I won't have time to go over it properly yet. You're welcome to keep working on it though, and once we pull into into the branch I'll take a closer look.

I must admit, working on animations is always fun. I can understand why you took this one on :)

ExCellRaD:
alright, closing this one

PR #158, opened 2016-04-09
Animation Tweening

ExCellRaD:
#155, but for the new branch


craftworkgames:
I pulled this into the branch last night and went to take a look. I noticed the demo doesn't seem to do anything. 

I can tell you've put a lot of work into this but it would be really helpful if you could provide a working demo so I can understand how you think the API should be used. I'm sure I could figure it out given enough time, unfortunately, my time is pretty limited these days.

Thanks.

PR #164, opened 2016-04-11
Tweening

ExCellRaD:
Changed the demo project to use the template, removed the shape stuff.
Also don't pull yet, I'm not done! Still needs a lot of fixes + more EasingFunctions to come


Source/Demos/Demo.Tweening/Game1.cs
```diff
@@ -0,0 +1,88 @@
...
+                .Tween(s => s.Alpha, 3000, 0, new StepEasing(5))
+                .Tween(s => s.Alpha, 3500, 1)
+                .Move(0, new Vector2(100, 100)) //IMoveable
+                .Move(3300, new Vector2(600, 50), new CurveEasing(curve))
+                .Move(6600, new Vector2(50, 400))
+                .Move(8000, new Vector2(100, 100))
+                .Set(s => s.Rotation, 5000, 3.14f) //dont interpolate
+                .Set(s => s.Rotation, 5500, 0f)
+                .Tween(s => s.Color, 0, Color.Red) //colors too
+                .Tween(s => s.Color, 1000, Color.White);
```
craftworkgames: I don't really like the way we are using milliseconds here. Everything else in the library works in seconds as a float.

Source/MonoGame.Extended/Animations/Easing/CurveEasing.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Microsoft.Xna.Framework;
+
+namespace MonoGame.Extended.Animations.Easing
+{
+    public class CurveEasing : EasingFunction
+    {
+        public CurveEasing(Curve curve) {
```
craftworkgames: I don't mean to nitpick but your coding style doesn't fit with the rest of the framework. I find it quite annoying when the brackets are on the same line as the method signature. Obviously I can fix this up after the merge but it's one more thing I'm going to have to do.

Source/Demos/Demo.Tweening/Game1.cs
```diff
@@ -0,0 +1,88 @@
...
+namespace Demo.Tweening
+{
+    public class Game1 : Game
+    {
+        // ReSharper disable once NotAccessedField.Local
+        private readonly GraphicsDeviceManager _graphicsDeviceManager;
+        private SpriteBatch _spriteBatch;
+        private Sprite _sprite;
+        private Camera2D _camera;
+        private Animator _animator;
```
craftworkgames: I'd rather call this something more specific. We've already got a `SpriteSheetAnimator`. What type of animator is this? A `TweeningAnimator` maybe?  Also, I noticed it derives from `GameComponent`. I'd like to talk about having more game components later but I think they should all follow a consistent naming convention. e.g. `TweeningAnimatorComponent` or something.

Source/Demos/Demo.Tweening/Game1.cs
```diff
@@ -0,0 +1,88 @@
...
+                .Move(0, new Vector2(100, 100)) //IMoveable
+                .Move(3300, new Vector2(600, 50), new CurveEasing(curve))
+                .Move(6600, new Vector2(50, 400))
+                .Move(8000, new Vector2(100, 100))
+                .Set(s => s.Rotation, 5000, 3.14f) //dont interpolate
+                .Set(s => s.Rotation, 5500, 0f)
+                .Tween(s => s.Color, 0, Color.Red) //colors too
+                .Tween(s => s.Color, 1000, Color.White);
+            _animator.AddAnimation(animation);
+            _animator.RunAnimation("testanimation", true);
```
craftworkgames: The API is pretty cool but it will be interesting to see how well it works in a real world game situation.

https://www.youtube.com/watch?v=tu7jaK2PXIM&nohtml5=False

craftworkgames:
So I took at closer look at what you've done tonight. I think this will turn out to be a really nice feature of the library. We really need to think about how it will be used in a real game though.

Source/MonoGame.Extended/Animations/Easing/CurveEasing.cs
ExCellRaD: Isn't this possible using source control somehow? I'm not too familiar but I've heard of automatic bracket styling somewhere before.

ExCellRaD:
I made a demo for the easing functions I added, you can look up the controls in the update method of the tweening demo. Bezier easing (what Spine uses) functionality isn't done yet, that will be for after the curve implementations.
These are already 10 functions, some even have customizable properties (also more to be added).

Easing can be stepped, there's In/Out functionality.

Source/Demos/Demo.Tweening/Game1.cs
```diff
@@ -0,0 +1,130 @@
...
+                .Move(8000, new Vector2(100, 100))
+                .Set(s => s.Rotation, 5000, 3.14f) //dont interpolate
+                .Set(s => s.Rotation, 5500, 0f)
+                .Tween(s => s.Color, 0, Color.Red) //colors too
+                .Tween(s => s.Color, 1000, Color.White);
+            _animatorComponent.AddAnimation(animation);
+            _animatorComponent.RunAnimation("testanimation", true);
+
+            var group = new AnimationTrackGroup<Sprite>(_sprite);
+            group.
```
craftworkgames: Err.. you left a syntax error here.

ExCellRaD:
Fixed syntax error, don't worry about the other changes yet.

craftworkgames:
I'm just looking at the code periodically. Obviously, you're still working on it so I'm not going to pull yet.

Before it's finished I'd really like to have a nice demo that shows a bunch of different animations used in a realistic way. I'm not sure if you're interested in doing something like that?

ExCellRaD:
> I'm not sure if you're interested in doing something like that?
I could do that, but I'd rather not until some more systems are implemented (primitive batch, entity system... ) and I could just make a complete demo game out of it. I'll probably add some documentation in the wiki though if that's alright?

craftworkgames:
I was looking through these changes last night.

Some of it is very good but I think we are going to take a different direction with the animations code to keep it in line with existing animations features.

The problem is, there's so many changes that it's not really clear how I'm going to merge the relevant bits back into the develop branch. At this stage, the easiest way seems to be copying and pasting the relevant bits of code into a new branch.

In future I think we're going to have to find a more controlled way of implementing these big features. I don't really know what that is, but what we're doing now isn't really working.

Issue #159, opened 2016-04-09
Proposal: add "bleeding" viewport adapter


emkrcko:
I have implemented what I call a bleeding viewport adapter which is a variation of BoxingViewportAdapter with added benefit of better scaling (well, a bit nicer) to different resolutions. 

The idea behind this viewport adapter is to define "bleed" areas (horizontal and vertical) which could be safely cut off if needed.
For example, you define your world to be 1348x900 pixels and create all graphics for that resolution, but you define a "bleed" of 148x140 (so that's 74px from left and right, and 70px from top and bottom edges) which creates a "safe area" where you put all important, interactive stuff. Outside of the safe area you add content which is not important for gameplay.

These example values (1348x900 world, with 148x140 bleed) allow you to have game which scales at least to 3:2, 16:10, 16:9, 4:3 and 5:3 resolutions without letterboxing (and that's most of mobile/tablet devices out there).

The point of opening this issue is that I think this might be useful to others so maybe it could be added to MonoGame.Extended?

In essence, this behavior could be added to BoxingViewportAdapter since setting bleed to 0 makes it behave exactly the same as BoxingViewportAdapter.

Here's the code:

```cs
using Microsoft.Xna.Framework;
using Microsoft.Xna.Framework.Graphics;

namespace MonoGame.Extended.ViewportAdapters {

    public class BleedingViewportAdapter : ScalingViewportAdapter {

        public int HorizontalBleed { get; private set; }
        public int VerticalBleed { get; private set; }

        public BleedingViewportAdapter(GraphicsDevice graphicsDevice, int virtualWidth, int virtualHeight, int horizontalBleed, int verticalBleed)
            : base(graphicsDevice, virtualWidth, virtualHeight) {
            HorizontalBleed = horizontalBleed;
            VerticalBleed = verticalBleed;
        }

        public BleedingViewportAdapter(GraphicsDevice graphicsDevice, int virtualWidth, int virtualHeight)
            : this(graphicsDevice, virtualWidth, virtualHeight, 0, 0) {
        }

        public override void OnClientSizeChanged() {

            var viewport = GraphicsDevice.Viewport;

            var worldScaleX = (float)viewport.Width / VirtualWidth;
            var worldScaleY = (float)viewport.Height / VirtualHeight;

            var safeScaleX = (float)viewport.Width / (VirtualWidth - HorizontalBleed);
            var safeScaleY = (float)viewport.Height / (VirtualHeight - VerticalBleed);

[...]

PR #166, opened 2016-04-12
Add support for defining bleed areas in BoxingViewportAdapter

raleksandar:
This PR adds support for bleed areas as discussed in #159.

Both constructors are overloaded with additional horizontal/vertical bleed parameters. 
I have also added None value to the BoxingMode enum which is set when no letterboxing (or pillarboxing) occurs.

### DIFF
--- a/Source/Demos/Demo.ViewportAdapters/Game1.cs
+++ b/Source/Demos/Demo.ViewportAdapters/Game1.cs
@@ -48,7 +48,7 @@ namespace Demo.ViewportAdapters
 
             // the boxing viewport adapter uses letterboxing or pillarboxing to maintain aspect ratio
             // it's a little more complicated and needs to listen to the window client size changing event
-            _boxingViewportAdapter = new BoxingViewportAdapter(Window, GraphicsDevice, 800, 480);
+            _boxingViewportAdapter = new BoxingViewportAdapter(Window, GraphicsDevice, 800, 480, 88, 70);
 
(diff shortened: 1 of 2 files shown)


craftworkgames:
Looks good.

Have you tested this with the Viewport adapters demo? Ideally, we should be updating the demos when new features are added.

raleksandar:
Yes I have, it works as before. Meaning no bleed areas defined. I can change the demo to use BoxingViewAdapter with bleed areas if you want. I'll just have to offset text a bit.

raleksandar:
I've updated the demo, please take a look when you can, thanks :)

craftworkgames:
Cool. Well it seems to work, but the demo doesn't really make it clear there's different ways to use the boxing adapter. At the very least I think we need is a way to switch between using bleed and not in the demo.

If you don't want to fiddle around with the demo too much I can pull now and add it to my to-do list to get done later?

raleksandar:
Yeah, I agree, that would be good to have in a demo. I could add that sometimes next week, as soon as I find some free time.

craftworkgames:
> I could add that sometimes next week, as soon as I find some spare time.

Great. As soon as you've done that I'll pull. I know what it's like trying to find spare time :smile: I appreciate it. Thanks.

craftworkgames:
Merged. I'll add it to my to-do list to show the difference between the bleeding and non-bleeding modes. Thanks.

PR #169, opened 2016-04-17
Particle rework

ExCellRaD:
Don't pull yet, this is just to look at the changes. lots of unneeded code for now, most of it will be implemented via the tweening branch already.


Source/Demos/Demo.Particles/Game1.cs
```diff
@@ -93,7 +93,7 @@ private void ParticleInit(TextureRegion2D textureRegion)
             {
                 Emitters = new[]
                 {
-                    new ParticleEmitter(500, TimeSpan.FromSeconds(2.5), Profile.Ring(150f, Profile.CircleRadiation.In))
+                    new ParticleEmitter(500, TimeSpan.FromSeconds(2.5), Profile.Ring(new CircleF(Vector2.Zero, 150f), Profile.CircleRadiation.In))
```
craftworkgames: Why is this better than just passing in a radius?

Source/MonoGame.Extended/Particles/Profiles/Profile.cs
```diff
@@ -1,60 +1,48 @@
...
-            return new LineProfile { Axis = axis, Length = length };
+        public static Profile Line(LineF line) {
+            return new OutlineProfile(line);
         }
 
-        public static Profile Ring(float radius, CircleRadiation radiate)
-        {
-            return new RingProfile { Radius = radius, Radiate = radiate };
+        public static Profile Ring(CircleF circle, CircleRadiation radiate) {
+            return new OutlineProfile(circle, radiate, circle.Center);
```
craftworkgames: Why is constructor injection better here? In most cases I would say constructor injection is a good thing but at some point we are going to want to deserailize these objects from a file and constructors can get in the way.

Source/Demos/Demo.Particles/Game1.cs
ExCellRaD: Oh it's definitely not, I just wanted to preserve the way the demo worked for now. Should've made the circle in the ring method, my bad.

Source/MonoGame.Extended/Particles/Profiles/Profile.cs
ExCellRaD: Just to make sure that the shape is always there before using the profile. Should I keep them as parameterless constructors?

craftworkgames: Actually, the real problem with using shapes in the profiles is that shapes have a position. This conflicts with the fact that the emitter also has a position. For example, what does it mean that you've created a point profile from a `PointF`? Maybe I just don't know where you're going with this.

ExCellRaD: > This conflicts with the fact that the emitter also has a position.

While this is true, this also makes it so that you can have shapes that aren't exactly centered on the emitter, this can be useful for when working with IPathF shapes. It might also allow for an emitter to have multiple profiles spawning the same type of particles from differen positions.

> what does it mean that you've created a point profile from a PointF? 

Nothing more than it not requiring an explicit type for that.

craftworkgames:
Let me be honest here.

When the time comes I'm going to take a good hard look at this branch and try to consider which changes belong in the library and which ones don't. At this stage, without really knowing where you're going with it I'm not completely convinced it's all going to make it in.

Keep in mind the more changes you make, the more difficult my job becomes. The code is going to have to be tested and I need to see that the changes provide real value.

You've obviously got some good ideas. I don't doubt that, but it's your job to convince me that these changes are worth it. If you can create a demo and / or write something that shows how these changes provide value it's going to make my life a lot easier.

craftworkgames:
Oops. I pulled the request by accident.

PR #182, opened 2016-04-27
Support Unicode emoji characters in BitmapFont

dtaylorus:
The .NET Char type currently used to identify a glyph in a BitmapFont is
limited to 16 bits, while some Unicode characters (surrogate pairs,
which include Emoji) require 32 bits.

The fix is to use the Int32 type to represent a Unicode character, and
to parse UTF-16 .NET input strings as 32-bit Unicode code point
sequences.

### DIFF
--- a/Source/MonoGame.Extended/BitmapFonts/BitmapFont.cs
+++ b/Source/MonoGame.Extended/BitmapFonts/BitmapFont.cs
@@ -14,14 +14,30 @@ namespace MonoGame.Extended.BitmapFonts
             LineHeight = lineHeight;
         }
 
-        private readonly Dictionary<char, BitmapFontRegion> _characterMap;
+        private readonly Dictionary<int, BitmapFontRegion> _characterMap;
 
         public int LineHeight { get; private set; }
 
-        public BitmapFontRegion GetCharacterRegion(char character)
+        public BitmapFontRegion GetCharacterRegion(int character)
         {
             BitmapFontRegion region;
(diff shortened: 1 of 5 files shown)


craftworkgames:
Looks good.

Is this a breaking change in any way? 
Do we need to provide an overload that takes a `char` as well?
What about loading the `.fnt` files with the content pipeline? Does that need any changes?

Thanks.

dtaylorus:
This is not a breaking change, since you can pass a `char` type to an `int` parameter without casting. So, I don't believe we need an overload.

The pipeline works fine, too. I processed a `.fnt` file with emoji characters in it and verified that they work in the Demo.BitmapFonts project. I didn't include that in the change since I used Microsoft's copyrighted Segoe UI Emoji font to test, and I'm not sure that's ok to submit.

craftworkgames:
Awesome.

Well it would be nice to find a font we can use in the demo but aside from that I think this PR is complete enough. I'll merge it soon. Thanks.

dtaylorus:
Awesome. Thanks! I've got one more ready after you submit this one.

craftworkgames:
I had to remove some of the parameters in the `DrawString` methods because they didn't work properly.

For example, setting rotation should rotate the whole string, but instead it rotates each individual character.

To maintain backwards compatibility I made one of the methods private and changed the signature of the public one to handle both situations with the additional ability to set `layerDepth`.

It would be nice to handle rotation, scale, effects in the future but for now this will have to do.

Merged. Thanks.

Issue #174, opened 2016-04-24
Suggestion: Interval<T>


guFalcon:
Hi, I'm [Psilo](http://community.monogame.net/users/throbax/activity) the developer of [Throbax TD](http://community.monogame.net/t/throbax-td-windows/6617) and I'd like to share some code with you.

# Introduction
We basically are [three guys](http://www.unterrainer.info/Home#/about) developing a game (which is already on steam early-access by the way) and during the last few years I generated quite a lot of code. A few weeks ago I saw the MonoGame.Extended project and I want to contribute. Our game-engine is quite extensive so we'll do this step-by-step starting with the smaller, more refined parts of it.

The idea is to discuss every feature/extension/structure here so that we get a feeling if you people even want it in your library or if I have to modify it and to which extent.
If I should propose to add things already present in the library, then I'm terribly sorry. I don't want to be rude, I just don't have the time to get perfectly accustomed with MonoGame.Extended right now. That will change over time, promised.
So bring on the good ideas / critique / votes or help.

# `Interval<T>`
When programming a game or, frankly speaking, when programming anything, you will be confronted with intervals.
I remember reaching the point, where I began thinking about a separate class, when I was working on the particle engine. I had about 30 values to track per particle (velocity, starting position, rotation, acceleration...) and each one of them had an upper- and lower boundary. As I'm not paid per line of code, I searched for a way to somehow reduce the length and improve the readability of it (int minNumberOfParticles, maxNumberOfParticles; float minRotation, maxRotation...).
The Interval class is a handy way to produce shorter and more readable code while introducing a new level of consistency and, as I will explain right now, accuracy as well.
 
## Accuracy
We don't always take care about it, but intervals don't only have upper- and lower bounds. They also have the property to include or exclude both bounds as well.
And when we have to take care about that, we almost certainly treat it differently each time.
 
Think about it. What do you mean by: "Pick a number between 0 and 100!".
It could be one of the following:
```
[0,100] …0, 1, 2, 3,…, 98, 99, 100
]0,100] …    1, 2, 3,…, 98, 99, 100
[0,100[  …0, 1, 2, 3,…, 98, 99
]0,100[  …    1, 2, 3,…, 98, 99
```

Most of the time we mean to include both boundaries, but we should be able to override that behavior.
 
## The requirements
So we want a generic class that you may instantiate using
	Only an upper and lower bound (you would use that constructor most of the time) and it would default the treatment of the boundaries as inclusive.
[...]

PR #183, opened 2016-04-27
Add range<T>

ExCellRaD:
This adds `Range<T>` to the core library #174. The random values are left out because those are still in the Tweening branch.

### DIFF
--- a/Source/MonoGame.Extended.Tests/MonoGame.Extended.Tests.csproj
+++ b/Source/MonoGame.Extended.Tests/MonoGame.Extended.Tests.csproj
@@ -56,6 +56,7 @@
     <Compile Include="Collections\DequeTests.cs" />
     <Compile Include="Collections\PoolTests.cs" />
     <Compile Include="MockGameWindow.cs" />
+    <Compile Include="RangeTest.cs" />
     <Compile Include="Shapes\PolygonFTests.cs" />
     <Compile Include="Shapes\CircleTests.cs" />
     <Compile Include="Sprites\SpriteTests.cs" />
--- a/Source/MonoGame.Extended.Tests/RangeTest.cs
+++ b/Source/MonoGame.Extended.Tests/RangeTest.cs
@@ -1,0 +1,57 @@
+﻿using System;
+using Microsoft.Xna.Framework;
+using NUnit.Framework;
+
+namespace MonoGame.Extended.Tests
+{
+    [TestFixture]
+    public class RangeTest
+    {
+        [Test]
+        public void ConstructorTest() {
+            //can pass min < max
+            Assert.DoesNotThrow(() => new Range<int>(10, 100));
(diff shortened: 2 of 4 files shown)


craftworkgames:
Yep. This looks pretty good.

Just one thought... The INTERPOLATOR stuff should *not* be on this class. I know it's currently commented out but I don't think it ever belongs on this class. 

@guFalcon Do you have any other comments?

craftworkgames:
Oh, and this class is a perfect candidate for some unit tests.

ExCellRaD:
No random value method then? I thought it would be something nice to have in the struct itself, but I can see it working the other way around too, where you would use the interpolator only where you want random values. 

>Oh, and this class is a perfect candidate for some unit tests.

I'll get on that

craftworkgames:
> No random value method then? I thought it would be something nice to have in the struct itself, but I can see it working the other way around too

Yeah, it creates a dependency problem. We don't want low level data types depending on things like the `InterpolatorStore` or `FastRand`. Doing it like that tends to lead to a very coupled mess of code that becomes difficult to untangle later.

Think about how it works in the .NET framework. You'd never expect the `int` or `float` types to use the `Random` class internally. On the other hand it's reasonable to think that you can generate random `int`'s or `float`'s by *using* `Random`.

ExCellRaD:
> On the other hand it's reasonable to think that you can generate random int's or float's by using Random.

We could add `T FastRand.NextFromRange<T>(Range<T>)` then? or would an extension method be more appropriate so that you could still call `T Range<T>.RandomValue()` ?

craftworkgames:
> We could add T FastRand.NextFromRange<T>(Range<T>) then?

Something like this would be my gut feeling. There's probably more discussion to be had around the `FastRand` class. Obviously it originated in the particles feature so it might need some more thought when pulling it up into the root namespace.

This PR looks complete enough. I'll merge it soon. Thanks.

guFalcon:
Took a quick look at it and it looks awesome. Nice work ExCellRaD.

One thing though:
Why would you have 2 properties `isDegenerate` and `isProper` where `isProper == !isDegenerate` ?

As on the random-topic I'm with Dylan.

And uuh uuh...
Maybe that's covered somewhere in MG.Ex already but when doing your randomizer, look that you can easily replace the function with some Perlin/Simplex noise. Paid off for some multiplayer-implementations (deterministic lock step) where you'd have to know what values the other players are rolling. Been there ;) But my implementation is clunky.

ExCellRaD:
>Why would you have 2 properties isDegenerate and isProper where isProper == !isDegenerate ?

Why keep one if you can have both? :) They came from the particle Range classes, I just thought I'd keep them.

guFalcon:
*gggg*
np. just asking.

craftworkgames:
Merged. Thanks @ExCellRaD.

Issue #136, opened 2016-03-10
[Tiled] Rendering large maps uses a lot of memory


craftworkgames:
![image](https://cloud.githubusercontent.com/assets/3201643/13553805/c1fd9022-e3e0-11e5-8e79-c1c6bd5c18b9.png)

Currently we are rendering each layer of the *entire* map to a render target and caching that render target in memory for future renders. This gives the best possible frame rate performance at the cost of memory.

Ideally, we'd like to have the ability to use different *draw strategies* for different situations. I've had a couple of attempts at implementing this but it's more complicated than I first realized. This discussion is about what we can do about it.

We started discussing this issue in #132 but that was off topic. Here's the relevant bits.

> It turns out that drawing to a render target every frame is *really* slow. With some minor tweaks I managed to get the frame rate up a *lot*.

> 1. We can't get rid of the render target completely. If we do that it'll re-introduce an old bug that causes gaps between the tiles.
> 2. The optimization I implemented tonight is a bit of a hack, I simply rendered the *entire* map to a render target *once* and drew that. It works, but means we have a *really* large texture loaded into the GPU. (Trading GPU memory for frame rate performance). This works on a PC but won't fly on a mobile device.
> 3. Only rendering the visible part of the map isn't trivial either. I tried doing this by re-rendering to the render target every time the visible rectangle changes (each time the camera moves one whole tile) and unfortunately the rendering becomes very *jumpy*.

> What we really need is a more balanced approach. Possibly rendering the map in several large *chunks* to keep the number of render target draws down and the total GPU memory at a reasonable level.

> The way the map is drawn is always going to be a trade off between memory and frame rate. I don't think there's ever going to be one solution that fits all situations.

> I propose that we provide a way to override the `DrawStrategy` so that people can render the map different ways in different situations. This way we can always provide a default strategy that works pretty well in most situations but can be overridden when people have different requirements.

> I suspect the way you draw small maps is going to be quite different for large maps. Also, the constraints on each platform is going to be different. On a PC with a decent graphics card it seems reasonable to simply render the entire map all at once. This gives a great frame rate at the cost of some GPU memory. On a phone, the map could be split up into smaller chunks, etc.

> There's also animated tiles to consider. We haven't implemented them yet, but drawing a large static map is going to be very different from drawing one with animated tiles.

> @LithiumToast said:
[...]

PR #190, opened 2016-05-01
Added support for Tiled animated tiles

CJRii:
This will allow the running of animations from Tile Layers. The SpriteBatch extension methods have been modified to now accept an optional parameter of a GameTime instance. If one is not provided, the animations will only show the first tile of the animation. The core of the change is that now all TiledTile objects will potentially have a reference to a TiledTileSetTile which contains all the information about the animation. This also has the side-effect of being able to access the Additional Properties of the tileset tiles.
Things that will need to be updated in the future:

- When the support to display Object Layers is implemented, animations will have to implemented there as well
- When layers are refactored to be more memory friendly as discussed in issue #136 the animation code will more than likely need updating as well.

The TiledMaps Demo has been updated to utilize the new functionality.


craftworkgames:
I did a quick code review and it looks good.

Did you update the demo to include animated tiles?

I'll take a closer look later.

CJRii:
I did update the TiledMaps demo application. I should also note, this is a breaking change in that the Tiled assets from older apps will have to run through the pipeline tool again to insert the new fields the reader is expecting.

LithiumToast:
@CJRii I think this line

```
_uniqueTileSetTiles = _animatedTiles.Select(t => t.TileSetTile).Distinct().ToList();
```

is going to cause extra garbage to be generated since it's invoked every frame.

CJRii:
@LithiumToast That line is inside the **if** block starting at line 67:

`if (_renderTarget == null)`

so it is going to only run once per layer because `_renderTarget` will be set in that same block.

LithiumToast:
@CJRii Ah you are right. Should looked at the whole file not just the diff.

craftworkgames:
Wonderful work mate. Merged :smile:

wmccullough:
Is there any way to get this change as part of a nuget package? Sorry for the necro but I'm in a position where I can't build from source atm.

Sorry for the necro.

PR #191, opened 2016-05-02
Particle rework

ExCellRaD:
Try 2 at updating the particle code.

Removed Range, RangeF and ColorRange in favor of using generic Range<T> in ParticleReleaseParameters.
Removed Axis, was just a normalized Vector2.
Changed how interpolation modifiers work, certain modifiers can have interpolators which control specific values of the particles. Those modifiers are AgeModifier and VelocityModifier, but more can be added. 
This still needs some more testing.

Stuff that got removed from previous PR:
No shape changes, no FastRand move, no profile changes.

Also these commits need to get squashed, but I failed horribly at trying that. this can probably get done when pulling right?


craftworkgames:
This works for me. Thanks.

There was one run-time error because one of the `Range` values was around the wrong way. 

Other than that it's pretty good. Your bracketing style still bothers me because it doesn't fit with the rest of the project and I've found no easy way of fixing it up.

ExCellRaD:
> Your bracketing style still bothers me because it doesn't fit with the rest of the project and I've found no easy way of fixing it up.

I'm gonna try to find a way to fix that, probably by changing my VS settings.

craftworkgames:
Maybe these settings?

image

What happens when you press Ctrl + K followed by Ctrl + D?

ExCellRaD:
The brackets seem to be under "new lines", but I think resharper is overwriting those values. I'll change them there.

It might be good to look into some solution like [Rebracer](https://github.com/SLaks/Rebracer) so that people who like different formatting styles have different settings for this project. I haven't used it yet, so it might need some testing before considering.

LithiumToast:
@ExCellRaD You can reformat a selection, file, or whole project by using [Code Cleanup](https://www.jetbrains.com/help/resharper/2016.1/Code_Cleanup__Index.html?origin=old_help).

PR #194, opened 2016-05-03
Angle struct

ExCellRaD:
This struct is based on the one from xenko engine, it allows you to use the unit type you want for angles. Supported unit types are: radians, degrees, gradians and revolutions.
It has methods to wrap (between -pi and pi) or to wrap positive (between 0 and 2*pi).
It also has methods to get the angle from a vector, and to create a vector (unit or specified length) from an angle.
Reasoning behind using a double for the value is because all the .net angle math functions use doubles, this will allow for less casts and higher precision.

### DIFF
--- a/Source/MonoGame.Extended.Tests/AngleTest.cs
+++ b/Source/MonoGame.Extended.Tests/AngleTest.cs
@@ -7,11 +7,11 @@ namespace MonoGame.Extended.Tests
     [TestFixture]
     public class AngleTest
     {
-        private const double DELTA = 0.000000000001;
+        private const float DELTA = 0.0000001f;
         [Test]
         public void ConstructorTest()
         {
-            var value = 0.5;
+            var value = 0.5f;
             var radians = new Angle(value);
             var degrees = new Angle(value, AngleType.Degree);
             var gradians = new Angle(value, AngleType.Gradian);
@@ -26,35 +26,35 @@ namespace MonoGame.Extended.Tests
         public void ConversionTest()
         {
             //from radians
-            var radians = new Angle(Math.PI);
+            var radians = new Angle(MathHelper.Pi);
             Assert.AreEqual(180d, radians.Degrees, DELTA);
             Assert.AreEqual(200d, radians.Gradians, DELTA);
             Assert.AreEqual(0.5, radians.Revolutions, DELTA);
             //to radians
-            var degrees = new Angle(180d, AngleType.Degree);
-            var gradians = new Angle(200d, AngleType.Gradian);
-            var revolutions = new Angle(0.5, AngleType.Revolution);
-            Assert.AreEqual(Math.PI, degrees.Radians, DELTA);
-            Assert.AreEqual(Math.PI, gradians.Radians, DELTA);
-            Assert.AreEqual(Math.PI, revolutions.Radians, DELTA);
+            var degrees = new Angle(180f, AngleType.Degree);
+            var gradians = new Angle(200f, AngleType.Gradian);
+            var revolutions = new Angle(0.5f, AngleType.Revolution);
+            Assert.AreEqual(MathHelper.Pi, degrees.Radians, DELTA);
+            Assert.AreEqual(MathHelper.Pi, gradians.Radians, DELTA);
+            Assert.AreEqual(MathHelper.Pi, revolutions.Radians, DELTA);
         }
 
         [Test]
         public void WrapTest()
         {
-            for (var d = -10d; d < 10d; d += 0.1d)
+            for (var f = -10f; f < 10f; f += 0.1f)
             {
-                var wrappositive = new Angle(d);
+                var wrappositive = new Angle(f);
                 wrappositive.WrapPositive();
 
-                var wrap = new Angle(d);
+                var wrap = new Angle(f);
                 wrap.Wrap();
 
                 Assert.GreaterOrEqual(wrappositive.Radians, 0);
-                Assert.Less(wrappositive.Radians, 2d * Math.PI);
+                Assert.Less(wrappositive.Radians, 2d * MathHelper.Pi);
 
-                Assert.GreaterOrEqual(wrap.Radians, -Math.PI);
-                Assert.Less(wrap.Radians, Math.PI);
+                Assert.GreaterOrEqual(wrap.Radians, -MathHelper.Pi);
+                Assert.Less(wrap.Radians, MathHelper.Pi);
             }
         }
 
@@ -62,7 +62,7 @@ namespace MonoGame.Extended.Tests
         public void VectorTest()
         {
             var angle = Angle.FromVector(Vector2.One);
-            Assert.AreEqual(-Math.PI / 4d, angle.Radians, DELTA);
+            Assert.AreEqual(-MathHelper.Pi / 4d, angle.Radians, DELTA);
             Assert.AreEqual(10f, angle.ToVector(10f).Length());
 
             angle = Angle.FromVector(Vector2.UnitX);
@@ -70,7 +70,7 @@ namespace MonoGame.Extended.Tests
             Assert.IsTrue(Vector2.UnitX.EqualsWithTolerance(angle.ToUnitVector()));
 
             angle = Angle.FromVector(-Vector2.UnitY);
-            Assert.AreEqual(Math.PI / 2d, angle.Radians, DELTA);
+            Assert.AreEqual(MathHelper.Pi / 2d, angle.Radians, DELTA);
             Assert.IsTrue((-Vector2.UnitY).EqualsWithTolerance(angle.ToUnitVector()));
         }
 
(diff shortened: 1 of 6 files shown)


LithiumToast:
>Reasoning behind using a double for the value is because all the .net angle math functions use doubles, this will allow for less casts and higher precision.

When it comes to games, floats are most often good enough. Often times floats are preferred for algorithms in games since the extra 4 bytes involved with `double` is unnecessary for most values. Saving on 4 bytes here and there might not sound like a big deal, bit with real time applications like games everything helps.  Also if people use `SpriteBatch` with this struct, it will casted to `float` regardless. In fact, I think some profiling would need to be done to make any claim that casting values degrades performance...

ExCellRaD:
> I think some profiling would need to be done to make any claim that casting values degrades performance...
Didn't claim it would increase performance, only that it would require less casts, but I also think this should be done. Could be important information when designing structs and classes.

LithiumToast:
> Didn't claim it would increase performance

Ok I'll say it more simply

>higher precision

Games often don't need it.

Perhaps a `AngleF` and `AngleD` or something is appropriate?

craftworkgames:
I don't think we can take code from the xenko engine under their license terms.

ExCellRaD:
[this](https://slimdx.googlecode.com/svn-history/r1657/branches/v2/SlimDX2/Sources/SlimDX2.Math/Angle.cs) is the actual source. It's also linked in xenko's file.

ExCellRaD:
> Perhaps a AngleF and AngleD or, Angle32 and Angle64 something is appropriate? 

I think we should only use float values when the angle functions use float values, I don't know whether or not something like that is feasible though (implementing them in this library). The main functions being Sin, Cos, Atan2 and similar. The .net functions all have doubles as parameters and return values so I just don't see the point of casting them to float even though the added precision is not needed.

LithiumToast:
>The .net functions all have doubles as parameters and return values so I just don't see the point of casting them to float even though the added precision is not needed.

@ExCellRaD Why does it matter what .NET framework uses for their math functions? Games use `float` when dealing with angles for the majority of their API. `SpriteBatch` is one example. The `AngleSingle` you just linked is another.

ExCellRaD:
> Games use float when dealing with angles

Thats because most engines (read the math functions in those engines) including the one I linked actually use floats, this library - using the .net functions - doesn't right now. 
SpriteBatch does this:
`(float)Math.Sin(rotation)`
`(float)Math.Cos(rotation)`
Cast to a double, call function, cast back to float. 2 times because of two functions.
Using a double it would only be: Call function. This while even having a little bit more precision.

Now your second point still stands, `SpriteBatch` does still take a float as parameter which is something we can't change. That shouldn't be the deciding factor though, we could implement `Angle` as a double into `PrimitiveBatch` which could (and probably should) be used as spritebatch when using this library.

LithiumToast:
@ExCellRaD I think the solution here is to provide `float` and `double` math functions and leave angles as `float`.

ExCellRaD:
I also think that is a prefectly good solution. Someone needs to make those tough.

LithiumToast:
@ExCellRaD Check out https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Math.cs

Cos and Sin are internal calls to some optimized code most likely written in C or C++. So that's a no, on `float` math functions.

I think casting between `float` and `double` is fine. Why does it matter so much anyways? Performance shouldn't be a problem and all XNA / MonoGame that have been built use those functions already. Why break the established way to do it now?

ExCellRaD:
> Why change the established way to do it now?

I think you're missing the point here, the change is supposed to be from `float` to `Angle`, not from `float` to `double`. You can still use floats, but you automaticaly cast them to a double when using it as an `Angle` which is what happens now anyway because of the functions having doubles as parameters. Using the `double` is not the change, using `Angle` is.

LithiumToast:
Well, there should be a `float` and `double` version of `Angle` then. Not everyone is going to need or want to use `double` as the backing value for their angles; especially in scenarios where it's stored in a struct or class.

ExCellRaD:
Could you give an example?

LithiumToast:
@ExCellRaD 

```
public struct MyWorldObj
{
    public Vector2 Position;
    public Angle Rotation;
}
```

If `Angle` is 4 bytes, the struct takes 12 bytes in memory. If `Angle` is 8 bytes, the struct takes 16 bytes in memory. Things like that matter for algorithms when looping over an array of structs like this; more memory can fit in one cache line for the CPU, thus the 12 byte version will run faster on average.

craftworkgames:
> this is the actual source. It's also linked in xenko's file.

If we're complying with the license terms that's cool. It looks like we need to include the license terms at the top of the file just like xenko does. I also usually like to ask the original author permission as a courtesy if possible but it's probably not a big deal here.

And about the `float` vs `double` issue I'm going to make a call on this.

Let's stick with `float`.

Sorry to shoot you down on this @ExCellRaD but I think @LithiumToast has some very good points. 

More importantly, using `float` is a widely accepted practice in lots of game frameworks / engines for many reasons. The actual reasons don't really matter, we are an extension of MonoGame and MonoGame uses `float`s so that's what we are going to do.

Even the original SlimDX and xenko implementations just used `float`s so I don't see any reason to try and reinvent the wheel here. We don't need any `AngleF`, `AngleD`, `Angle32` or `Angle64` variations either. Just keep it simple.

ExCellRaD:
> The actual reasons don't really matter, we are an extension of MonoGame and MonoGame uses floats so that's what we are going to do.

I agree with that, but not with the other points. It is reason enough though to use float though, so I'll get to changing that.

>It looks like we need to include the license terms at the top of the file just like xenko does.

That's a good idea, but I'm not sure it's necessary. I'll do it just to make sure though.

ExCellRaD:
It seems that a commit from #191 didn't go through, I added that here as it was only a small variable change on 2 spots.

Source/MonoGame.Extended/Particles/Modifiers/Interpolators/ColorInterpolator.cs
```diff
@@ -18,7 +18,7 @@ public sealed class ColorInterpolator : IInterpolator
         public HslColor FinalColor { get; set; }
         
         public unsafe void Update(float amount, Particle* particle) {
-            particle->Color = HslColor.Lerp(InitialColor, FinalColor, particle->Age);
+            particle->Color = HslColor.Lerp(InitialColor, FinalColor, amount);
```
craftworkgames: Why is this changing from `Age` to `amount`?

ExCellRaD: Because the age should only be used in the `AgeModifier`, using amount makes it so that `VelocityModifier` (and maybe future classes like that) also work.

craftworkgames: Okay, I'll take your word for it. I hope you've tested it.

Source/MonoGame.Extended.Tests/AngleTest.cs
```diff
@@ -0,0 +1,88 @@
+﻿using System;
+using Microsoft.Xna.Framework;
+using NUnit.Framework;
+
+namespace MonoGame.Extended.Tests
+{
+    [TestFixture]
+    public class AngleTest
+    {
+        private const float DELTA = 0.0000001f;
```
craftworkgames: Nowhere in the [C# naming conventions](https://msdn.microsoft.com/en-us/library/x2dbyw72(v=vs.71).aspx) does it say to use ALL_CAPS. The [internet agrees](http://stackoverflow.com/questions/242534/c-sharp-naming-convention-for-constants) with this. So these constants should use standard field or property naming `Delta` or `_delta`.

craftworkgames:
I'll fix up these few minor issues.

Thanks.

Source/MonoGame.Extended.Tests/AngleTest.cs
ExCellRaD: That was a naming convention we used at school, so it should be `_delta` for private constants and `Delta` for public ones from now on?

craftworkgames: Correct.

You could read through the Microsoft guidelines if you like, but as a rule of thumb the easiest thing to do is to look at examples in the .NET framework. For example, one of the constants in the .NET framework is [`int.MinValue`](https://msdn.microsoft.com/en-us/library/system.int32.minvalue(v=vs.110).aspx).

`private` variable naming conventions are not actually defined in the guidelines but as you've already noticed we're using the underscore prefix. The reason for this is to differentiate private member variables from properties when you're reading the code. It's also helpful because you don't have to qualify everything with `this`.

I use a tool called [Reshaper](https://www.jetbrains.com/resharper/) that really helps with this kind of stuff. It does cost a bit of money but even using the 30 day trial is useful to learn loads of stuff about how to write really clean code. It kind of teaches you as you go.

Btw.. I've noticed a huge improvement in your PR's lately. I appreciate the effort.

guFalcon: @craftworkgames Maybe we could push the .DotSettings files of the various projects as well. I think (but alas, I only THINK that) it's where Resharper keeps the project-specific settings. Would be awesome if everyone owning Resharper would have the same settings automatically...

ExCellRaD: > Maybe we could push the .DotSettings files of the various projects as well.
Very good idea, it would help a lot for people like me who are used to different conventions and schemes. I don't know how that works exactly but it's definitely worth looking into

PR #196, opened 2016-05-04
Tiled map: Fix for Tile Id and Tile animation.

vadimline:
- Calculate correct TiledTileSetTile id inside TileSet. (Due to documentation Tile id is local id inside TileSet http://doc.mapeditor.org/reference/tmx-map-format/), this will correctly render animation if several tilesets used in map.

- Fix tile animation rendering. Prevent first frame of the animation tile to be always rendered.


craftworkgames:
Thanks for the fixes.

vadimline:
You are welcome :)

PR #199, opened 2016-05-05
Added MouseIsDown listener

Scellow:
Added a missing mouse listener that is useful when we need to do actions while mouse button is down

### DIFF
--- a/Source/MonoGame.Extended/InputListeners/MouseListener.cs
+++ b/Source/MonoGame.Extended/InputListeners/MouseListener.cs
@@ -37,6 +37,7 @@ namespace MonoGame.Extended.InputListeners
         public event EventHandler<MouseEventArgs> MouseDown;
         public event EventHandler<MouseEventArgs> MouseUp;
         public event EventHandler<MouseEventArgs> MouseClicked;
+        public event EventHandler<MouseEventArgs> MouseIsDown;
         public event EventHandler<MouseEventArgs> MouseDoubleClicked;
         public event EventHandler<MouseEventArgs> MouseMoved;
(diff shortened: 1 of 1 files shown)


craftworkgames:
Why? Doesn't this defeat the purpose of having event based inputs?

Scellow:
I don't think so, how would you do for a FPS game where you need to hold the mouse down to shoot?

ExCellRaD:
> I don't think so, how would you do for a FPS game where you need to hold the mouse down to shoot?

For that just have a `bool` called "shooting". Set it to true with `MouseDown` and set it to fase with `MouseUp`. Having the `MouseDown` event makes it so that it gets invoked every frame (you probably don't want to shoot every frame).

craftworkgames:
> I don't think so, how would you do for a FPS game where you need to hold the mouse down to shoot?

I guess what I'm saying is, firing an event every frame is overkill when you can simply put it directly in your `Update` method.

            var mouseState = Mouse.GetState();

            if (mouseState.LeftButton == ButtonState.Pressed)
                Shoot();

> Just have a `bool` called "shooting". Set it to true with `MouseDown` and set it to false with `MouseUp`. 

This is of course another way to do it using events. To me, an event is something that *happened*, not something that is *happening*.

> you probably don't want to shoot every frame.

Yes, this is true. In my experience it's never as simple as calling `Shoot` every frame. Somewhere you'd still need to implement some sort of delay between bullets.

Input listeners provide event based input and are intended to provide an alternative way to handle input when you need it. For example, when you're implementing buttons on a user interface. 

They are not intended to be a replacement for state based input which is typically used for gameplay.

PR #205, opened 2016-05-30
Collision Detection: Brute-force Broadphase & SAT Narrowphase

LithiumToast:
Do not merge until ready. As of this comment, this is a preliminary pull request; it's to allow for feedback while I develop collection detection.

Collision detection design is based off of Box2D in C++ and Farseer Physics in C# (which is based on Box2D).

Goal for this pull request: Brute-force Broadphase with Separating-Axis-Thorem Narrowphase.


craftworkgames:
> Collision detection design is based off of Box2D in C++ and Farseer Physics in C#

When you say based off these 2 libraries what exactly are you intending to create? If we are going down the path of creating a full blown physics engine I think our time would be better spent writing a thin wrapper around Farseer.

What I was hoping to provide in MonoGame.Extended from the start is some basic helper classes that let you write collision detection code in games that *don't* require full blown physics. This is a [very different set of requirements](http://gamedev.stackexchange.com/questions/44020/xna-c-platformer-physics-engine-or-tile-based/44308#44308) and I think it would be a huge waste of our time if we take the wrong path here.

LithiumToast:
@craftworkgames Collision *detection* is based off Box2D; they use the same principle of a broad phase (see [IBroadPhase](https://farseerphysics.codeplex.com/SourceControl/latest#SourceFiles/Collision/IBroadPhase.cs) and [ContactManager](https://farseerphysics.codeplex.com/SourceControl/latest#SourceFiles/Dynamics/ContactManager.cs)) and narrow phase. Physics is collision *response*, which here is left as a pluggable interface to implement. 

The flow of the program is like this:
- Broad phase finds **possible** colliding bodies by quickly checking the bodies respective bounding volumes for intersection.
- Narrow phase takes the output of the broad phase and checks if the colliding bodies truly intersect by using more computationally expensive techniques such as convex polygon to convex polygon intersection or separating axis theorem.
- Collision response "phase" takes the output of the narrow phase and applies a "response" such as correcting the position of two bodies so they don't intersect.

craftworkgames:
@LithiumToast Awesome. I just wanted to make sure we are on the same page. What you are doing sounds great :+1:

craftworkgames:
Looks like you missed a file or something and broke the build.

http://build.craftworkgames.com/viewLog.html?buildId=1105&tab=buildResultsDiv&buildTypeId=MonoGameExtended_CI

LithiumToast:
Interestingly I named a folder `Debug` and I think it got ignored for staging.

LithiumToast:
This wasn't ready for merging as it's still not ready to be used, but okay.

craftworkgames:
Yeah, you're right. My bad. I got a little excited ;)

I'm reverting this until you're ready.

I think it might be better to submit these big PR's to a different branch instead to make sure we keep `develop` reasonably clean.

I've created a `collision-detection` branch. Please use that next time.

PR #209, opened 2016-06-11
Bug: Adjust tiled tiles to effectively draw from lower left corner

kevhiggins:
I've noticed what appears to be a bug in the MonoGame.Extended Tiled layer rendering.

Here is what my map looks like in Tiled
tiled-view

Here is what my map looks like in the game:
game-view

The difference appears to be that MonoGame.Extended draws the tile sprite from the top left corner of the tile, while Tiled itself draws sprites from the bottom left corner (going up) of its tiles.


craftworkgames:
Look reasonable. Thanks.

Did you happen to test this with the Tiled demo?

kevhiggins:
Good idea, I did not.

I just checked it out, and the demo still properly reflects the tiled map.

Issue #211, opened 2016-06-13
TiledLayer draws only as high as it is wide


geselle-jan:
When I use `MonoGame.Extended.Maps.Tiled.Draw(this SpriteBatch spriteBatch, TiledLayer layer, Camera2D camera);` with a map that is higher that it is wide, it draws the layer from top to bottom only as far as the maps width, after that I get only transparency.
When I increase the maps width in tiled I can confirm that the height of the part of the map that gets drawn is directly correlated to the width. As soon as the map gets wider than high, the problem disappears.

geselle-jan:
I just found the issue:

In `TiledTileLayer.cs` in Line 72 there was:

`_renderTarget = new RenderTarget2D(_renderTargetSpriteBatch.GraphicsDevice, _map.WidthInPixels, _map.WidthInPixels);`

But it should be:

`_renderTarget = new RenderTarget2D(_renderTargetSpriteBatch.GraphicsDevice, _map.WidthInPixels, _map.HeightInPixels);`

craftworkgames:
Thanks for the contribution :)

PR #212, opened 2016-06-13
fixed issue #211 - tilemap height can now be higher than wide

geselle-jan:
see #211 for further info


craftworkgames:
Wonderful. A small typo, but so very wrong.

Thanks for fixing it :)

PR #217, opened 2016-07-01
ShapeBatch

LithiumToast:
- New `ShapeBatch` class. Currently supports drawing lines, triangles, quads, circles, arcs, rectangles and sprites out of the box with minimal setup.
- New Shapes demo using `ShapeBatch` class.
screen shot 2016-06-30 at 5 19 17 pm
- New `ShapeBuilder` class to generate a list of 3D points for circles, arcs and rectangles with position, size and rotation if applicable. Used by `ShapeBatch` to get the points of a circle, arc and rectangle on the fly.
- PrimitiveBatch's design has been simplified and improved, but still needs some work with fragmenting large batches into multiple smaller batches. Deferred batching also needs to be fixed.
- Size and SizeF extra functionality (used for sprites).


charles-esterbrook:
I see that inside struct Size, this:
`public static Size Empty => new Size(0, 0);`
Was changed to:
`public static Size Empty => new SizeF(0, 0);`
I'm guessing that was unintentional.

LithiumToast:
@charles-esterbrook Good catch. I fixed it.

LithiumToast:
- Fixed deferred rendering. I have yet to fix the `DrawCircle` method, but the `DrawArc` method works fine (both are very similar). I'll get to it soon.
- `PrimitiveBatch` has to be inherited to be used (`ShapeBatch` inherits from `PrimitiveBatch`).
- `ShapeBatch` now uses only `PrimitiveType.TriangleList` instead of a combination of `PrimitiveType.LineList` and `PrimitiveType.TriangleList` for lines and triangles respectively. Lines are rendered as quadrilaterals with `PrimitiveType.TriangleList`. This allows for lines to have a width, which was not supported with line list primitives in MonoGame (DirectX). The added bonus is merging draw calls for deferred rendering is more likely to happen with only one type of primitive.
- `ShapeBuilder` is now a static class. It takes a delegate which is used to output the points of a shape.
- Added `RenderGeometryBuilder` static class. It's primary purpose is to translate points created using `ShapeBuilder` into vertices for rendering with `PrimitiveType.TriangleList`.
- Added `RenderGeometryBuffer` which just encapsulates an array of vertices and an array of vertex indices. This allows for `RenderGeometryBuilder` to be used outside the use of `PrimitiveBatch`.

craftworkgames:
Is this ready to merge? It's quite difficult to tell with these massive PR's. It makes me nervous pulling them into the `develop` branch without really knowing what could break or whatever.

LithiumToast:
@craftworkgames It's almost ready. I have to write the code to tessellate shapes with both "fill" and "stroke" modes. I could close this pull request and point it at `collisions` branch, or you could open a new branch for this. It definitely shouldn't be going into develop.

craftworkgames:
@LithiumToast It's proving quite difficult to deal with these big pull requests. 

 - they hang around for ages
 - the changes get out of date and create merge conflicts
 - there's a *lot* of code to go over

It might be better to break them up into a bunch of smaller PR's. 

I know we said that it shouldn't go into the `develop` branch, but I'm actually coming around to the idea that we should just do it. If it's ready to go we might as well just get it in, even if it's not perfect.

LithiumToast:
Yeah I agree. I'm going to close this and release an do a pull request for `PrimitiveBatch` soon.

craftworkgames:
I've created a `graphics` branch for you.

lofcz:
What happened to this?

craftworkgames:
> What happened to this?

As you can probably imagine, open source projects such as this one tend to be driven by a small number of primary contributors (myself) and occasional pull requests from others. Most of these contributors are doing it for their own interests and come and go as they get motivated or as time permits. This often leads to the scenario that things are started and not finished.

I would love to get everything done myself but of course as the project grows bigger that's simply not possible because my own spare time is limited.

Of course, if you're interested and have the time you are quite welcome to take on this work in submit a pull request. Just remember to please include a demo to show how it works.

LithiumToast:
@lofcz Unfortunately, I don't have the time to work on this any further. The code is mostly there. It is just too much to think about a generic renderer for every scenario of a game's needs. Especially when `SpriteBatch` is so well optimized. If you want to render shapes for debugging purposes I suggest you look around the web for `PrimitiveBatch`. If you want to render shapes for gameplay I suggest you look at the code and roll your own specific to your game's needs.

craftworkgames:
Good to hear from you @LithiumToast. What are you doing with yourself these days?

> It is just too much to think about a generic renderer for every scenario of a game's needs.

Just to be clear. We never really intended to create a generic renderer that suits the needs of every game. The original idea was simply to create a renderer that can do shapes and that's it. 

Of course, sometimes when you set out with certain intentions, things can get hairy in the details. I don't really know how or why we ended up where we did but I understand this happens sometimes. I've been coding for long enough to know things are never as simple as they seem.

That said, if there's interest in resurrecting this idea and finding a way to keep the scope small and simple I'm all for it.

LithiumToast:
> What are you doing with yourself these days?

Finishing up at the university.

> Just to be clear. We never really intended to create a generic renderer that suits the needs of every game. The original idea was simply to create a renderer that can do shapes and that's it.

Usually, 2D shapes have to be [polygon triangulated](https://en.wikipedia.org/wiki/Polygon_triangulation) either for triangle strip or triangle list GPU primitives. There can be a lot of assumptions going on here for a game's needs. One of the things I learned when looking at how other projects did this was to use a compute or geometry shader. Since MonoGame doesn't have an API for either type of shaders that means all the calculations have to be done on the CPU. That really sucks. Furthermore, shaders are usually used for certain features when drawing lines such as end caps (butt, round, square), line joins (miter, round, bevel), and dotted/solid anti-aliased outlines.

craftworkgames:
For reference here's the open MonoGame issue about geometry shaders: https://github.com/MonoGame/MonoGame/issues/4567

Of course, doing this stuff on the CPU is going to have some limits. That doesn't mean it's not useful in certain situations. Quite often I find myself needing to draw a shape or two and performance is not a huge concern.

PR #218, opened 2016-07-01
RectangleF tests added.

lozzajp:
Added some basic tests, @craftworkgames if this is the style you like I will add some more and look at other areas you might want added to.

### DIFF
--- a/Source/MonoGame.Extended.Tests/Shapes/RectangleFTests.cs
+++ b/Source/MonoGame.Extended.Tests/Shapes/RectangleFTests.cs
@@ -8,7 +8,44 @@ namespace MonoGame.Extended.Tests.Shapes
     public class RectangleFTests
     {
         [Test]
-        public void RectangleF_Intersects_Test()
+        public void ConstructorsAndProperties()
+        {
+            // Create a RectangleF from ints and from floats.
+            var intRectangleF = new RectangleF(10, 20, 64, 64);
+            var floatRectangleF = new RectangleF(7.5f, 17.5f, 62.5f, 62.5f);
+
+            // Test basic constructing
+            Assert.AreEqual(new RectangleF(10, 20, 64, 64), intRectangleF);
+            Assert.AreEqual(new RectangleF(10f, 20f, 64f, 64f), intRectangleF);
+
+            Assert.AreNotEqual(new RectangleF(10f, 20f, 64f, 64f), floatRectangleF);
+            Assert.AreNotEqual(new RectangleF(7.5f, 17.5f, 62.5f, 62.5f), intRectangleF);
+
+            Assert.AreEqual(new RectangleF(7.5f, 17.5f, 62.5f, 62.5f), floatRectangleF);
+
+            // Test int + float during construction
+            Assert.AreEqual(new RectangleF(5 + 2.5f, 10 + 7.5f, 60 + 2.5f, 60 + 2.5f), floatRectangleF);
+
+            // Test float + float during construction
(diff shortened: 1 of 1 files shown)


Source/MonoGame.Extended.Tests/Shapes/RectangleFTests.cs
```diff
@@ -8,7 +8,44 @@ namespace MonoGame.Extended.Tests.Shapes
...
+            Assert.AreEqual(17.5f + 62.5f, floatRectangleF.Bottom);
+
+            // IsEmpty and static Empty
+            Assert.AreEqual(false, floatRectangleF.IsEmpty);
+            Assert.AreEqual(false, intRectangleF.IsEmpty);
+            Assert.AreEqual(true, new RectangleF().IsEmpty);
+            Assert.AreEqual(new RectangleF(), RectangleF.Empty);
+        }
+
+        #region Deprecated Tests
```
craftworkgames: I'd rather not "deprecate" tests. We either have them or we don't. In other words, if you have two tests that cover the same thing delete one. Otherwise, keep them.

Source/MonoGame.Extended.Tests/Shapes/RectangleFTests.cs
```diff
@@ -8,7 +8,44 @@ namespace MonoGame.Extended.Tests.Shapes
...
+
+            // Test float + float during construction
+            Assert.AreEqual(new RectangleF(3.75f + 3.75f, 17.5f, 62.5f, 62.5f), floatRectangleF);
+
+            // Test sides
+            Assert.AreEqual(7.5f, floatRectangleF.Left);
+            Assert.AreEqual(7.5f + 62.5f, floatRectangleF.Right);
+            Assert.AreEqual(17.5f, floatRectangleF.Top);
+            Assert.AreEqual(17.5f + 62.5f, floatRectangleF.Bottom);
+
```
craftworkgames: It's generally better to have a lot of small tests vs large ones that test multiple things. If you find yourself writing a comment to describe what you're testing, that's a sign you should extract that code into a new test.

craftworkgames:
You've made a good start. Writing good unit tests is actually pretty difficult. I've been writing unit tests at my day job for about 5 years now and I still consider myself an amateur.

There's a lot of misinformation about how to write good unit tests on the internet. However, it's worth doing a little reading up to get a sense of what makes a good test.

The highest voted couple of answers on this [stack overflow question](http://stackoverflow.com/questions/62625/how-do-you-know-what-to-test-when-writing-unit-tests) and [this one](http://programmers.stackexchange.com/questions/21133/how-to-write-good-unit-tests) should give you an idea of how to improve your tests.

Also, be careful of writing tests that have zero value. For example, what is this test actually trying to prove?

    var floatRectangleF = new RectangleF(7.5f, 17.5f, 62.5f, 62.5f);
    Assert.AreEqual(new RectangleF(5 + 2.5f, 10 + 7.5f, 60 + 2.5f, 60 + 2.5f), floatRectangleF);

Would it be possible to make that test **fail** by *changing the implementation* of RectangleF? Probably not.

Remember, every unit test comes at a small cost. Testing a property like `Right` not really going to bring a whole lot of value. We are better off focusing our energy into testing methods like `Union`, etc.

lozzajp:
Alright then, at least I have learned one thing so far, I am very inexperienced with testing.

I will have a look around, read those posts you linked so far. See if I can rewrite something worth contributing.

If you happen to get around to writing up tests for this one, it may give me more of an idea and I could try help out testing other F shapes etc.

craftworkgames:
Closing this issue since it's been hanging around for a while. We can always re-open it.

PR #219, opened 2016-07-02
Modify TiledMap to expose the list of its tilesets publicly.

weaversam8:
I wrote this because if a tilesheet is stored inside a Tiled Map file, there is currently no publicly acessible way to access the TiledTileSheet object for that tilesheet with only a reference to a TiledMap. This means that if there are properties stored inside a tilesheet that is inside a map, there is NO WAY to access them.

I've never done a PR before, so forgive me if I've done something wrong, but as far as I've seen, this shouldn't affect anything else...

### DIFF
--- a/Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
+++ b/Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
@@ -16,13 +16,13 @@ namespace MonoGame.Extended.Maps.Tiled
             _graphicsDevice = graphicsDevice;
             _layers = new List<TiledLayer>();
             _objectGroups = new List<TiledObjectGroup>();
-            _tilesets = new List<TiledTileset>();
 
             Width = width;
             Height = height;
             TileWidth = tileWidth;
             TileHeight = tileHeight;
             Properties = new TiledProperties();
+            Tilesets = new List<TiledTileset>();
             Orientation = orientation;
         }
 
@@ -32,7 +32,6 @@ namespace MonoGame.Extended.Maps.Tiled
                 tiledLayer.Dispose();
         }
 
-        private readonly List<TiledTileset> _tilesets;
         private readonly GraphicsDevice _graphicsDevice;
         private readonly List<TiledLayer> _layers;
         private readonly List<TiledObjectGroup> _objectGroups;
@@ -41,6 +40,7 @@ namespace MonoGame.Extended.Maps.Tiled
         public int Height { get; }
         public int TileWidth { get; }
         public int TileHeight { get; }
+        public List<TiledTileset> Tilesets { get; }
         public Color? BackgroundColor { get; set; }
         public TiledRenderOrder RenderOrder { get; set; }
         public TiledProperties Properties { get; private set; }
@@ -56,7 +56,7 @@ namespace MonoGame.Extended.Maps.Tiled
         public TiledTileset CreateTileset(Texture2D texture, int firstId, int tileWidth, int tileHeight, int spacing = 2, int margin = 2)
         {
             var tileset = new TiledTileset(texture, firstId, tileWidth, tileHeight, spacing, margin);
-            _tilesets.Add(tileset);
+            Tilesets.Add(tileset);
             return tileset;
         }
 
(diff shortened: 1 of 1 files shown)


Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
```diff
@@ -41,6 +40,7 @@ public void Dispose()
         public int Height { get; }
         public int TileWidth { get; }
         public int TileHeight { get; }
+        public List<TiledTileset> Tilesets { get; }
```
craftworkgames: It's not ideal that we are exposing mutable `List` properties. In this case it's probably okay but most of the other collection properties are exposed as `IEnumerable`'s to prevent adding and removing elements after loading. 

The `ObjectGroups` property should have never been exposed as a `List`. Now it's a breaking change to go back the other way unfortunately.

Now that we're on .NET 4.5 it might be better to change them to `IReadOnlyCollection` instead. Although, with a little redesign we could get rid of the `Create` method pattern entirely and expose our own custom collections that allow *normal* `Add` and `Remove` operations.

Anyway, starting to rant.

craftworkgames:
I'm happy to accept this change but I think there's a few other improvements we should make to the API besides this. They require a bit more thought though.

See my other comment and let me know if you want to make any other changes before I pull.

McSora:
Hey Guys,
I just saw this pull request and got interested for I did the exact same thing a few days ago, exposed the tilesets of the tiled map to get the list...
But then i wondered if this is what the TiledMap API should really provide.
If you look at the TiledTileLayer, it exposes the method:
public TiledTile GetTile(int x, int y)

Please correct me if i am wrong but i believe when working with a tiledMap we could access the tilesetTiles the same way... like:
public TiledTileSetTile GetTilesetTile(int x, int y)
That would be possible, for the TiledTileLayer already knows the Map it belongs to.

In another situation i would need to access a specific TilesetTile i would have the exact TileID in the tileset collection.
This specific ID is held by the TiledTile you access via TiledLayer.GetTile(x,y).
So we could extend the Map with the method GetTile(int ID) which could call a method of the Tileset-Collection .GetTile(int ID) 
(Sure, I know that until now the Tileset collection is just a generic list)

To be honest I cannot think of a situation in which i would need to access the whole exposed Tileset-Collection of the map. Everytime I need to access a TilesetTile ... maybe to access the TileProperties i would know where to look exactly by passing either X and Y or the TileID.
Correct me if i am missing something :)

Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
LithiumToast: @craftworkgames Have you considered [`IReadOnlyList<T>`](https://msdn.microsoft.com/en-us/library/hh192385)?. `List<T>` implements the interface out of the box so could just cast `List<T>` to `IReadOnlyList<T>` and expose the latter in the API.

weaversam8: Sounds like it'd be a good change... I'll add a commit with that cast...

LithiumToast: @weaversam8 Cool. You could even have a local variable for the tile sets and then just have an explicit property getter that returns an `IReadOnlyList<TiledTileset>`. Here is a snippet:

```
private List<TiledTileset> _tilesets;

public IReadOnlyList<TiledTileset> Tilesets
{
    get { return _tilesets; }
}
```

Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
```diff
@@ -34,13 +34,14 @@ public void Dispose()
...
         private readonly List<TiledLayer> _layers;
+        private readonly List<TiledTileset> _tilesets;
         private readonly List<TiledObjectGroup> _objectGroups;
 
         public int Width { get; }
         public int Height { get; }
         public int TileWidth { get; }
         public int TileHeight { get; }
-        public List<TiledTileset> Tilesets { get; }
+        public IReadOnlyList<TiledTileset> Tilesets { get { return  _tilesets; } }
```
weaversam8: How's this look?

LithiumToast: That should work. 😃

craftworkgames:
Coolies.

PR #223, opened 2016-07-08
New Tiled renderer

rafaelalmeidatk:
I successfully rendered the tiles in the screen, but there are gaps between the tiles. Maybe it's due to the fact that I'm using Triangle Lists and not Triangle Strips, so the vertices aren't shared. Now the layers have a Z property, it's not being used by now, but if it's necessary, it's there.

The PR isn't ready to merge, I have to apply the viewport and the camera (and fix the gaps). Only tested with orthographic maps. I also don't draw the background color, I will implement it later when the main point is done.


rafaelalmeidatk:
> I suggest you clone a fresh copy of the develop branch and apply your changes from there. It seems you are working off a really old branch.

Thanks, I didn't notice.

> Also you can fix the gaps between tiles issue by having TiledMap.Draw() like this:

The first time I sent the code I was already drawing the tiles with point clamp, I also thought it would fix, but only in some cases it does. See this images, the window size is different, so sometimes the gaps appears, sometimes not.

screenshot_38
screenshot_37

LithiumToast:
I added these lines to before applying render passes and `DrawIndexedPrimitives` and the gaps are gone:

```
_graphicsDevice.BlendState = BlendState.AlphaBlend;
_graphicsDevice.SamplerStates[0] = SamplerState.PointClamp;
```

rafaelalmeidatk:
The gaps still are there. I wrote it:

```
_graphicsDevice.BlendState = BlendState.AlphaBlend;
            _graphicsDevice.SamplerStates[0] = SamplerState.PointClamp;
            _graphicsDevice.SetVertexBuffer(_vertexBuffer);
            _graphicsDevice.Indices = _indexBuffer;

            foreach (var pass in _basicEffect.CurrentTechnique.Passes)
            {
                pass.Apply();
                _graphicsDevice.DrawIndexedPrimitives(
                    PrimitiveType.TriangleList,
                    0,
                    0,
                    _primitivesCount
               );
            }
```

The gaps don't appears to you even if you reside the window?

rafaelalmeidatk:
I implemented the camera movement and noticed that the gaps don't appear everywhere, see:

screenshot_39

Is there a way to join this vertices in order to they don't "split"?

LithiumToast:
@rafaelalmeidatk TriangleList is the way to go here because TriangleStrip would require two tiles to share vertex information which is not always possible since two tiles could be using different texture coordinates. The gaps problem is with how sampling is done for texture across the triangles for the tile. I tested zooming the camera in/out using `PointClamp` and it seems to work for the most part, but there seems to be some subtle tearing sometimes.

rafaelalmeidatk:
@LithiumToast I pushed the vertices 0.5f to the left, right and bottom. This removed all the gaps in every zoom I tested. Would this be ok?

LithiumToast:
Oh, yes, you need to apply a translate to the projection matrix if you have not already.

```
var projection = Matrix.CreateTranslation(-0.5f, -0.5f, 0) * Matrix.CreateOrthographicOffCenter(...);
```

There is a subtle difference between the position of vertices and the position of texels from the texture. Short story is texels are sampled from the centre of a pixel and thus the projection matrix needs to a translation of half a pixel.

LithiumToast:
Just tested this half pixel adjustment. No tearing, no gaps.

craftworkgames:
Great work guys. It's looking good.

I'm not in front of a computer at the moment but I'll take a closer look later.

rafaelalmeidatk:
Well, I think good part is done. Viewport and camera are working pretty well. Here is how it is:

gif

Now only the visible tiles need be drawn, but how to do it, since the vertex buffer draws everything?

LithiumToast:
@rafaelalmeidatk  

> Now only the visible tiles need be drawn, but how to do it, since the vertex buffer draws everything?

You don't. That optimization before was to reduce CPU work since each tile has to be looped over every frame. Because the vertices and indices are no longer computed every frame and reside in video memory close to the GPU, there is negligibly CPU work to render the map.

rafaelalmeidatk:
> You don't. That optimization before was to reduce CPU work since each tile has to be looped over every frame. Because the vertices and indices are no longer computed every frame and reside in video memory close to the GPU, there is negligibly CPU work to render the map.

Interesting! If I create a map with more than 33x10 tiles in the develop branch, the map turns black, but now I created an 80x10 map, filled everything with tiles and is running very smooth with this new renderer, even with 128x128 tile size.

I found a bug: if you place some transparent image above the tiles and another one on the background, the image on the top will lose the transparency. Example:

screenshot_41

Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
```diff
@@ -99,16 +175,67 @@ public TiledObjectGroup GetObjectGroup(string name)
...
+            _basicEffect.View = _viewMatrix;
+
+            _graphicsDevice.DepthStencilState = _depthBufferState;
+            _graphicsDevice.BlendState = BlendState.AlphaBlend;
+            _graphicsDevice.SamplerStates[0] = SamplerState.PointClamp;
+
+            // Images draw
+            foreach (var pass in _basicEffect.CurrentTechnique.Passes)
+            {
+                var imageLayers = _layers.Where(layer => layer is TiledImageLayer);
```
LithiumToast: It's not a good idea to constantly query the layers every frame. It be better if the query result was computed and cached as a variable when the layers are initialized or when they change.

LithiumToast:
@rafaelalmeidatk 

>I found a bug: if you place some transparent image above the tiles and another one on the background, the image on the top will lose the transparency.

Probably a depth buffer issue. Read Shawn's blog on depth buffering to understand how things work behind the scenes: https://blogs.msdn.microsoft.com/shawnhar/2009/02/18/depth-sorting-alpha-blended-objects/.

The solution would be to sort layers by their Z value before building the vertices so further away layers have a chance to be rendered first.

Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
rafaelalmeidatk: I solved it, thanks

Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
```diff
@@ -99,16 +177,60 @@ public TiledObjectGroup GetObjectGroup(string name)
...
         }
         
-        public void Draw(SpriteBatch spriteBatch, Rectangle? visibleRectangle = null, GameTime gameTime = null)
+        public void Draw(SpriteBatch spriteBatch, Rectangle? visibleRectangle = null, float zoom = 1.0f, GameTime gameTime = null)
         {
-            foreach (var layer in _layers.Where(i => i.IsVisible))
-                layer.Draw(spriteBatch, visibleRectangle, gameTime: gameTime);
+            var rect = visibleRectangle.HasValue ? visibleRectangle.Value : Rectangle.Empty;
+
+            _worldMatrix.Translation = new Vector3(-rect.X, -rect.Y, 0);
```
LithiumToast: Use this:

```
        public void Draw(Camera2D camera)
        {
            _basicEffect.World = camera.GetViewMatrix();

            _graphicsDevice.DepthStencilState = _depthBufferState;
            _graphicsDevice.BlendState = BlendState.AlphaBlend;
            _graphicsDevice.SamplerStates[0] = SamplerState.PointClamp;
 ...
```

rafaelalmeidatk:
> The solution would be to sort layers by their Z value before building the vertices so further away layers have a chance to be rendered first.

I placed a Z value on every layer and it is working well to place they on the correct order. I am understanding what is happening, the background is drawn first, then the transparent image, then the tiles are drawn in the middle of the two images, right? I can sort the layers by the Z value, but how do I draw only the vertices of some Z value? For example: draw the background image (Z = 100), then the vertices of some tiles (Z = 200), then other image (Z = 300) and then the other vertices of some tiles (Z = 400).

LithiumToast:
@rafaelalmeidatk You would need to keep track of the primitive starting index and count for each layer and use `GraphicsDevice.DrawIndexedPrimitives(...)` to draw the indexed triangles for each layer.

rafaelalmeidatk:
@LithiumToast 
I changed the Draw method a bit and it worked. I have no idea if I am redrawing the tiles more than necessary, but I could not do what you said, sometimes only one layer is drawn perfectly and the others very weird, sometimes only one layer is drawn. I'd read [this](https://msdn.microsoft.com/en-us/library/windows/desktop/bb147325(v=vs.85).aspx) article but nothing.

I created a variable to keep track the vertices and used it as `startIndex`, see:

```
var tilesVerticesSoFar = 0; // It runs before looping every layer

[...]

// It runs for every tile layer
var primitivesCount = tileLayer.VerticesList.Length / 2;

_graphicsDevice.DrawIndexedPrimitives(
    PrimitiveType.TriangleList,
    0,
    tilesVerticesSoFar,
    primitivesCount
);

tilesVerticesSoFar += tileLayer.VerticesList.Length;

[...]
```

LithiumToast:
@rafaelalmeidatk The primitives are triangles. It has to be a multiple of 3 not 2.

See the helper method:

public static class PrimitiveTypeExtensions
    {
        internal static int GetPrimitiveCount(this PrimitiveType primitiveType, int verticesOrIndicesCount)
        {
            switch (primitiveType)
            {
                case PrimitiveType.LineStrip:
                    return verticesOrIndicesCount - 1;
                case PrimitiveType.LineList:
                    return verticesOrIndicesCount / 2;
                case PrimitiveType.TriangleStrip:
                    return verticesOrIndicesCount - 2;
                case PrimitiveType.TriangleList:
                    return verticesOrIndicesCount / 3;
                default:
                    throw new ArgumentException("Invalid primitive type.");
            }
        }
    }

LithiumToast:
@rafaelalmeidatk The article about depth buffering was saying that pixels which rendered and are closer to the screen via their Z values will prevent other pixels from being rendered which are behind. This is most likely the problem you shown in the previous screenshot. To solve this the pixels which are behind other objects need to be rendered first. This can be achieved by filling the vertices into the vertex buffer starting with the furthest way layer from the screen and working towards the closest layer to the screen. Multiple draw calls are not needed unless the transformation matrix needs to be applied per layer.

rafaelalmeidatk:
@LithiumToast 
> The primitives are triangles. It has to be a multiple of 3 not 2.

I changed the code to `var primitivesCount = tileLayer.VerticesList.Length / 3;` and the result was [that](http://i.imgur.com/oDeAAwZ.png). If I have 4 vertices for each tile, then `4 / 2 = 2`, so I have 2 primitives, it is, 2 triangles, or I am confounding something?

> Multiple draw calls are not needed unless the transformation matrix needs to be applied per layer.

The images vertices don't use a vertex buffer, they are drawn using the `DrawUserIndexedPrimitives` because I need change the texture of the effect before drawing it. Because it, I need split the draws for every layer. The only problem now is that I don't know how to draw the tiles vertices separately for each layer.

LithiumToast:
@rafaelalmeidatk It's not the vertices count which need to be used... It's the indices count. Yes there is 4 vertices, but for each triangle there needs to be 3 vertices. We use indices to reference the vertices. Take a look at my original post with the diagrams. I talk about the indices there.

rafaelalmeidatk:
@LithiumToast I am not understanding what I need put on DrawIndexedPrimitives, it's very confused to me. The primitive type is TriangleList, ok. Then i need specify the baseVertex, I still don't know exactly what it is, but I guess it is from where the Index Count stopped? So, for the first layer it is 0, for the second it is the index count from the last layer more one. I guess the same to the startIndex. The primitiveCount I know what is now, it's the number of indices divided by 3, correct?

I made it (running on each layer):
```
var indexCount = tileLayer.Tiles.ToList().Count * 6; // Each tile has 6 indexes
var primitivesCount = PrimitiveTypeExtensions.GetPrimitiveCount(PrimitiveType.TriangleList, indexCount);
_graphicsDevice.DrawIndexedPrimitives(
    PrimitiveType.TriangleList,
    tilesIndexesSoFar,
    tilesIndexesSoFar,
    primitivesCount
);

tilesIndexesSoFar += indexCount + 1;
```

It draws the map correctly, with exception of some random tiles that are drawn on it:
screenshot_45
Every time I rerun the game, this random tiles are in some different position.

LithiumToast:
@rafaelalmeidatk Check out the [XNA msdn documentation on `DrawIndexedPrimitives`](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.graphicsdevice.drawindexedprimitives.aspx).

primitiveType: We need `PrimitiveType.TriangleList` because that's how the vertex buffer and index buffer are filled, with triangles (2 triangles per tile). 
baseVertex: I believe the code applies the vertex offset when building the indices so this can be left as 0. This is useful if the vertex offset was not or can not be pre-applied and indices are referencing a different subset of the vertices.
minVertexIndex: Not applicable to MonoGame.
numVertices: Not applicable to MonoGame.
startIndex: The beginning of the index buffer to use.
primitiveCount: The number of primitives to render. Because the primitive type is triangle list, this is really the number of triangles to render.

rafaelalmeidatk:
@LithiumToast 
> I believe the code already applies the vertex offset when building the indices, right?

I think so, if by "vertex offset" you mean this `index`:

```
var index = 0;
foreach (var layer in tileLayers)
{
    tileVertices.AddRange(layer.RenderVertices());
    var tilesCount = layer.Tiles.Where(x => x.Id != 0).ToList().Count;
    for (var i = 0; i < tilesCount; i++)
    {
        var thisTileIndexes = new short[6];
        thisTileIndexes[0] = (short)(4 * index);
        thisTileIndexes[1] = (short)(4 * index + 1);
        thisTileIndexes[2] = (short)(4 * index + 2);
        thisTileIndexes[3] = (short)(4 * index + 1);
        thisTileIndexes[4] = (short)(4 * index + 3);
        thisTileIndexes[5] = (short)(4 * index + 2);
        tileIndexes.AddRange(thisTileIndexes);
        _tilesPrimitivesCount += 2;
        index++;
    }
}
```
Well, I changed the code to this:
```
var tilesCount = tileLayer.Tiles.ToList().Count;
var indexCount = tilesCount * 6;
var primitivesCount = tilesCount * 2;
_graphicsDevice.DrawIndexedPrimitives(
    PrimitiveType.TriangleList,
    0,
    tilesIndexesSoFar,
    primitivesCount
);
tilesIndexesSoFar += indexCount + 1;
```
The map is still entirely drawn, but now the "random tiles" are even more weird:
screenshot_46

LithiumToast:
```
thisTileIndexes[0] = (short)(4 * index);
```

Why multiply by four?

Source/Demos/Demo.TiledMaps/Game1.cs
```diff
@@ -101,7 +101,7 @@ protected override void Draw(GameTime gameTime)
             _spriteBatch.Begin(transformMatrix: _camera.GetViewMatrix());
 
             // you can draw the whole map all at once
-            _spriteBatch.Draw(_tiledMap, gameTime: gameTime);
+            _spriteBatch.Draw(_tiledMap, _camera, gameTime: gameTime);
```
LithiumToast: The map doesn't need a `SpriteBatch` to draw anymore.

Source/MonoGame.Extended/Maps/Tiled/SpriteBatchExtensions.cs
```diff
@@ -6,11 +6,6 @@ namespace MonoGame.Extended.Maps.Tiled
 {
     public static class SpriteBatchExtensions
     {
-        public static void Draw(this SpriteBatch spriteBatch, TiledMap tiledMap, Rectangle? visibleRectangle = null, GameTime gameTime = null)
-        {
-            tiledMap.Draw(spriteBatch, visibleRectangle, gameTime);
-        }
-
         public static void Draw(this SpriteBatch spriteBatch, TiledMap tiledMap, Camera2D camera, GameTime gameTime = null)
```
LithiumToast: Same here; map doesn't use a `SpriteBatch` to draw anymore.

Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
```diff
@@ -83,6 +100,68 @@ public TiledObjectGroup CreateObjectGroup(string name, TiledObject[] objects, bo
             return objectGroup;
         }
 
+        public TiledMap Build()
+        {
+            var tileVertices = new List<VertexPositionTexture>();
+            var tileIndexes = new List<short>();
+            var tileLayers = _layers.OfType<TiledTileLayer>();
```
LithiumToast: Sort the layers here by Z value to fix the depth buffer issue. You want layers which are closer to the screen to processed last.

LithiumToast:
@rafaelalmeidatk Just use the working code you had before and sort the tile layers by the Z value before you build the vertices and indices for the map.

rafaelalmeidatk:
@LithiumToast When the method Build is called, the layers already are sorted by the Z value. I don't think the working code is a good idea because it seems to draw the entire map in the loop for every tile layer, instead of drawing only the vertices of it layer.

LithiumToast:
@rafaelalmeidatk Are the tile layers already sorted back to front?

> it seems to draw the entire map in the loop for every tile layer

So when the vertices and indices are added to the arrays for a tile layer just take note of the index count and the primitive count for the tile layer so you pass that to `DrawIndexedPrimitives()` method to draw the subset of the vertices and indices for each tile layer.

rafaelalmeidatk:
@LithiumToast Wow, what a stupid mistake I was doing :P
I think the reader is ready, I just wan't to point that now each map can only have one tileset. The limit is due to the fact that I can't change the textures when drawing the tiles. This problem have some solutions, like creating a new texture that is the union of all tilesets used in the map, is this a good idea?

craftworkgames:
First let me thank both of you guys @rafaelalmeidatk and @LithiumToast for making great progress on this. It will really improve the most popular feature of the library. It's shaping up to be one one of the best Tiled renderers out there :)

> The limit is due to the fact that I can't change the textures when drawing the tiles. This problem have some solutions, like creating a new texture that is the union of all tilesets used in the map, is this a good idea?

I'm sure we can solve this problem but I don't think creating a union of the tileset textures is the right way to go about it. I haven't taken a look at the code but the general idea is to group the rendering into "batches" in much the same way a sprite batch would do it.

Getting it working is the first step, making it efficient is another. We may have to make a few passes at refactoring the code to get it right but it should be possible to optimize it because we know the constraints of the input data.

Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
```diff
@@ -80,19 +83,19 @@ public List<VertexPositionTexture> RenderVertices()
                 }
                 var vertices = new VertexPositionTexture[4];
                 vertices[0] = new VertexPositionTexture(
-                    new Vector3(tile.X * TileWidth - 1, tile.Y * TileHeight, Z),
+                    new Vector3(tile.X * TileWidth - 0.9f, tile.Y * TileHeight, Z),
```
LithiumToast: Why is there a position offset of 0.9 applied along the X-axis?

LithiumToast:
The easiest way to have multiple textures is to just switch the active texture when needed.

```
effect.Texture = ...;
```

However, this is not a the best solution as it's a relatively expensive operation in most cases to bind a texture and upload it to the GPU (this happens on `effect.Apply()`). It's possible that the textures would need to be swapped in and out many times, potentially killing performance.

One more complicated solution would be to build a giant texture atlas for all the tiles. Since the tiles from different tilesets are most likely the same size, it's fairly easy to take all the texture information of all the tiles across multiple tilesets and place them into a new giant texture. The texture coordinates of the tile in the larger texture can be found by using a texture atlas similar to sprite sheets. The maximum number of tiles that can be placed in a texture is limited by the maximum texture size. I think it's safe to assume that the lowest texture size that is supported on all hardware is 2048x2048 pixels since this was the limit for XNA reach profile. If each tile is 16x16 pixels let's say, then there can fit 128 * 128 = 2^7 * 2^7 = 2^14 = 16384 tiles in the single large texture. For most games it would probably take one of these large texture atlas to fit in all the tiles necessary to render the map. If there are a lot of tiles used in a game, different maps which don't have the same tilesets could use different texture atlases.

Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
rafaelalmeidatk: `0.5f` was still showing some gaps, so I changed to `1f`, but this offset was visible with some zoom-in, so I ended with `0.9f` that don't show the gaps and isn't visible. I forgot to change the offset on the projection matrix, I will do it now.

rafaelalmeidatk:
Union the tilesets can be expansive if someone uses two very larges tilesets, but what about creating a new texture that is the union of only the utilized tiles?

Other solution I through is divide the tiles list of the layers by tilesets, so if some layer has more than one tileset, we draw this tiles separately. The problem is what @LithiumToast mentioned, change the texture so many times can be expensive, but instead of be per tile, this would be per layer, and is rare the cases when someone use more than, for example, 3 differents tilesets.

craftworkgames:
> It's possible that the textures would need to be swapped in and out many times, potentially killing performance.

It's certainly possible that performance will be killed by texture swapping but this is in the user's control. If the user is sensible when they are creating their tileset images the number of texture swaps will be minimal.

I'm not saying it *definitely won't be a problem*. It certainly could be, but I don't think it's a problem we should be worrying about too much. Implementing a complicated solution could turn out to be a huge waste of time and create new bugs that we didn't anticipate.

LithiumToast:
@craftworkgames Yeah I agree; for now just swap the textures when needed.

rafaelalmeidatk:
So, can I do the second idea (separate the tiles by tileset on each layer and swap the textures when necessary)?

I noticed a strange behavior: if the map has exactly 5 layers (layer of tiles or images), the last one (with higher Z value) isn't drawn. I have no idea why this happens, if one more layer is added (even if it's empty) all the layers are drawn.

craftworkgames:
> So, can I do the second idea (separate the tiles by tileset on each layer and swap the textures when necessary)?

The tricky part is making sure we can still support all of the different map types in Tiled. Isometric maps in particular need tiles rendered in a certain order to look right. Although, there might be some other ways we can achieve this.

In terms of testing, it's probably a good idea to create a bunch of different maps that cover as many of the Tiled features we can and keep those maps in source control. Perhaps as part of the Tiled demo. The primary goal is to be able to easily test the different Tiled features so we know we haven't broken anything.

Once we have good testing in place, refactoring the code and adding optimizations becomes a lot less stressful.

rafaelalmeidatk:
Sure, I will try to successfully render an Isometric map, just need learn how they works.

rafaelalmeidatk:
Done! I will work on the rendering order when I have more time.
screenshot_1

craftworkgames:
It's looking really great :)

I'm definitely going to include this in the next official release. The plan is to get this and the GUI system working and as many of the bug fixes as possible. 

If you do stop working on this just let me know what still needs to be done and I'll go over it after the merge. I'm hoping to improve the Tiled demo we have a good story around regression testing.

rafaelalmeidatk:
The support to the render order is ready. I still think that increase the size of the vertices isn't a good solution for the gaps. It's working for default sizes, but when I put my game on fullscreen, there are lines on the screen  (as you can see [here](https://cloud.githubusercontent.com/assets/11896398/17009270/6dcecc8c-4ecf-11e6-9270-1990dae69de4.png)), and they keep appearing on the whole map when you move the camera. However, if I remove the increased amount, the lines appear way more often. What could I do?

LithiumToast:
@rafaelalmeidatk What does the projection, world and view matrix look like?

Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
```diff
@@ -83,6 +99,68 @@ public TiledObjectGroup CreateObjectGroup(string name, TiledObject[] objects, bo
...
+
+            var highestZ = _layers.Max(layer => layer.Z);
+            _viewMatrix = Matrix.CreateLookAt(
+                new Vector3(0f, 0f, highestZ + 100.0f),
+                Vector3.Zero,
+                Vector3.Up
+            );
+
+            _projectionMatrix = 
+                Matrix.CreateTranslation(-0.9f, -0.9f, 0f) * Matrix.CreateOrthographicOffCenter(
```
LithiumToast: Translate by this amount, not -0.9f.

```
Matrix.CreateTranslation(-0.5f, -0.5f, 0)
```

LithiumToast:
@rafaelalmeidatk Try `-0.5f` instead of `-0.9f` when translating the projection matrix.

rafaelalmeidatk:
@LithiumToast No differences (if any, very imperceptibles), but this parameters seems to change some things, you can see [here](https://cloud.githubusercontent.com/assets/11896398/17045626/6daacfc6-4fa1-11e6-836e-a95ce1ed1d74.png) that now the lines also appear in the default window size. But why does it? Isn't these parameters only to translate the projection? After all, what are this lines?

LithiumToast:
@rafaelalmeidatk What does it look like when zooming the camera in and out?

craftworkgames:
> TriangleList is the way to go here because TriangleStrip would require two tiles to share vertex information which is not always possible since two tiles could be using different texture coordinates. 

Is this really true? It seems to me that using a connected mesh would solve these issues we're having with gaps. When we render the map as a triangle list, we're not doing much better than using the `SpriteBatch` I would think. Unless I'm misunderstanding something.

> Try -0.5f instead of -0.9f when translating the projection matrix.

Fiddling with the projection matrix like this is never really going to work. There's only one correct answer here and it's probably going to be -0.5f or 0.0f. Using an strange offset like 0.9f is only going to make it *appear* like things are working, until they're not again.

If you guys are okay with it I think I should pull these changes into the `develop` branch. I want the new renderer in the next release and I'm sure we can get it working. Once I pull it I'll take a closer look at the remaining bugs and see if I can help to fix them.

LithiumToast:
@craftworkgames Yes, it's impossible to have a triangle strip for tiles since neighbouring tiles can share different texture coordinates. In other words, two neighbouring tiles can use two different texture regions. The only vertices which can be connected are the two triangles which make up each tile. The gaps problem is solved anyways by explicitly setting `PointClamp` for  the sampler state.

`SpriteBatch` does use triangle list for graphic primitives but the main difference here is that this renderer has the vertices and indices in a static vertex buffer and static index buffer. Drawing the map using `SpriteBatch` requires that the CPU loops over every tile every frame to submit it to the dynamic vertex and index buffer handled by `SpriteBatch`. Drawing the map now only requires looping over all the layers and drawing the appropriate vertices and indices from the static vertex and index buffer. Static buffers also have better performance in general than dynamic buffers.

As for the projection offset, `SpriteBatch` optionally adds the `-0.5f` translation if needed. See [here](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/SpriteBatch.cs#L140). Other than that and the sampler state, the renderer should behave exactly as before when using `SpriteBatch`. Unless, the vertices and/or indices are computed in a wrong way.

craftworkgames:
@LithiumToast Okay great. That all makes sense.

So what is left to do before pulling into `develop`?

rafaelalmeidatk:
@LithiumToast Here is it: http://i.imgur.com/KY9s3c9.gif
The lines don't show when the size is proportional to the `BoxingViewportAdapter` (virtual: 360x240). I tried to reproduce this problem in the Tiled demo but I could not, maybe it is an issue only in my game?
@craftworkgames I think the only thing left is the animated tiles, idk when I will do it.

craftworkgames:
>  I tried to reproduce this problem in the Tiled demo but I could not, maybe it is an issue only in my game?

That's unfortunate. If you do figure out the problem please let us know. If we can't reproduce the issue it makes it very difficult to fix.

> I think the only thing left is the animated tiles, idk when I will do it.

Okay cool. I really appreciate all the work you've done so far. I'm not overly concerned if you don't have time to finish everything the important thing is that we have a kind of TODO list of the remaining work.

When I get around to it I'd like to merge this PR and improve the Tiled demo to cover all of the features we support. This way we'll have a way to test everything as we move forward and a clearer understanding of what needs to be done for the next release.

LithiumToast:
@rafaelalmeidatk Read [this](http://drilian.com/2008/11/25/understanding-half-pixel-and-half-texel-offsets/) and [this](https://msdn.microsoft.com/en-us/library/windows/desktop/bb219690(v=vs.85).aspx).

You could try this: 

```
halfPixel.X = 0.5f / (float)GraphicsDevice.PresentationParameters.BackBufferWidth;
halfPixel.Y = 0.5f / (float)GraphicsDevice.PresentationParameters.BackBufferHeight;
```

rafaelalmeidatk:
@LithiumToast Sorry, I could not understand this articles. I tried this:

    var halfPixel = new Vector2();
    halfPixel.X = 0.5f / (float)_graphicsDevice.PresentationParameters.BackBufferWidth;
    halfPixel.Y = 0.5f / (float)_graphicsDevice.PresentationParameters.BackBufferHeight;

    _projectionMatrix = 
        Matrix.CreateTranslation(halfPixel.X, halfPixel.Y, 0f) * Matrix.CreateOrthographicOffCenter(
            0,
            _graphicsDevice.Viewport.Width,
            _graphicsDevice.Viewport.Height,
            0,
            1.0f, highestZ + 100.0f
        );

And this (on the creation of the vertices):

    for (var i = 0; i < vertices.Length; i++)
    {
        vertices[i].Position.X -= 0.5f;
        vertices[i].Position.Y -= 0.5f;
    }

No success. Do I have to use shaders?

@craftworkgames The list:
- Animated tiles
- Solve this tiles gaps and side effects once for (most important)

LithiumToast:
@rafaelalmeidatk You mentioned that the Tiled demo didn't have artifacts like in your game? What platform are you using? If Windows, are you using DirectX or OpenGL? I suspect that using no translation is going to fix it for your case and in every other case the `0.5f` projection translation is going to fix the problem.

rafaelalmeidatk:
@LithiumToast Yes, I am using OpenGL. I made a test with the Tiled demo using 32x32 tiles. Now the gaps are showing again, even with the `0.9f` offset for each vertex side, the most weird is that this doesn't occur with the other map that has 124x124 tiles. Do I need to insert the `-0.5f` translation to the projection and then remove the offset from the vertex positions? If so, the gaps will appear again.

LithiumToast:
@rafaelalmeidatk Where do you get `-0.9f` from? The projection matrix should have a translation of `x:-0.5f,y:-0.5f` or `x:0,y:0` depending on the platform. There is no need compute an offset to the vertices themselves... That's the responsibility of vertex shader. 

The gaps problem is a different problem which is solved by using point clamp texture sampling... Adding a translation to the projection matrix should not be related to the gaps problem at all.

craftworkgames:
> I made a test with the Tiled demo using 32x32 tiles. Now the gaps are showing again

Please keep that code handy. If we find ways to reproduce bugs like this I'd like to keep to code around so that we have a strong test suite.

rafaelalmeidatk:
@LithiumToast The `0.9f` was to expand the vertex positions to don't show the gaps (forcing the overlap), I removed it now. With the `-0.5f` translation on the projection matrix, the gaps don't show (I made other thing to it works, I'll comment bellow). But if I zoom-in or zoom-out, they appear again.

> The gaps problem is a different problem which is solved by using point clamp texture sampling

But it isn't. What I made now was fix the camera position like this:

    camera.Position = new Vector2((int)camera.Position.X, (int)camera.Position.Y);

It solves the problem, but only when the zoom is an integer value, the floating values (like `1.075`) make the vertex positions wrong and show the gaps sometimes, see: http://i.imgur.com/ZRh5Gnp.gif

LithiumToast:
Okay so first off there is some confusion as to what "gaps" mean.

This screenshot shows gaps and is fixed by using `PointClamp` sampler state.
screen shot 2016-07-28 at 10 18 47 am

This screenshot shows texture bleeding and is the current problem.
screen shot 2016-07-28 at 9 27 04 am

LithiumToast:
@rafaelalmeidatk Use this for calculating the vertex positions of a tile to fix bleeding.

```
                vertices[0] = new VertexPositionTexture(
                    new Vector3(point.X, point.Y, Z) + new Vector3(-0.5f, 0, 0),
                    textureCoordinateTopLeft
                );
                vertices[1] = new VertexPositionTexture(
                    new Vector3(point.X + tileWidth, point.Y, Z) + new Vector3(0.5f, 0, 0),
                    new Vector2(textureCoordinateBottomRight.X, textureCoordinateTopLeft.Y)
                );
                vertices[2] = new VertexPositionTexture(
                    new Vector3(point.X, point.Y + tileHeight, Z) + new Vector3(-0.5f, 0, 0),
                    new Vector2(textureCoordinateTopLeft.X, textureCoordinateBottomRight.Y)
                );
                vertices[3] = new VertexPositionTexture(
                    new Vector3(point.X + tileWidth, point.Y + tileHeight, Z) + new Vector3(0.5f, 0, 0),
                    textureCoordinateBottomRight
                );
```

rafaelalmeidatk:
Got it working! Wow, the solution was too simple, thanks @LithiumToast. I was changing the position of the vertex, after searching for texture bleeding, I realized that I should change the UV coordinates instead. Now this lines don't show anymore, even with a very minimal zoom. The `-0.5f` translation on the projection matrix isn't needed too, but as you said, the SpriteBatch does it. How can I check if the platform is OpenGL to add this `-0.5f` offset too (it only adds when the game is using OpenGL, see [here](https://github.com/mono/MonoGame/blob/e7345ba69423fa420a19a7d0b030a15df2cea7fe/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs#L203))?

@craftworkgames Now the only thing left are the animation sprites, but I think it is ready to merge ;D

LithiumToast:
@rafaelalmeidatk I'm not sure. You can go spelunking in MonoGame's [`SpriteBatch`](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/SpriteBatch.cs) to figure out how they conditional add the projection offset.

craftworkgames:
> Now the only thing left are the animated sprites, but I think it is ready to merge ;D

I was waiting for a good time to merge this (when I have time to properly test it) but I don't think that's going to happen for a while so let's just get it over with and do it now :)

craftworkgames:
So obviously there's still going to be a few things left to do:
 - update the demos to include any features not yet demoed.
 - close any issues resolved by this PR.
 - open and new issues created by this PR (e.g. animated tiles)

craftworkgames:
- there's something messed up with the camera on the new renderer

rafaelalmeidatk:
@craftworkgames Sorry, I was clamping the position of the camera to a int value in an attempt to solve the texture bleeding problem a while ago. I will submit a PR to remove it.

PR #225, opened 2016-07-30
correct width calculation for bitmap fonts

fabienbrooke:
I found an issue while trying to position text according to its width. There would be an extra amount of width, coming from the `XAdvance` property for the last character. Since the last character is not advancing to another we should use the sum of the character region's `XOffset` and `Width` instead.

### DIFF
--- a/Source/MonoGame.Extended/BitmapFonts/BitmapFont.cs
+++ b/Source/MonoGame.Extended/BitmapFonts/BitmapFont.cs
@@ -45,14 +45,19 @@ namespace MonoGame.Extended.BitmapFonts
         {
             var width = 0;
             var height = 0;
+            var codePoints = GetUnicodeCodePoints(text).ToArray();
 
-            foreach (int c in GetUnicodeCodePoints(text))
+            for (int i = 0, l = codePoints.Length; i < l; i++)
             {
                 BitmapFontRegion fontRegion;
+                var c = codePoints[i];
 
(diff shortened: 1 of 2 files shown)


fabienbrooke:
We could also add a `LetterSpacing` property to BitmapFont and also add that while adding the `XAdvance` value. Since that is a new feature rather than a fix I've left it out of this PR though. Let me know if that's something you'd like too.

craftworkgames:
> I found an issue while trying to position text according to its width. 

Thanks for fixing this. The changes look great. Have you tested it with the demo?

> We could also add a public LetterSpacing property to BitmapFont and add that amount in while adding the XAdvance value too. 

Seems like a reasonable feature. If I understand it correctly, it wouldn't be a breaking change if the default value is zero. I'm happy if you want to make another PR for it. Ideally, I'd like to get to a stage where we have every feature included in the demos to make them easy to test.

fabienbrooke:
Yeah I ran the demo to check there was no regression. Mouse coords are centered below paragraph text correctly.

fabienbrooke:
And yeah for LetterSpacing it would be backwards compatible. Following from the CSS description:
> Indicates inter-character space **in addition to** the default space between characters.
Can add a line of text in the bitmap font demo that shows wide spacing.

craftworkgames:
> I'll make the new PR after this one is merged.

Yes, please do.

I'll merge this one now.

PR #226, opened 2016-08-02
Fix camera position

rafaelalmeidatk:


Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
```diff
@@ -185,7 +185,6 @@ public TiledObjectGroup GetObjectGroup(string name)
         
         public void Draw(Camera2D camera, GameTime gameTime = null)
         {
-            camera.Position = new Vector2((int)camera.Position.X, (int)camera.Position.Y);
```
craftworkgames: This isn't the problem. I've already tried removing this line.

craftworkgames:
I believe the problem has something to do with the way the `World`, `View` or `Projection` matrices are being calculated.

You can see it in the Tiled demo (and some of the other demos) if you resize the window.

I'd expect the visible part of the map to scale but it should remain visible.

craftworkgames:
This code here is probably where I'd start looking:

            _projectionMatrix = 
                Matrix.CreateTranslation(projectionHalfPixel) * Matrix.CreateOrthographicOffCenter(
                    0,
                    _graphicsDevice.Viewport.Width,
                    _graphicsDevice.Viewport.Height,
                    0,
                    1.0f, highestZ + 100.0f
                );

When the window is resized, the viewport size will change, so I *think* the projection matrix will need to be recalculated somehow.

craftworkgames:
I just noticed if I render the map like this:

    _tiledMap.Draw(new Camera2D(GraphicsDevice), gameTime);

It seems to produce a pretty good result. At this point, resizing the window also resizes the map. However, this is not respecting the camera as I have no viewport adapter at all. Definitely not what I was expecting.

craftworkgames:
Okay I think I fixed the most of the issues. Hopefully I didn't break anything. If you could take a look and see if it all still works for you that'd be great.

No need for this PR. I'll close it.

PR #233, opened 2016-08-10
Add LetterSpacing to BitmapFont

fabienbrooke:
As mentioned in https://github.com/craftworkgames/MonoGame.Extended/pull/225.

Adds letter spacing support to the BitmapFont class through a public `LetterSpacing` property, defaulting to 0 and thus backwards compatible.

LetterSpacing indicates inter-character space in addition to the default space between characters.


craftworkgames:
Great work.

PR #234, opened 2016-08-12
SceneGraph Optimization: Cache and Lazy Evaluate World and Local Transform Matrices 

LithiumToast:
When a scene graph gets large, computing the world and local transform for scene nodes every time it's requested is not ideal. This fixes that by caching the matrices and updating them on request only if they are dirty. The logic is separated out into a `BaseTransform` class which `Transform2D` and `Transform3D` inherit. `SceneNode` uses `Transform2D`.
Tested to ensure functionality is the same with SceneGraph demo.

EDIT: I accidentally started this from `graphics` branch. I can redo this pull request from `develop` or, if it's okay with you, the `graphics` branch will be merged into `develop` with this pull request.


craftworkgames:
There's way too many changes here for me to reasonably consider without adequate testing.

> and Transform3D inherit.

It concerns me that you've started evolving the library down a 3D path. I wasn't planning on doing that anytime soon.

LithiumToast:
I'll close this and re-do it from `develop` branch later this week so this only has the necessary changes.

> It concerns me that you've started evolving the library down a 3D path.

Well, considering there is a `Camera2D`, I thought the standard would be to add a suffix to classes, interfaces, etc with "2D" or "3D". This discussion belongs in an issue for sure. I can remove `Transform3D` when I re-do this though.

craftworkgames:
> I thought the standard would be to add a suffix to classes, interfaces, etc with "2D" or "3D". This discussion belongs in an issue for sure. 

Yeah. I remember the other day you mentioned renaming the `Camera2D` to `OrthographicCamera`. While I agree that's probably a better name, changes like this really does need to be discussed first.

I really like the work you're doing on the library. It's great stuff. But I'm struggling to handle the big incoming PR's. I really don't know what to do with them.

PR #237, opened 2016-08-20
Matrix2D & Transform2D

LithiumToast:
-Add Matrix2D
-Add Transform2D
-Add Math namespace (without namespace rule, i.e. all files in the folder "Math" will still have the namespace "MonoGame.Extended")
-Add `float` helper extensions to `FloatHelper` in Math namespace.
-Move/rename `Vector2Extensions` to `Vector2Helper` to the Math namespace.

I had some trouble getting the platform demo to build. It's complaining that `SpriteComponent` is missing. I commented out the build errors.


craftworkgames:
Please don't do any of these things:

-Add Math namespace (without namespace rule, i.e. all files in the folder "Math" will still have the namespace "MonoGame.Extended")
-Add float helper extensions to FloatHelper in Math namespace.
-Move/rename Vector2Extensions to Vector2Helper to the Math namespace.

> I had some trouble getting the platform demo to build. It's complaining that SpriteComponent is missing. I commented out the build errors.

Gah, my bad. I fixed these things locally and forgot to push. It's resolved now.

craftworkgames:
Almost everything in `FloatHelper` can be done with the standard `Math` or `MathHelper`. I'm not convinced it's worth having it at all.

LithiumToast:
How's that?

craftworkgames:
> How's that?

Okay, let's use a few examples from `FloatHelper`.

            var x = 123.4f.Round();
            var y = (float)Math.Round(123.4f);

Most of the methods (e.g. `Round`) are just changing one line of code from what people are used to seeing to a slight variation in syntax. In my opinion, this doesn't really buy us anything. It adds a little bit of connotative load to anyone reading the code, bloats the library with crap and in my experience is generally considered bad practice to do such trivial extension methods on primitive types.

The `Clamp` extension already exists in MonoGame's `MathHelper`.

            var a = 123.4f.Clamp(200f, 300f);
            var b = MathHelper.Clamp(123.4f, 200f, 300f);

The only slightly interesting one is `ApproximatelyEqualTo`. You've renamed it from `EqualsWithTolerance` which I agree isn't the nicest name around, but it does have one advantage in that it shows up next to `Equals` with intellisense.

            var q = 123.4f.Equals(123.5f);
            var r = 123.4f.ApproximatelyEqualTo(123.5f, tolerance: 0.1f);
            var s = 123.4f.EqualsWithTolerance(123.5f, tolerance: 0.1f);

So as a general rule I'd prefer not to have extension methods on primitive types like `float`. Although, I realize in some non-trivial cases it would be nice to have a handy helper method. Maybe it would be better to have our own `MathHelperExtended` static class to compliment `MathHelper`.

LithiumToast:
Alright.

Source/MonoGame.Extended/Transform.cs
```diff
@@ -0,0 +1,281 @@
...
+            }
+        }
+
+        /// <summary>
+        ///     Gets or sets the rotation angle.
+        /// </summary>
+        /// <value>
+        ///     The rotation angle.
+        /// </value>
+        public float RotationAngle
```
craftworkgames: I think we should just call this `Rotation` to be consistent with other parts of the framework. Also, since you've got documentation, we should probably mention that it's in radians.

craftworkgames:
Very nice. 

I've made one small comment about renaming `RotationAngle` to `Rotation`.

I'll merge this tonight. If you don't get to the rename I'll do it after the merge.

LithiumToast:
I guess it's okay to call it just `Rotation` for 2D since the rotation axis is always the Z-axis. However when we get to 3D, a rotation is defined by an axis and and angle in radians, `RotationAxis` and `RotationAngle` respectively. That's why in 2D I also called it `RotationAngle`.

craftworkgames:
> when we get to 3D

When we get to 3D there's going to be lot's of things to think about but considering this is the `Rotation` property on the `Transform2D` class we've already implied that it's 2D only.

craftworkgames:
> a rotation is defined by an axis and and angle in radians, RotationAxis and RotationAngle respectively.

I'd also be okay with `Rotation` and `RotationAxis` respectively. I know it's not as explicit but I still think it's pretty intuitive and quite common in other frameworks.

PR #239, opened 2016-08-22
Fix System.FormatException thrown during float parsing

Noxalus:
Hello,

I'm on a french environment, and I have this System.FormatException when I try to run some examples due to float parsing (in french, float numbers use comma). This commit is enough to solve this issue.

Keep up the good work ;)


craftworkgames:
Thanks :smile:

PR #240, opened 2016-08-22
SceneNode & Transform2D

LithiumToast:
- Fix bug in `Transform2D` where local matrix wouldn't be updated if the parent's local matrix updated.
- Fix `Scale` and `Rotation` property in `Matrix2D`.
- Add `WorldPosition`, `WorldRotation`, and `WorldScale` to `Transform2D`.
- Update `SceneNode` to use `Transform2D` as base class.
- Update SceneGraph demo; works as expected.

PR #242, opened 2016-08-23
Transform

LithiumToast:
Same as #240.

Turns out the problem was one line in `SceneNode`:

```
...
                    var scale = WorldScale + drawable.Scale;

                    spriteBatch.Draw(texture, position, sourceRectangle, drawable.Color, rotation, drawable.Origin, scale, drawable.Effect, 0);
```

Fix was:

```
var scale = WorldScale * drawable.Scale;
```


craftworkgames:
There's still a problem.

It looks like the wheels of the car are rotating around the wrong way.

PR #245, opened 2016-08-25
SizeF class implicit operator added

Aurioch:
Added implicit operator to Monogame.Extended.SizeF which casts Vector2 to SizeF.

### DIFF
--- a/Source/MonoGame.Extended.Tests/Shapes/RectangleFTests.cs
+++ b/Source/MonoGame.Extended.Tests/Shapes/RectangleFTests.cs
@@ -33,5 +33,14 @@ namespace MonoGame.Extended.Tests.Shapes
 
             Assert.AreEqual(rect1, rect2);
         }
+
+        [Test]
+        public void PassPointAsConstructorParameter_Test()
+        {
+            var rect1 = new RectangleF(new Vector2(0, 0), new SizeF(12, 56));
+            var rect2 = new RectangleF(new Vector2(0, 0), new Point(12, 56));
+
+            Assert.AreEqual(rect1, rect2);
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


craftworkgames:
In gitter you mentioned that

> Heh. I also noticed that RectangleF doesn't have a constructor which uses either Point or Vector2, so it might be good idea to add those two for the sake of convenience.

Could you add another `RectangleF` unit test to prove that the constructor already works if you pass it a `Vector2` or `Point`?

Aurioch:
Test added and test passes. I hope I've written correct test, though.

Source/MonoGame.Extended.Tests/Shapes/RectangleFTests.cs
```diff
@@ -24,5 +24,14 @@ public void Rectangle_Intersects_Test()
...
         }
+
+        [Test]
+        public void PassVector2AsConstructorParameter_Test()
+        {
+            var rect1 = new RectangleF(new Vector2(0, 0), new SizeF(12.34f, 56.78f));
+            var rect2 = new RectangleF(new Vector2(0, 0), new Vector2(12.34f, 56.78f));
+
+            Assert.AreEqual(rect1, rect2);
+        }
```
craftworkgames: Looks great :)

craftworkgames:
> Test added and test passes. I hope I've written correct test, though.

The trick to knowing if the test is correct is to think about what would make the test fail. For example, in this case, if the implicit conversion didn't exist on the `SizeF` class would the test fail? You could try this by removing the implementation of the implicit conversion (just throw NotImplementedExteption).

In an ideal world, we'd actually write the test first, make sure it fails, then implement the code to make it pass. This is called Test Driven Development (TDD). Some people take this to the extreme, but in reality it's quite difficult to get into this habit.

craftworkgames:
Anyway, I'm good to pull this if you're ready?

Aurioch:
Thanks for the info. Just tried it as you suggested and test fails where it should fail.

So... I'm done here. ^^

craftworkgames:
Thanks for that. It was fun :smile:

PR #246, opened 2016-08-27
Objectgroups for Tiles!

TheUltimateHose:
Those changes make the "per tile ObjectGroups" available. They are used by Tiled for tile collision using Objects.

### DIFF
--- a/Source/MonoGame.Extended.Content.Pipeline/Tiled/TiledMapWriter.cs
+++ b/Source/MonoGame.Extended.Content.Pipeline/Tiled/TiledMapWriter.cs
@@ -47,6 +47,8 @@ namespace MonoGame.Extended.Content.Pipeline.Tiled
                         writer.Write(frame.Duration);
                     }
                     WriteCustomProperties(writer, tile.Properties);
+
+                    WriteObjectGroups(writer, tile.ObjectGroups);
                 }
                 WriteCustomProperties(writer, tileset.Properties);
             }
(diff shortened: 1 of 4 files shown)


craftworkgames:
Okay, so these changes clearly haven't been tested properly. The Tiled demo was broken and there's clearly some bugs (like a null reference exception in TmxTile).

I've reverted the changes out of `develop` for now.

LithiumToast:
@craftworkgames Hey, just a tip. You could try forking the pull-request to test it out rather than merging it first.

craftworkgames:
> Hey, just a tip. You could try cloning or forking the pull-request to test it out rather than merging it first.

That's a good idea. I'll keep it in mind for the future.

However, there's no excuse for not testing properly. I don't expect things to be perfect, but some reasonable level of testing is expected for every PR. 

It's quite frustrating for me when things don't work. It can sometimes take hours to figure out what's going on. I first have to figure out what the code is supposed do, then I have to figure out why it's not working, then I have to try and fix it. That's time I could be spending on more important things.

In an ideal world, we'd have a unit test or a demo for *everything*. Some open source projects require this level of testing and I can see why. It raises quality bar and builds confidence in the community. It's one of the things that separates great open source projects from good ones (the other thing is documentation).

I know I'm not perfect either. I don't write that many unit tests. I haven't created a demo for every feature and I've only written a handful of pages of documentation. But I do think we can continue to make incremental improvements in those areas and make life a little easier for everyone.

bsimser:
Maybe you don't want rules, but I think there should be (at least) one. In the absence of unit tests (unit testing MonoGame stuff is hard and we know it) maybe there should be a rule that all demos using whatever component you touched must work as-is or with the PR. Or is that too heavy handed. Cloning the PR and testing it before a merge is a good idea too.

PR #247, opened 2016-08-29
fix FramesPerSecondCounter

LithiumToast:
-Update the logic based off [Shawn's blog](https://blogs.msdn.microsoft.com/shawnhar/2007/06/08/displaying-the-framerate/) so the frames per second is really the number of frames _per second_, not the game time budget of the frame.
-Removed averaging. It's not used in the TiledMaps.Demo and is a small waste of CPU time and memory resources.
-Update `FramesPerSecondCounterComponent`.
-Tested TiledMaps.Demo. Works correctly.

EDIT: From the comments section on Shawn's blog post
> > Surely 1 / gameTime.ElapsedRealTime.TotalSeconds 
> > will therefore give the current framerate.
> 
> That will tell you how long it was since the previous call to Update, but that is not the same thing as your framerate!

> a) If the game is dropping frames, Update will be called more frequently in order to catch up. You want to time the number of actual draws that are taking place, not just these extra catch-up logic frames.
> 
> b) The time for a single Update can fluctuate widely, so the figure you get out of that will be too flickery to be easily readable.

### DIFF
--- a/Source/Demos/Demo.TiledMaps/Game1.cs
+++ b/Source/Demos/Demo.TiledMaps/Game1.cs
@@ -112,7 +112,7 @@ namespace Demo.TiledMaps
             _spriteBatch.Begin(samplerState: SamplerState.PointClamp, blendState: BlendState.AlphaBlend);
             _spriteBatch.DrawString(_bitmapFont, "WASD/Arrows: move", new Vector2(5, 32), textColor);
             _spriteBatch.DrawString(_bitmapFont, "RF: zoom", new Vector2(5, 32 + _bitmapFont.LineHeight), textColor);
-            _spriteBatch.DrawString(_bitmapFont, $"FPS: {_fpsCounter.AverageFramesPerSecond:0}", Vector2.One, Color.Black);
+            _spriteBatch.DrawString(_bitmapFont, $"FPS: {_fpsCounter.FramesPerSecond:0}", Vector2.One, Color.Black);
             _spriteBatch.DrawString(_bitmapFont, $"Camera: {_camera.Position}", new Vector2(5, 32 + _bitmapFont.LineHeight * 2), Color.Black);
             _spriteBatch.End();
 
--- a/Source/MonoGame.Extended/FramesPerSecondCounter.cs
+++ b/Source/MonoGame.Extended/FramesPerSecondCounter.cs
@@ -9,7 +9,7 @@ namespace MonoGame.Extended
         private TimeSpan _timer = OneSecondTimeSpan;
         private int _framesCounter;
 
-        public float FramesPerSecond { get; private set; }
+        public int FramesPerSecond { get; private set; }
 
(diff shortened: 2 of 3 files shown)


craftworkgames:
I like the simplicity of the new FPS counter but I think you need to take a little more care when ripping out code. Just because the demo doesn't use it doesn't mean it's not useful.

I can live without `AverageFramesPerSecond`. It does add a lot of extra complexity and memory requirements to a class that should be rather simple. Although, I expect there's probably someone out there using it. I guess we can deal with that if the issue arises.

There's probably no good reason to get rid of `TotalFrames`.

What's probably a little more concerning is that you've subtly changed the logic from Shawn's blog. His code looks like this:

	elapsedTime += gameTime.ElapsedGameTime;

	if (elapsedTime > TimeSpan.FromSeconds(1))
	{
		elapsedTime -= TimeSpan.FromSeconds(1);
		frameRate = frameCounter;
		frameCounter = 0;
	}

And your code looks like this:

	_timer += gameTime.ElapsedGameTime;
	
	if (_timer <= OneSecondTimeSpan)
		return;

	FramesPerSecond = _framesCounter;
	_framesCounter = 0;
	_timer = TimeSpan.Zero;

The change in logic is not super easy to spot, but it is in fact on this line:

    _timer = TimeSpan.Zero;

vs this line:

    _timer -= OneSecondTimeSpan;

Shawn specifically mentions this is the comments.

> Notice that I’m not resetting the elapsed time to zero, just subtracting one second from it. This will never loose time.

There's a really old article about time-steps on [Gaffer on Games](http://gafferongames.com/game-physics/fix-your-timestep/) that might be worth reading.

LithiumToast:
>but I think you need to take a little more care when ripping out code. Just because the demo doesn't use it doesn't mean it's not useful.

I understand, just that I think it's justified. With the old way, it did make sense, even necessary, because the time budget of each frame was found and then the average was taken from those values. With this way of calculating the frames per second, I don't see why the average would be of any use. In fact, I think it's rather confusing for users since they could be asking: "Should I use `CurrentFramesPerSecond` or `AverageFramesPerSecond`?". The fact that the demos don't need to use the average frames per second now was strong enough evidence for me that the averaging code with this way of calculating frames per second is redundant.

> There's probably no good reason to get rid of TotalFrames.

I don't see how that variable is useful for anything but trivia. All time sensitivity calculations use the `GameTime` class with `TimeSpan` structs, not frames.

craftworkgames:
I actually agree with you completely. You've made your argument well :wink: 

I do like classes that have a clear single responsibility. We'll leave out `AverageFramesPerSecond` and `TotalFrames`. If it bothers someone, we'll revisit it then. 

And to be honest, I never really did like the way it was before. If we do ever add an average frames per second counter it would probably be better as a separate class, maybe called `AverageFramesPerSecondCounter`. But yeah, let's leave it out for now.

PR #250, opened 2016-08-31
Added a new constructor and documentation to TextureAtlas class

Aurioch:
What the constructor does is take the `Dictionary<string, Rectangle>` of regions together with `Texture2D` and populates the `TextureAtlas.Regions` with them. Constructor is useful if the atlas regions are defined in the separate XML file which Content Pipeline deserializes into dictionary.

Also, I hope the documentation is... decent, at least, as I've done it by hand. If it's good enough, I might continue on with other classes.


Aurioch:
Tested, merged into mine, everything's fine. I need to use `var` more.

craftworkgames:
Awesome. Thanks.

PR #251, opened 2016-09-01
DynamicBatch2D - SpriteBatch API

LithiumToast:
`DynamicBatch2D`supports everything `SpriteBatch` can do except it also allows for passing a `Matrix2D` transform instead position, rotation and scale as separate values. The old API for `SpriteBatch` is marked as deprecated, but it does work allowing `DynamicBatch2D` to be a drop in replacement for `SpriteBatch`. The only major difference between `DynamicBatch2D` and `SpriteBatch` is that the number vertices, indices and batch items (or commands as I called them) are fixed upon instantiation. Of course the constructor has parameters to allow modifying these maximum numbers.

In the near future I'm going to add a `DrawShape` method or something similar to `DynamicBatch2D` to draw circles, rectangles, arcs, etc.


craftworkgames:
Thanks. I ran into a couple of minor issues with the project file.

> CSC error CS2001: Source file 'C:\TeamCity\buildAgent\work\982dbc876ae2b3b0\Source\MonoGame.Extended\Entities\Systems\DrawableComponentSystem.cs' could not be found. 
CSC error CS2001: Source file 'C:\TeamCity\buildAgent\work\982dbc876ae2b3b0\Source\MonoGame.Extended\Entities\Systems\ParticleEmitterComponentSystem.cs' could not be found. 
CSC error CS2001: Source file 'C:\TeamCity\buildAgent\work\982dbc876ae2b3b0\Source\MonoGame.Extended\Entities\Systems\SpriteBatchComponentSystem.cs' could not be found. 
CSC error CS2001: Source file 'C:\TeamCity\buildAgent\work\982dbc876ae2b3b0\Source\MonoGame.Extended\Entities\Systems\UpdatableComponentSystem.cs' could not be found. 

Easy enough to fix so I've merged it anyway.

PR #253, opened 2016-09-05
New overloads for BitmapFont DrawString

bsimser:
Added additional overloads to the BitmapFont DrawString methods to support the same functions the SpriteBatch.DrawString have. BitmapFont DrawString now supports scaling, orgins, SpriteEffects. Also updated the demo to showcase all of these.

### DIFF
--- a/Source/Demos/Demo.BitmapFonts/Game1.cs
+++ b/Source/Demos/Demo.BitmapFonts/Game1.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.Xna.Framework;
+﻿using System.Text;
+using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
 using Microsoft.Xna.Framework.Input;
 using MonoGame.Extended;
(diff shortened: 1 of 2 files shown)


craftworkgames:
Great work :smile: 

I'm really pleased that you updated the demo. A lot of people don't do that unfortunately.

There's still some room for improvement. I'd like to see rotation and scaling at a sentence level as we discussed before. Still, this is an excellent step in the right direction.

bsimser:
I agree the rotation and scaling should be done at both levels. That said I think the text would need to be rendered to a RenderTarget2D or something in order to do it at the sentence level which would require some work, not to mention some kind of additional parameter to the overloads to handle it (and an enum like SpriteEffects to indicate the type of rotation desired). A bunch of work but all doable. I'm a big believer of "version 1 sucks but ship it anyway" as there's always room for version 2. Thanks!

craftworkgames:
> That said I think the text would need to be rendered to a RenderTarget2D or something in order to do it at the sentence level

Yeah I was thinking about this approach too. My gut feeling says there's probably an easier way though.

> I'm a big believer of "version 1 sucks but ship it anyway" as there's always room for version 2.

I couldn't agree more :+1:

bsimser:
> Yeah I was thinking about this approach too. My gut feeling says there's probably an easier way though.

Probably. Might be able to create a transformation matrix and apply it to the entire thing before the draw call(s) and restore it back after. Will have to poke around in the extension method and try some stuff out when I get some time.

PR #257, opened 2016-09-12
Creation of the LayerDepthSystem

rafaelalmeidatk:
So, what it does:

- Makes it able to convert an entire Tile Layer to Entities (`TiledTileLayerToEntityService`).
- Dynamically generates depth values to each entity that has the `DepthComponent`.

## Results:
With one layer: http://i.imgur.com/gFp3nPZ.gif

With two or more layers:
With two or more layers GIF

## How it works

First of all, you register the `LayerDepthSystem` to the Entity Component System:

    _entityComponentSystem.RegisterSystem(new LayerDepthSystem());

You create the service that will convert the layers to entities:

    var tileService = new TiledTileLayerToEntityService(_entityComponentSystem, _tiledMap);

Then you convert all the layers you want:

    tileService.ConvertLayer("decorations");
    tileService.ConvertLayer("decorations2");

Now all the tiles have dinamically depth values, they are ordered by their TileD order. We can insert entities to "interect" with them just attaching the `DepthComponent` on them. The example I used is the player:

    player.AttachComponent(new DepthComponent());

If you change his depth level, him will go in front or behind the tiles.

This system is not done yet, I want to know what you all think. The plan here is let the user be able to create another `XDepthSystem`s that can manage the `DepthComponents` the same way I did.


craftworkgames:
@rafaelalmeidatk I'm pretty sure I know why you closed this because I've seen some of the discussion in gitter, but for future reference, please leave a comment about why an issue / PR is closed.

PR #261, opened 2016-09-26
Add Nez mention to README

LithiumToast:
Mention @prime31 's Nez library in README


craftworkgames:
I don't think we've ever actually used any code from Nez yet, but I'm sure we will at some point :)

LithiumToast:
@craftworkgames We used `Matrix2D` as base for ours and I have bounced ideas off @prime31 in the past before for development on MG.EX

craftworkgames:
Okay cool. Good to know.

Issue #263, opened 2016-10-03
[Tiled] offsety not supported in Tile Layers


Monix:
Just getting started with MG/MGE and was sad to stumble at my first hurdle where offsety doesn't seem to be supported. Tested with 0.5, but a quick look at the TmxTileLayer.cs in master shows no support for reading the property from the file, so I'm assuming this issue is still relevant.

Would really love to see this supported as this property is useful for vertically stacking layers in isometric maps.

Markup for the top layer:
`<layer name="1" width="20" height="20" offsetx="0" offsety="-64">`

Expected:
![tiledexpected](https://cloud.githubusercontent.com/assets/852857/19035401/ad6c8844-89ac-11e6-8c1d-dd3667368a68.PNG)

Actual:
![tiledactual](https://cloud.githubusercontent.com/assets/852857/19035404/b430f160-89ac-11e6-83e2-4404cb94135a.PNG)

PR #264, opened 2016-10-04
Support to layers offset

rafaelalmeidatk:
This fixes the #263 issue.

### DIFF
--- a/Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
+++ b/Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
@@ -79,7 +79,7 @@ namespace MonoGame.Extended.Maps.Tiled
                     continue;
 
                 var region = _map.GetTileRegion(tile.Id);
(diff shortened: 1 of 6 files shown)


Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
```diff
@@ -79,7 +79,7 @@ internal VertexPositionTexture[] BuildVertices(float depth)
                     continue;
 
                 var region = _map.GetTileRegion(tile.Id);
-                var point = GetTileLocationFunction()(tile);
+                var point = GetTileLocationFunction()(tile).ToVector2() + new Vector2(OffsetX, OffsetY);
```
rafaelalmeidatk: I converted the `GetTileLocationFunction()(tile)` to `Vector2` to keep the offset precisions, intead of casting the offsets from float to int.

craftworkgames:
Great thanks.

PR #266, opened 2016-10-10
Some Basic Primitives.

LithiumToast:
Some basic primitives. Includes full documentation along with the unit tests for, every, damn, method and operator.

Primitives: `Point2`, `Size2`, `Segment2D`, `Ray2D`, `BoundingBox2D`.

I added a `Point2` because [technically it is _not_ the same as a `Vector2`](http://math.stackexchange.com/questions/645672/what-is-the-difference-between-a-point-and-a-vector). However, I added all the operators to `Point2` which make sense, including translation by a displacement vector which I would expect would be used most commonly.

`Size2` might not be useful since we already have a `Size` and `SizeF`, but later I'm going to add a `Size3` which would have a base, width and height fields which would be used for a `BoundingBox3D` for example. I added `Size2` just to be consistent with naming.

`Segment2D` and `Ray2D` both use the helper method `RayHelper.IntersectsSlab` for finding the intersection between a `BoundingBox2D`.

Other than that, I also used NUnit`s [TestCaseSource](http://nunit.org/index.php?p=testCaseSource&r=2.5) for the unit tests and added a string name for each test explaining the assertion. These strings become visible in ReSharper's GUI extensions for Visual Studio which is nice.

screen shot 2016-10-09 at 7 03 57 pm

### DIFF
--- a/Source/MonoGame.Extended.Tests/MonoGame.Extended.Tests.csproj
+++ b/Source/MonoGame.Extended.Tests/MonoGame.Extended.Tests.csproj
@@ -69,6 +69,11 @@
     <Compile Include="Particles\ParticleBufferTests.cs" />
     <Compile Include="Particles\Profiles\PointProfileTests.cs" />
     <Compile Include="Particles\Profiles\RingProfileTests.cs" />
+    <Compile Include="Primitives\BoundingBox2DTests.cs" />
+    <Compile Include="Primitives\Point2Tests.cs" />
+    <Compile Include="Primitives\Ray2DTests.cs" />
+    <Compile Include="Primitives\Segment2DTests.cs" />
+    <Compile Include="Primitives\Size2Tests.cs" />
     <Compile Include="RangeTests.cs" />
     <Compile Include="Shapes\PolygonFTests.cs" />
     <Compile Include="Shapes\CircleTests.cs" />
--- a/Source/MonoGame.Extended.Tests/Primitives/BoundingBox2DTests.cs
+++ b/Source/MonoGame.Extended.Tests/Primitives/BoundingBox2DTests.cs
@@ -1,0 +1,327 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using Microsoft.Xna.Framework;
+using MonoGame.Extended.Primitives;
+using NUnit.Framework;
+
+namespace MonoGame.Extended.Tests.Primitives
+{
+    [TestFixture]
+    public class BoundingBox2DTests
+    {
+        public IEnumerable<TestCaseData> ConstructorTestCases
+        {
+            get
+            {
+                yield return
+                    new TestCaseData(new Point2(), new Size2()).SetName(
+                        "The empty bounding box has the expected position and direction.");
+                yield return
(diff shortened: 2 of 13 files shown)


craftworkgames:
Looks great.

So I should pull this PR and you'll do another one with the next part?  I like the idea of keeping each PR a bite sized chunk.

LithiumToast:
Yeah this PR is ready to be merged.

I would just like to add the note that all these primitives are mutable by design. This is also one of the reasons why I added `Size2` since I didn't want to have to edit and break `SizeF` so it can be mutable. It is my point of view that immutable structs are a pain to work with for primitives, especially in games where speed and performance is arguably more of a concern than business applications.

I rather _not_ have to do something like to change just one value: 

```
var newHeight = 20;
myObject.BoundingBox = new BoundingBox(myObject.BoundingBox.Centre, new Size2(myObject.BoundingBox.Radius.Width, newHeight));
```
And rather do:

```
var newHeight = 20;
myObject.BoundingBox.Height = newHeight;
```

[Microsoft even breaks their guidelines for when to use a struct vs class (see top answer).](http://stackoverflow.com/questions/521298/when-to-use-struct).

craftworkgames:
> I would just like to add the note that all these primitives are mutable by design. This is also one of the reasons why I added Size2 since I didn't want to have to edit and break SizeF so it can be mutable.

Yeah, I understand your reasoning. I've come to agree that there are cases like these when the "guidelines" are not appropriate. 

As long as we've given it some thought (which you've obviously done) I think it makes perfect sense to have these low level types as mutable structs.

Sometimes I regret adding `SizeF` to the library. Maybe it's worth having another discussion about how we evolve the primitives. I'd be okay with getting rid of `SizeF` in favor of `Size2` if it makes sense as long as we consider the impact first.

craftworkgames:
Merged. I don't have time to do any testing on it at the moment. I trust that it's all good.

LithiumToast:
I spent more time than I care to mention documenting and testing everything. You should really look it over though, I might of missed something! 😃 
If anything breaks or there is a typo in documentation, I'll be happy to fix it.

>Sometimes I regret adding SizeF to the library. Maybe it's worth having another discussion about how we evolve the primitives. I'd be okay with getting rid of SizeF in favor of Size2 if it makes sense as long as we consider the impact first.

That would be fine. Though, I do think that having `int` for primitive values in games is not very common? Is there an example where having `int` values makes more sense compared to `float` values for primitives? `float` can handle `int` values up to certain precision. I would rather scrap the whole `PrimitiveTypeF` stuff in favour for just `PrimitiveType` with `float` values; way less confusing in my opinion.

craftworkgames:
There's certainly cases when using `int` makes more sense. It's especially useful when you don't want to have to worry about precision sensitive comparisons.

Ironically, in the 90's they used to do the exact opposite. They would build everything using `int` and "simulate" floating point math using their own scale. My memory of it is rather vague, but I do remember poking around the original Doom source code and finding some amazing things.

That said, we could probably go through the library and consider all the places where we use `int` vs `float`. It could be that we can use `float` in a lot more places which would make some things easier.

Issue #188, opened 2016-05-01
Add more `DrawString` overloads for `BitmapFont`


craftworkgames:
There's a whole bunch of [extra `DrawString` overloads we could implement](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.spritebatch.drawstring.aspx) for `BitmapFont`'s. In particular, we don't currently have a way to [set the `layerDepth` at all](http://community.monogame.net/t/using-layerdepth-with-drawstring-and-bitmapfont/7583).

craftworkgames:
I added layer depth when merging #182 but the other overloads a more complicated.

bsimser:
I wonder if this should be like the Draw extensions for the other classes modeled on the DrawString methods, each one just collapsing down to a single method. I took a look at what's there now but it's just the two methods (with one internal called by the public one). Maybe just do a rewrite of this file to structure it like the DrawString methods, using the one method that exists as the one that the others call?

bsimser:
If you want you can assign this to me. Just working on it now.

Adding in support for StringBuilder text so you can do this:

`_stringBuilderText = new StringBuilder();`
`_stringBuilderText.AppendLine("Lorem Ipsum comes from sections 1.10.32 and 1.10.33");`
`_stringBuilderText.AppendLine("of \"de Finibus Bonorum et Malorum\" (The Extremes");`
`_stringBuilderText.AppendLine("of Good and Evil) by Cicero, written in 45 BC. This");`
`_stringBuilderText.AppendLine("book is a treatise on the theory of ethics, very");`
`_stringBuilderText.AppendLine("popular during the Renaissance. The first line of");`
`_stringBuilderText.AppendLine("Lorem Ipsum, \"Lorem ipsum dolor sit amet..\", comes");`
`_stringBuilderText.AppendLine("from a line in section 1.10.32.");`
`_stringBuilderPosition = new Vector2(50, 60);`

And this:

`_spriteBatch.DrawString(_bitmapFont, _stringBuilderText, _stringBuilderPosition, Color.AliceBlue);`

To get this:

![image](https://cloud.githubusercontent.com/assets/301535/18252980/a7cac596-734e-11e6-8da4-8cac5764457d.png)

bsimser:
Not sure how useful this will but here it is with rotation support (rotating each character, not the entire text). 

[...]

PR #269, opened 2016-10-11
Fixed scale for BitmapFont draws

DelvalletQuentin:
Hey, here's a little fix for issue #188.

I think the actual scale, doing at an individual letter level, was not really useful, it's much more convenient to have a scale for the full sentence. This way we would require less bitmap fonts to write texts of multiple size in projects.

Here are some examples using the demo :

untitled
untitled2
untitled3

I think it should be a great temporary solution before finding one that would also correct the rotation issue, which I haven't investigated yet.

I've directly modified the existing draw, I don't really know if we need both solution for scaling.


craftworkgames:
Ah yeah, sorry I didn't get back to this until now. Kinda forgot it was here.

It's an improvement over what we had previously, but it would be really nice to get rotation working properly (i.e. rotating the whole string rather than each individual letter).

PR #271, opened 2016-10-18
Removed the X and Y properties keeping Position

musbah:
I removed the X and Y properties keeping Position property to make things less redundant.
I also added another constructor that takes a Vector2 position directly.

edit: I wanted to fix the spacing a bit in the 2nd commit, but I totally forgot to write a good message...

### DIFF
--- a/Source/MonoGame.Extended/Maps/Tiled/TiledObject.cs
+++ b/Source/MonoGame.Extended/Maps/Tiled/TiledObject.cs
@@ -6,12 +6,12 @@ namespace MonoGame.Extended.Maps.Tiled
 {
     public class TiledObject
     {
-        public TiledObject(TiledObjectType objectType,int id,int? gid,float x,float y,float width,float height) : this(objectType,id,gid,width,height)
+        public TiledObject(TiledObjectType objectType, int id, int? gid, float x, float y, float width, float height) : this(objectType,id,gid,width,height)
         {
             Position = new Vector2(x,y);
         }
         
-        public TiledObject(TiledObjectType objectType,int id,int? gid,Vector2 position,float width,float height) : this(objectType,id,gid,width,height)
+        public TiledObject(TiledObjectType objectType, int id, int? gid, Vector2 position, float width, float height) : this(objectType,id,gid,width,height)
         {
             Position = position;
         }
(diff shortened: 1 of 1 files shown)


Source/MonoGame.Extended/Maps/Tiled/TiledObject.cs
```diff
@@ -1,18 +1,26 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using Microsoft.Xna.Framework;
 using MonoGame.Extended.Shapes;
 
 namespace MonoGame.Extended.Maps.Tiled
 {
     public class TiledObject
     {
```
craftworkgames: Rather than having 2 public constructors and a private one you can simply have 2 public constructors and call one from the other using `this()`

musbah: So something like this?

		public TiledObject(TiledObjectType objectType,int id,int? gid,float x,float y,float width,float height) : this(objectType,id,gid,new Vector2(x,y),width,height)
		{

		}

		public TiledObject(TiledObjectType objectType,int id,int? gid,Vector2 position,float width,float height)
		{
			ObjectType = objectType;
			Id = id;
			Gid = gid;
			Width = width;
			Height = height;
			Points = new List<Vector2>();
			Properties = new TiledProperties();
			Position = position;
		}

I did not think of creating the `Vector2` in `this()` , sorry about that.

Source/MonoGame.Extended/Maps/Tiled/TiledObject.cs
```diff
@@ -23,11 +31,9 @@ public TiledObject(TiledObjectType objectType, int id, int? gid, float x, float
         public int? Gid { get; }
         public TiledObjectType ObjectType { get; }
         public string Name { get; set; }
-        public float X { get; }
-        public float Y { get; }
         public float Width { get; }
         public float Height { get; }
-        public Vector2 Position => new Vector2(X, Y);
+        public Vector2 Position { get; private set; }
```
musbah: I'm a bit new when it comes to C# properties and what not, so I though that you needed private set at all times to change the value. Being able to change it in the constructor does make sense now that I think about it though...

musbah:
I think I'm going to close this PR and create a new one with all the changes, don't need many unnecessary commits.

craftworkgames:
> don't need many unnecessary commits.

If the only reason is too many commits, don't worry about it.

 1. Github squishes PR's down into one commit by default these days.
 2. I really don't care either way.

The main thing I'm looking for in PR's is something of reasonable size, so that I can quickly see what's coming in. Really large PR's are a super pain in the ass.

I really appreciate what you've done. I hope to see more in the future. Thanks 😄

PR #272, opened 2016-10-18
Removed the X and Y properties keeping Position

musbah:
I removed the X and Y properties keeping Position property to make things less redundant.
I also added another constructor that takes a Vector2 position directly.

Reposted again because something weird was going on with the fork, so recreated a new one. Modified the code according to @craftworkgames's recommendations.


craftworkgames:
Great work 😄

PR #274, opened 2016-10-20
Solution reorganized: All namespaces splitted into multiple small shared projects (.shproj)

raizam:
The original need was to pick a set of small standalone classes, without loading the whole library. 
It would be desirable to avoid having a single big monolitic assembly.
The goal is to break the library into small pieces, and aims for a more decoupled feel.

This new solution layout also helps figure out how each area of Monogame.Extended are dependent to each other.

**These changes will have absolutely no side effects on anyone**


bsimser:
Not sure if @craftworkgames will have something to say about this (or if this was from his request) but this seems pretty excessive and a major change. Splitting the system up into projects means you now need a new nuget package for each project (that seems like a PITA) and even if you didn't, why would you need this? Separate projects does not decouple the system, just makes for more maintenance when assembling a release. Also libraries don't get loaded, you use what you need and the way everything is split up via namespaces makes sense. If you don't need a subsystem then don't use it. The fact it exists in the binary doesn't matter and doesn't affect your memory footprint of your app. The only thing it would affect is the shipping size of the binaries, but frankly the size of MGE is pretty much negligible at this point. In any case, just wanted to put my 2c here but it's ultimately Dylan's call. I don't think separate projects is a good idea or necessary, but that might just be me.

raizam:
It's really up to you guys, if you like it you can merge, otherwise you don't. 
This is my setup, and It's practical to me =)

LithiumToast:
This comment probably belongs in an issue...

The content pipeline projects and test projects should probably be split up as well if this is going to happen. 

However, I think some more thought should be considered as to how to slice the project up before we proceed... I think splitting up into separate libraries based on namespaces is a good starting point, but perhaps things should be considered a little more delicately in some situations. 

If someone needs to reduce the shipping size of their game, they are free to hack up the library to his or her liking since it is all open source code. So, I don't think we should really worry about size of the code in each namespace that much. I think it makes more sense to focus on modularity.

However, one of the main problem seems to be the dependencies... For example the texture atlases feature is used in many other features. To me, this says that texture atlases is more or less common code and should be considered apart of a "core" library. Do we want  a "core" library? Or is it acceptable to have multiple common feature libraries such as `TextureAtlases` or even `Primitives` (when it gets flushed out more) which most people would most likely have to add as dependencies separately?

Also while we are on the subject of moving around code, perhaps we could add the namespace for 2D and 3D?

raizam:
> I think it makes more sense to focus on modularity.

While this would imply major breaking changes, this would be benificial in the long term. A loosely couple architecture would be very flexible.

> However, one of the main problem seems to be the dependencies...

What would work in many cases would be add some well designed interfaces in the core library, that would decouple things. But this would imply more work to the user for instanciation. A IoC/dependency injection solution would then be needed.

> To me, this says that texture atlases is more or less common code and should be considered apart of a "core" library.

I noticed the current TextureAtlas design, based on TextureRegion2D considers a texture element as inner Rectangles, and this is not exactly my need. What I currently have in mind, rather than a Rectangle, would be a polygon with N sides describing the precise outline of the texture element. The reason I need it is to have detailed 2D meshes that are polygon as well, and this would allow precise collisions detection. (planning to use farseer).
So in my case, I'm not wishing TextureAtlases as it is to be desirable to be in the Core part.
Now if a more flexible approach is possible, for example by having abstracting it, and having a ITextureAtlas interface work, then it could work for me.

There is indeed a lot to think about.

raizam:
>  I think it makes more sense to focus on modularity.

This is precisely the direction I'm hoping for, while this will introduce major breaking changes, if this is well done it would be worth it.

raizam:
I'm working on new classes using these shared projects and I'm realizing it's really not practical to work with: it breaks a lot of Visual Studio code navigation features, such as namespace resolutions, it breaks intellisence also. So I'm closing this pull request, as this is not a good working condition.

Such approach may still be used for in deployment context.
Also I haven't made these projects by hand, I wrote a handy command line tool that generates them such as:
`make_shproj DefaultNamespace /includeFolder1 /includeFolder2/*/Subfolder`

I can share it if needed.
Hopefully this was still the occasion to consider a more modular path.

craftworkgames:
Hey guys, it's about time I piped in here and said something. I actually started writing a reply on Friday but I got busy / distracted and never finished it.

It turns out most of what I wanted to say has already been said, but let me just be clear on my position with this.

First, let's not get confused about what we are talking about. If I understand correctly, @raizam tried splitting the library into multiple [*Shared Projects*](https://developer.xamarin.com/guides/cross-platform/application_fundamentals/shared_projects/) (note: I linked to the Xamarin documentation because I think it explains them better than [Microsoft's documentation](https://blogs.msdn.microsoft.com/dotnet/2014/04/21/sharing-code-across-platforms/)).

[Shared Projects (SP) are closely related to Portable Class Libraries (PCL)](https://developer.xamarin.com/guides/cross-platform/application_fundamentals/building_cross_platform_applications/sharing_code_options/) in the sense that they are one of the ways to share code between projects targeting different platforms. 

MonoGame.Extended is currently a PCL. That's how it's able to be referenced from many different platforms without the need to recompile. There's pro's and con's to both SP's and PCL's. We are currently living with the PCL approach, and in my opinion it works pretty well for our needs.

To be clear, I haven't seen a compelling reason to introduce SP's into the mix at this point. While they do bring some nice things to the table I don't think there's enough value for the needs of this particular project. I wouldn't have merged this PR anyway.

Now, that said, I think there's absolutely a separate discussion to be had about splitting the library up into a couple of different packages. Dependencies are certainly a good reason to split, although, the details need to be fleshed out a lot more first.

> - Splitting the system up into projects means you now need a new nuget package for each project (that seems like a PITA) and even if you didn't, why would you need this?
> - Do we want a "core" library? 
> - Or is it acceptable to have multiple common feature libraries such as TextureAtlases or even Primitives (when it gets flushed out more) which most people would most likely have to add as dependencies separately?
> - Also while we are on the subject of moving around code, perhaps we could add the namespace for 2D and 3D?

These are all indeed good questions but let's not discuss them here. Instead, let's use a new issue #275

@raizam I believe [TexturePacker](https://www.codeandweb.com/texturepacker) supports polygon based texture regions. Although, I doubt they would be very useful for collision detection polygons as well. In my experience, you typically want to keep rendering polygons and collision polygons separated for more control. I mean, in some cases it might make sense to make them the same, but I wouldn't want to be forced into that situation.

We could of course support polygon based texture regions. The way I would do it though, is to keep the existing TextureRegion class with a bounding rectangle and add optional polygon data to it. Keep in mind, this is really an optimization. It reduces the [texture memory footprint and overdraw at a cost to CPU](https://www.codeandweb.com/texturepacker/tutorials/cocos2d-x-performance-optimization). It makes perfect sense for certain shaped sprites but in a lot of cases rectangles is still the right way to go.

PR #278, opened 2016-10-26
Add support for Multiple Tilesets per TiledTileLayer in Tiled Map Renderer.

OptimusPi:
Tiled Map rendering now supports multiple tilesets. Previously there was a bug where tiles made with alternate tilesets would draw undefined images.

TiledTileLayer holds an array of tiles to draw. This was based on their order from left to right/top to bottom in the grid. The new ordering is based on each TiledTileset to allow tiles with the same TiledTileset to be placed next to each other in memory.

This is to support the best choice for map rendering optimization available at this time. The old code would draw the map with 1 call per TiledTileLayer. This new code has 1 draw code per Tileset, per TiledTileLayer. (The re-ordering of tiles described in the above paragraph is necessary to prevent 1 draw call per TiledTile per TiledTileset.)

Note: This is my first ever PR on GitHub so please don't be gentle with me; let me know if and what I did wrong and how to improve. Thanks :)


craftworkgames:
Hey mate

The changes look great 😄 

The one thing we seem to be missing though is an updated demo or some kind of test to prove that it works. Would you happen to have an example TMX file that uses multiple tilesets we can include in the demo?

As a general rule, when I'm merging a PR it helps greatly if I have some way to test the changes. Small changes are usually okay since I can tell what it's going to do by looking at the code. Anything bigger than a few lines though and it gets tricky.

Other than that your PR is perfect 👍

OptimusPi:
Good idea. I can definitely do that, would it be appropriate to update Demos.Platformer to show this?

craftworkgames:
Sure thing

OptimusPi:
Before code change:
https://i.imgur.com/UCe6KJl.png

After code change:
https://i.imgur.com/jjyz5aj.png

I have commited my change to my branch too, just now to demonstrate I have added to Demos.Platformer:

-Added tiny-grey-cloud.png, Updated level-1.tmx to include 2 tilesets (tiny-tiles and tiny-grey-cloud) and used both tilesets in creating the map. And of course had to update Content.mgcb which also changed Demo.Platformer.csproj.

Thanks!

craftworkgames:
Great work 👍

PR #279, opened 2016-10-26
BoundingRectangle

LithiumToast:
- Renamed `BoundingBox2D` to `BoundingRectangle`.
- Added closest point method for `BoundingRectangle`.
- Added tests relevant tests.


craftworkgames:
Cool.

Do you think it could potentially be confusing to have a `BoundingRectangle` and a `RectangleF` in the library?

LithiumToast:
I don't think so since they have different definitions. It's easy enough to add an implicit operator to convert between anyways. I already did just that for `Rectangle`, but not `RectangleF` yet.

craftworkgames:
Okay.

PR #281, opened 2016-10-27
BitmapFonts now returning a "rectangle" size.

lozzajp:
I have done what tests I can and it appears to work. May want to hold off on the PR until a few more people can confirm it works as expected for them.

Edge cases checked a bit too, ie one line strings in this version match one line strings from before so hopefully nothing would break anyone else using this to measure single lines.

### DIFF
--- a/Source/MonoGame.Extended/BitmapFonts/BitmapFont.cs
+++ b/Source/MonoGame.Extended/BitmapFonts/BitmapFont.cs
@@ -49,17 +49,20 @@ namespace MonoGame.Extended.BitmapFonts
             var totalHeight = 0;
             var height = 0;
 
-            var newlineCodePoint = 13;
+            const int newlineCodePoint = 10;
             var codePoints = GetUnicodeCodePoints(text).ToArray();
 
             for (int i = 0, l = codePoints.Length; i < l; i++)
             {
                 BitmapFontRegion fontRegion;
                 var c = codePoints[i];
+                var d = c;
+
+                if (i < l - 1) d = codePoints[i + 1];
 
                 if (_characterMap.TryGetValue(c, out fontRegion))
                 {
-                    if (i != text.Length - 1)
+                    if (i != text.Length - 1 && _characterMap.ContainsKey(d))
                         width += fontRegion.XAdvance + LetterSpacing;
                     else
                         width += fontRegion.XOffset + fontRegion.Width;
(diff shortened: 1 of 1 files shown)


Source/MonoGame.Extended/BitmapFonts/BitmapFont.cs
```diff
@@ -44,28 +44,48 @@ internal static IEnumerable<int> GetUnicodeCodePoints(string s)
 
         public Size GetSize(string text)
         {
+            var totalWidth = 0;
             var width = 0;
+            var totalHeight = 0;
             var height = 0;
+
+            const int newlineCodePoint = 10;
```
craftworkgames: Is there any reason why this line couldn't be `const int newlineCodePoint = '\n';`?

Source/MonoGame.Extended/BitmapFonts/BitmapFont.cs
```diff
@@ -44,28 +44,48 @@ internal static IEnumerable<int> GetUnicodeCodePoints(string s)
...
             var codePoints = GetUnicodeCodePoints(text).ToArray();
 
             for (int i = 0, l = codePoints.Length; i < l; i++)
             {
                 BitmapFontRegion fontRegion;
                 var c = codePoints[i];
+                var d = c;
```
craftworkgames: We might be getting a few too many single letter variable names making the core more difficult to understand. I think now is the time to rename them to something more descriptive.

craftworkgames:
Looking good so far.

Ideally, we'd have unit tests for this stuff, but I understand it can sometimes be difficult to write tests for code that depend on textures and things like that.

Source/MonoGame.Extended/BitmapFonts/BitmapFont.cs
lozzajp: Will do, was just trying to follow the "convention" in place thus far. Do you have any problem with me commenting as well?

craftworkgames: In general we should be trying to write [self documenting code](https://blog.codinghorror.com/coding-without-comments/) by naming methods and variable names clearly.

A few comments are helpful if they are truly adding *value* to the code. Just don't add comments to code where it's already obvious what's going on.

For example, adding a comment like this is completely useless and just adds clutter to the code. Please don't do this. If you feel a strong need to write comments, try renaming some variables first.

    // gets the unicode code points
    var codePoints = GetUnicodeCodePoints(text).ToArray();

craftworkgames:
> My tests seem ok so far, pending what @westbrma thinks of it, this may be good to go.

I'll wait a little longer for @westbrma to take a look, but otherwise I'm happy to merge it.

craftworkgames:
Great work 👍

PR #283, opened 2016-10-28
Changed obsolete demo code

musbah:
While browsing the demos I noticed that some code was following the obsolete method of doing things, so I modified them. It all seems to be straightforward so there shouldn't be any problems but you never know.

Also sorry about the useless Unit Test commits that don't do anything... I was testing something out and I totally forgot to remove them...

edit: By the way, it doesn't really have anything to do with this (well except for my mistakes...) but the `AngleTest.cs` unit tests work fine on my machine.


craftworkgames:
Great. Thanks :)

PR #285, opened 2016-10-30
Fixed missing tiles when rendering large Tiled maps

### LINKED ISSUE #284: [Tiled] Maps with many tiles render incorrectly
I was playing around with a few of the demos today (from the develop branch) and noticed that if I used "untitled.tmx" with the TiledMaps demo it would only display a strip of tiles towards the top of the map even though the entire map should have shown the grass tile.  Another interesting thing to note is that these tiles only came from the "Grass" layer and neither of the other two layers were displayed.

For reference here is what the map looks like in the Tiled editor:
tilededitor

And here it is at runtime:
runtime
Note that the tiles stop rendering part way through the last row.  The coordinates of the first invisible tile are (384, 32) which is tile number 15,884 rendered RightDown (not including any other layers).

I then erased all the grass tiles outside of the "main" map area but left the map size 500x500 and ran the demo again:
runtimeedited

As you can see it rendered correctly this time so it seems to be tied to the number of non-blank tiles and not the size of the map itself.  Please let me know if any other information would be helpful.

HyperionMT:
This fixes #284 by setting baseVertex properly in call to DrawIndexedPrimitives when vertex buffer is larger than ushort.MaxValue

### DIFF
--- a/Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
+++ b/Source/MonoGame.Extended/Maps/Tiled/TiledMap.cs
@@ -185,22 +185,17 @@ namespace MonoGame.Extended.Maps.Tiled
 
                 foreach (var layer in _layers)
                 {
-
-
-                    var tiledTileLayer = layer as TiledTileLayer;
-
-                    if (tiledTileLayer != null)
+                    if (layer is TiledTileLayer)
                     {
-                        var tileLayer = tiledTileLayer;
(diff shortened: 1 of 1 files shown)


musbah:
I tested it on a few different tiled maps, looks good , seems to have fixed the problem.  Also I love the fact that you removed `tiledTileLayer` and just used `tileLayer ` , the way it was written before was weird...

craftworkgames:
All I can say is, awesome work 👍 😄

PR #286, opened 2016-10-30
BoundingRectangle Radii

LithiumToast:
- Rename `Radius` to `Radii`.
- Removed old artifact code.

`BoundingRectangle` is 100% done now, imo.


craftworkgames:
> Rename Radius to Radii

Why?

LithiumToast:
Because there are two radiuses, or radii. One for the x-axis, and one for the y-axis.

craftworkgames:
Oh, you made it a `Vector2`. I didn't see that.

So what we're actually talking about here is the bounding ellipse of the rectangle? 

Wouldn't it perhaps be useful to keep the `Radius` property to represent the bounding circle as well?  

I guess you don't need it in your code but I'm surprised that using the bounding ellipse is the better way to go. I would have thought bounding circle calculations would be less CPU intensive.

LithiumToast:
It's not a bounding ellipse of the rectangle. `BoundingRectangle` is represented as the bounding rectangle of an ellipse. It's defined by a centre and half size for an AABB. This representation is useful for testing it against a circle.

> Wouldn't it perhaps be useful to keep the Radius property to represent the bounding circle as well?

Maybe. I'm not sure which would be more confusing. I think radii is better cause it clearly shows the intent.

> Oh, you made it a Vector2. I didn't see that.

The problem with `Size2` for a radius is that the names width and height can be confusing. It's essential the half width and half height that is the radii.

craftworkgames:
Yeah, that's cool. I understand now.

I wasn't suggesting that `Size2` is a better option, I was just confused. I made the assumption that it was previously a `float` (as you'd expect to find on a `Circle`).

Mostly I'm just trying to understand the changes with very little context.

LithiumToast:
> I wasn't suggesting that Size2 is a better option, I was just confused. I made the assumption that it was previously a float (as you'd expect to find on a Circle).

That's exactly why I think `Radii` is better because it shows the intent that is not just a float, but rather two floats.

craftworkgames:
@LithiumToast I merged this and noticed some of the unit tests are failing. Can you please look into it?

Other than that, great work 😄

PR #287, opened 2016-10-30
Precalculate tilecount per tileset when TiledTileLayer is created

HyperionMT:
The calculation of how many tiles were being drawn was being done (EffectPassesCount \* LayerCount \* TilesetCount) times each time the map was being drawn.  For small maps this wasn't a big deal but for larger maps such as the 500x500 untitled.tmx map storing the counts at the TiledTileLayer level took my FPS from ~37 to ~90 FPS.  Also tested against level01/02/03 and Platformer demo but those maps were too small to see any improvement.

### DIFF
--- a/Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
+++ b/Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
@@ -35,6 +35,7 @@ namespace MonoGame.Extended.Maps.Tiled
         private RenderTarget2D _renderTarget;
         private readonly List<TiledTile> _animatedTiles;
         private List<TiledTileSetTile> _uniqueTileSetTiles = new List<TiledTileSetTile>();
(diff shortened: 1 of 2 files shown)


Source/MonoGame.Extended/Maps/Tiled/TiledTileLayer.cs
```diff
@@ -60,6 +61,14 @@ private TiledTile[] CreateTiles(int[] data, IReadOnlyList<TiledTileset> tilesets
                         if (tilesets[tilesetIndex].FirstId <= data[tileDataIndex])
                             break;
 
+                    TiledTileset ts = tilesets[tilesetIndex];
+                    if (!this.tileCountByTileset.ContainsKey(ts))
```
LithiumToast: I think `Dictionary<TKey, TValue>.TryGetValue()` should be used here.

craftworkgames:
Great work.

There's still quite a few things I'd like to do on the new Tiled renderer before the next release but this is a great step in the right direction.

Thanks.

PR #289, opened 2016-11-03
Changing TileSet instances to Tileset 

musbah:
I had some time, so I though I would dive into that. I'm pretty sure I changed the names everywhere, unless I missed one. I made sure to test the demos and they all seem to work.

edit: also sorry about the first 3 commits, I tried everything I could to get rid of them but it wasn't working... I'm not sure why github is comparing the current code to old code since the changes have already been done in the repo... Sorry, not sure what's going on...

If needed, I can always recreate a new fork and only add the new changes. Sorry I just can't seem to figure out why it's keeping them even though I'm merging and everything...


OptimusPi:
Bravo!!!

craftworkgames:
Awesome.

PR #290, opened 2016-11-04
Only draw tiles that actually exist for each tileset/layer combination

HyperionMT:
One more performance optimization for Tiled map rendering (I don't think there's anymore low-hanging fruit in terms of performance).  I noticed that for any map (lets say 100x100 tiles for example) we were drawing 10,000 tiles for each tileset/layer combination even if the tile was blank.  Considering most layers beyond the "base" layer usually have more blank tiles than visible tiles, this is a huge waste.

By only drawing tiles for those that existed I was able to improve the FPS on my machine for untitled.tmx (500x500) from ~90 to ~1,100 FPS.  The vertices/indexes for DrawIndexedPrimatives are saved in a new class TiledRenderDetails on each layer at layer creation and fetched during TiledMap.Draw.

I've tested these changes on the Platformer and TiledTileMap demos.


musbah:
You've done so many amazing optimizations, thank for all of that.

Issue #144, opened 2016-03-18
[Tiled] Transparency does not work


alexeygritsenko:
Please Fix the opacity property
![opacity](https://cloud.githubusercontent.com/assets/10233486/13871788/2be977e4-ecf6-11e5-87af-0c494bbd0209.png)

craftworkgames:
What version are you using?  I could be wrong but I have a feeling this is already fixed in the `develop` branch.

alexeygritsenko:
lastet 0.4.64

craftworkgames:
@alexeygritsenko Did you test the code in `develop`? If you don't want to download and build the code yourself you can always use the NuGet package from our build server.

    http://build.craftworkgames.com/guestAuth/app/nuget/v1/FeedService.svc/

alexeygritsenko:
Yes thank you. I just wanted to ask where I can download the latest beta release.
However, your server address does not work. Why is there an older version of the exact same
![betaold](https://cloud.githubusercontent.com/assets/10233486/14010176/df97e990-f1a3-11e5-990e-31eb9f4cfb33.PNG)

craftworkgames:
You're almost there. You just need to include pre-release packages.

![image](https://cloud.githubusercontent.com/assets/3201643/14010263/548d2aae-f1df-11e5-91bd-bd11b7477044.png)

alexeygritsenko:
ok thanks, now I installed the latest beta of the example.
but a new problem now Tiled Maps
![betaerr](https://cloud.githubusercontent.com/assets/10233486/14010591/9f450900-f1a7-11e5-9b02-37b45986aa90.PNG)
 example does not start
Reading after the end of the stream is not possible.

craftworkgames:
We've upgraded to use MonoGame 3.5.

alexeygritsenko:
da ya tozhe
Yes, I, too,

alexeygritsenko:
![vers](https://cloud.githubusercontent.com/assets/10233486/14010683/845e7bb6-f1a8-11e5-99b7-31e974f3f5ab.PNG)

craftworkgames:
Ah well in that case I'm not sure. Try cleaning and rebuilding your content. Make sure it's the 3.5 version of the Pipeline tool.

alexeygritsenko:
i update pipeline, but new error
[...]

Issue #268, opened 2016-10-11
[Tiled] Split the Tiled renderer out of the map code


craftworkgames:
So a while back when we did some work on #230 it was a little unclear how we should handle rendering the Tiled maps to allow end users to configure and customize rendering for different situations.

Then recently I was reading about [Retained Mode](https://msdn.microsoft.com/en-us/library/windows/desktop/ff684178%28v=vs.85%29.aspx) rendering and I realized that there's many benefits to splitting the _model_ from the _rendering_. I think that's what we should do for the Tiled renderer (and perhaps other parts of the framework).

This would allow the responsibility of rendering the map be separated from the map data. The beauty of this approach is that different types of renderers could be implemented independently of a `TiledMap` allowing for different renderers to be optimized for different situations.

HyperionMT:
Is anybody currently working on this?  If not I'd like to take a shot.

What I'd like to do is create a FullMapRenderer with the existing code and create an IMapRenderer interface future renderers can implement.  I think IMapRenderer would look something like:

```
IMapRenderer
    void SwapMap(TiledMap newMap);

    [Obsolete]
    void Draw(Camera2D camera);
    void Draw(Matrix viewMatrix);
```


I think once that's done we might be able to extract an interface on the Tiled objects which the renderers can reference so they can support any future map formats as well.  Any comments/suggestions?

craftworkgames:
> Is anybody currently working on this? If not I'd like to take a shot.

Nobody is currently working on it.

You've been doing some great work already, so yeah, you're welcome to take this on.

To be honest, I'm not sure what the intention of `SwapMap` would be, but I'll leave it in your capable hands and see what you come up with.

Thanks.

HyperionMT:
I was thinking for projects with multiple maps (levels) you wouldn't want to create a new renderer just to display the next level, instead you would swap out the current map with a new one.

craftworkgames:
Yeah, that makes sense.

It would be nice to update the demo to swap between each of the different maps that way.

craftworkgames:
Completed in #292

Issue #235, opened 2016-08-17
[Tiled] Layers rendering incorrectly when not square


TheUltimateHose:
I was recently messing around with Tiled map rendering and came across the issue that when I tried to render a layer that was 16x19 tiles large, only the first 16x16 tiles would actually be rendered. The issue was solved by changing the map size to 19x19, but that is not really a great solution.

The code I used to render the layer looks like this:

```
_spritebatch.Start(transformMatrix: camera.GetViewMatrix(), samplerState: SamplerState.PointClamp);
_layer.Draw(_spritebatch);
_spritebatch.End();
```

And the followong code produces the same result:

```
_spritebatch.Start(transformMatrix: camera.GetViewMatrix(), samplerState: SamplerState.PointClamp);
_spritebatch.Draw(_layer);
_spritebatch.End();
```

As mentioned above, the three rows on the bottom would not render, but I didn't test that with different render orders (mine was right-down).

**EDIT: I just found that this bug was already reported (issue was closed)**

craftworkgames:
We've been working on the Tiled renderer a lot lately. I'm not sure if the bug exists in the new renderer or not but just in case I'll leave this issue open.

Would you be able to provide a TMX file to reproduce the issue?

TheUltimateHose:
[Here](https://gist.github.com/TheUltimateHose/b86d8d1f8841d5c58a954c1e15dd6c24)'s a gist with the TMX file, the texture can be found [here](http://kenney.nl/assets/roguelike-modern-city).

TheUltimateHose:
I just downloaded and updated to the latest available package from your build server (0.5.415-beta) and tested it myself and the above code worked just fine, rendering the entire map even if it's not square.

HyperionMT:
Can this be closed?  I was going to fix it as part of #268 but couldn't reproduce it.  I've also added a new level04.tmx file (non-square map) to the TiledMaps demo as part of that issue so we can make sure it doesn't break again.

craftworkgames:
> Can this be closed? I was going to fix it as part of #268 but couldn't reproduce it. I've also added a new level04.tmx file (non-square map) to the TiledMaps demo as part of that issue so we can make sure it doesn't break again.

Yeah, let's close it. If you can't reproduce it's most likely fixed. Besides, it's always difficult to fix things if you can't prove they are broken ;)

Issue #222, opened 2016-07-04
[Tiled] Layer type unsupported


weaversam8:
In attempting to render a Tiled map on the develop branch, I'm encountering this error:

```
System.NotSupportedException was unhandled
  HResult=-2146233067
  Message=Layer type  is not supported
  Source=MonoGame.Extended
  StackTrace:
       at MonoGame.Extended.Maps.Tiled.TiledMapReader.ReadLayerTypeProperties(ContentReader reader, TiledMap tiledMap, String layerName, String layerType)
       at MonoGame.Extended.Maps.Tiled.TiledMapReader.ReadLayer(ContentReader reader, TiledMap tiledMap)
       at MonoGame.Extended.Maps.Tiled.TiledMapReader.Read(ContentReader reader, TiledMap existingInstance)
       at Microsoft.Xna.Framework.Content.ContentTypeReader`1.Read(ContentReader input, Object existingInstance)
       at Microsoft.Xna.Framework.Content.ContentReader.InnerReadObject[T](T existingInstance)
       at Microsoft.Xna.Framework.Content.ContentReader.ReadObject[T]()
       at Microsoft.Xna.Framework.Content.ContentReader.ReadAsset[T]()
       at Microsoft.Xna.Framework.Content.ContentManager.ReadAsset[T](String assetName, Action`1 recordDisposableObject)
       at Microsoft.Xna.Framework.Content.ContentManager.Load[T](String assetName)
       at DungeonGame.Game.LoadContent() in C:\Users\Samuel\Documents\Xamarin\Projects\Dungeon\SharedProject1\Game.cs:line 65
       at Microsoft.Xna.Framework.Game.Initialize()
       at DungeonGame.Game.Initialize() in C:\Users\Samuel\Documents\Xamarin\Projects\Dungeon\SharedProject1\Game.cs:line 51
       at Microsoft.Xna.Framework.Game.DoInitialize()
       at Microsoft.Xna.Framework.Game.Run(GameRunBehavior runBehavior)
       at Microsoft.Xna.Framework.Game.Run()
       at DungeonGame.Program.Main() in C:\Users\Samuel\Documents\Xamarin\Projects\Dungeon\DungeonWindows\Program.cs:line 18
       at System.AppDomain._nExecuteAssembly(RuntimeAssembly assembly, String[] args)
       at System.AppDomain.ExecuteAssembly(String assemblyFile, Evidence assemblySecurity, String[] args)
[...]

PR #292, opened 2016-11-05
Separate out the Tiled rendering code

HyperionMT:
~~This isn't ready to merge yet~~ but since this is a large PR I wanted to get it out there and start getting feedback.

Highlights:

- New FullMapRenderer which consists of all the rendering related code that used to be in the Tiled* classes.
- Also ended up fixing #144 which (although not broken at the time that issue was created) was broken as part of the last Tiled Map rendering rewrite.
- TiledMaps demo
  - New level04.tmx file to test transparency and non-square maps
  - New level05.tmx file to test isometric left-up rendering
  - Pressing Tab will cycle through available maps (5th one takes a while to load because its large)
- Decided not to extract an interface from the Tiled* classes.  As I was thinking about the changes I wanted to make I realized I was turning the Tiled* classes into a sort of generic tilemap storage system which I wasn't sure was the right way to go.  I still think it would be nice for these renderers to work for any map but since we don't support any others I won't worry about it right now

ToDo:

- ~~Unit Tests, even though this code wasn't tested before I'd like to work out some way to test it~~  Wasn't able to test it all but got a few decent tests in
- ~~Tile Animations are broken again (looks like they broke sometime during the last rewrite).  If its an easy fix I'll do it as part of this PR, still need to investigate.~~  Too much for this PR.  Underlying issue is that texture vertices are built once and not updated even when the animation changes its frame.  I think we'd need a way to update the buffer without rebuilding (time consuming on larger maps) to get the animations working.
- ~~ChunkedMapRenderer, I also want to implement a renderer that will only render the chunk surrounding the camera but this might be better to handle in a separate PR.~~  Going to handle separately

Requests:

- ~~I'm not convinced the code to group tiles by their tilesets is needed/correct.  I'd appreciate it if @OptimusPi (I modified it, but I think it was originally your code) or someone else more knowledgeable could take a look at [my comment](https://github.com/HyperionMT/MonoGame.Extended/blob/98a4880f5660825fcfc4e0440c9bbd00043708a0/Source/MonoGame.Extended/Maps/Renderers/FullMapRenderer.cs#L191) and give me their opinion.  Ignoring the fact that it might mess up rendering of isometric maps, wouldn't it be useless since we're only rendering 1 tileset (texture) at a time?~~  Underlying problem is rendering all tiles for one tileset, not the reordering of tiles


craftworkgames:
> Unit Tests, even though this code wasn't tested before I'd like to work out some way to test it

Yeah, adding more unit tests is a great idea (although, not all unit tests add value, but that's a topic for another time).

Something we also need in terms of testing is a few different maps that that cover all of the different features we support. A demo that can easily switch between the maps would also be great.

> Ignoring the fact that it might mess up rendering of isometric maps, wouldn't it be useless since we're only rendering 1 tileset (texture) at a time?

For example, if we had demos / tests to prove that isometric maps and multiple tilesets are working, we could just run those after the refactor to make sure we haven't broken anything.

Fantastic work though @HyperionMT

HyperionMT:
So I think I've addressed everything I wanted to at this point and feel its ready to be merged when you're comfortable.

By the way, I'm not sure what's up with that failed test as I don't think I touched anything related to it.

> Something we also need in terms of testing is a few different maps that that cover all of the different features we support. A demo that can easily switch between the maps would also be great.

I updated the TiledMaps demo so "Tab" switches through all available maps.

> For example, if we had demos / tests to prove that isometric maps and multiple tilesets are working, we could just run those after the refactor to make sure we haven't broken anything.

How about a demo that proves its broken? ;)
isometricproblem

Basically what we have here is a new map level05.tmx that is setup as isometric and rendered left-right, bottom-top.  Due to this, the "top" tile should hide part of the tile "beneath" it.  There are 3 tilesets with 1 tile each.  Each column should display similar to the first however as you can see they don't.  The problem is not so much reordering of tiles as it is the fact that we draw all tiles for a single tileset at a time.  For orthogonal and isometric right-down this isn't an issue.  Anyway, since this issue would have existed before this PR I think its better to handle it separately.

craftworkgames:
> By the way, I'm not sure what's up with that failed test as I don't think I touched anything related to it.

Yeah, that test was broken the other day. I believe I fixed it in the `develop` branch. 

Did you merge the latest `develop` code into your local code before committing?

HyperionMT:
> Did you merge the latest develop code into your local code before committing?

Good call that's what it was, missing 2 changes.

Also I tested #235 and #222 while working on this change and couldn't reproduce either so I believe they're fixed.

craftworkgames:
Awesome work. I've closed those other 2 issues.

I'll merge this now.

HyperionMT:
Thanks @craftworkgames.  Also, I realize I didn't use the correct keywords to get #144 and #268 to close automatically (I promise I'll be better about this in the future).  Could you please close those two out when you get a chance?

PR #293, opened 2016-11-06
Minor bugs in TextureAtlases

Spessi:
**TextureAtlas.cs:**
The region name is now printed correctly in the exception.

**TexturePackerWriter:**
If I put sprites in CodeAndWeb TexturePacker in directories (Blah1/1.png, Blah2/1.png), they both are recognized as "1", because not only the file extension is removed, but also the file path. It's better to cut off the extension only (i.e. change the extension to null).


craftworkgames:
Great work :)

Issue #171, opened 2016-04-18
Mouse listener still active when window is Inactive


Scellow:
The Mouse listeners is still listening for inputs even if the window is not active (background, minimized)

I believe it's an issue since we can't disable it without removing them

OS tested:

Windows 10 x64

``` csharp
namespace TheRoguer.Client.Legacy
{
    public class PlayerController : InputListenerManager
    {
        private KeyboardListener m_keyboardListener;
        private MouseListener m_mouseListener; 

        public PlayerController(Realm realm) : base()
        {
            m_mouseListener = AddListener(new MouseListenerSettings());
            m_keyboardListener = AddListener(new KeyboardListenerSettings());

            m_mouseListener.MouseClicked += On_MouseClicked;
            m_keyboardListener.KeyPressed += On_KeyPressed;
        }

        private void On_KeyPressed(object sender, KeyboardEventArgs e)
        {
            Console.WriteLine(e.Key);
        }

        private void On_MouseClicked(object sender, MouseEventArgs e)
        {
            Console.WriteLine(e.Button);
        }
    }

}
```

Proposition:

I can check myself the state of the window and decide to handle the inputs or no, but it would be nice if we could add that option in the InputListenerSetting

Scellow:
My bad, it only affect the mouse listener!

One more thing, while window is active, if i press elsewhere the cursor comes back in the edge of that window

craftworkgames:
Yeah, good catch. Do you know how to detect the state of the window? Could you post the code.

ExCellRaD:
You can check if the game is active 
`if(Game.IsActive`
and if the cursorposition is within window bounds 
`&& GraphicsDevice.Viewport.Bounds.Contains(Mouse.GetState().Position))`

Aurioch:
Well, I have tried to address the issue.

The thing is... it is sort of an intended behavior, as the `Update` method is being called from the game even when it doesn't have focus, so it's the user's (that is, programmer) job to handle it (such as `if (Game.IsActive) _mouseListener.Update(gameTime)`). Unless we make it more complicated by passing `Game` as a parameter, there's no way to make the listener pause itself... and frankly, I don't think we should do it.

However, there is a small workaround: if you create and register `InputListenerComponent` and attach listeners to it, it will automatically pause all attached listeners when the game loses focus.

[...]

PR #299, opened 2016-11-07
Nuclex GUI

Aurioch:
Woohoo, I finally finished porting Nuclex with... let's say 80-90% usability.

So. I've put all the relevant code into `MonoGame.Extended.NuclexGui` namespace so that it doesn't clash with existing `MonoGame.Extended.Gui`, which is good to leave as a backup system we can work on. Main parts - and the ones that will be mostly used - are `GuiManager` class and controls inside `NuclexGui.Controls` namespace.

Only part of the code outside of `NuclexGui` namespace is `InputManager` class, located in `MonoGame.Extended.InputListeners` namespace. `InputManager` is used for hosting and managing main `InpuListeners` and is mandatory for `GuiManager` to work.

From the original version of Nuclex, I've disposed the plugin mechanism (though loading of renderers is still present and heavily abuses reflection and LINQ - check `FlatGuiVisualizer.FetchRenderers` method) and replaced Nuclex' devices with `InputListener` classes. Therefore, there's only an reliance on `InputListeners` namespace due to that and `InputManager`.

There is also a demo project showcasing the usage of the controls, with some commentary.
capture

I'll continue to work on squashing bugs as I stumble upon them, and will add new features on demand. Even if that demand is my own.
Also, I'm not sure about design choice of `InputManager` - it currently tracks only one `KeyboardListener`, one `MouseListener`, one `MousePadListener` and one `TouchListener`. I'm not sure how those classes handle multiple input devices of the same kind, and whether I should worry about that or not.


craftworkgames:
Sorry I didn't get around to merging this tonight. Ran out of time.

Could you please check the broken build? 

 - Don't forget to merge `develop` into your branch first. (that'll fix one of those broken tests)
 - I noticed there's a merge conflict with `ColorJsonConverter.cs`. I changed that recently, take care when merging this as to not reintroduce an old bug.

Other than that, it looks great :)

Aurioch:
There. I've dealt with the issue ^^ And it still works!

I also got rid of `InputManager` class and moved its functionality to `InputListenerComponent`, while at the same time dealing with #171. Input listeners still have to be manually stopped when game loses focus, but at least `InputListenerComponent` will do it for all listeners registered to it.

I've also made a small change to `ColorJsonConverter` by moving out `string` to `Color` conversion to `ColorHelper` class. Since it's used in `GuiSkinJsonConverter` as well.

craftworkgames:
Good work man.

I did a lot of code clean-up tonight. Mostly running Resharper over everything to get some code style consistency.

The key things I changed with the Nuclex code:

 - I've flipped it around so that the `InputListenerComponent` is decoupled from the GUI code. By using constructor injection, we can force people to set it up in the right order.
 - I felt a little uncomfortable with the `Support` namespace, so I moved that into the NuclexGui namespace. To be honest, I'm not sure where those classes should live at this stage.

Both of these changes could be improved. The truth is, I'm really tired, so I just wanted to make a start on it and go to bed.

Actually, there's still quite a few little code clean-up things that could be done but we'll get to those in time. I'm still thinking about how / if we should split this (and other features) out of the core project.

craftworkgames:
Oh and I moved the `Demo.NuclexGui` project into the `Demos` folder like all the others.

OptimusPi:
Thanks for having that in the demos folder. I am excited to run it and study it. Thanks Aurich, you're amazing!

Scellow:
Wouldn't it be better to use ``BitmapFont`` instead of ``SpriteFont`` ?

craftworkgames:
>Wouldn't it be better to use BitmapFont instead of SpriteFont?

Probably yes. Ideally, we'd let people choose, but that might not be worth the effort.

PR #300, opened 2016-11-09
Made BitmapFont ctr public

Scellow:
I'm not using the MonoGame content pipeline, so i need the contructor to be public to be able to manually load a BitmapFont from file


craftworkgames:
I guess there's no harm in making it public. Thanks.

PR #302, opened 2016-11-10
Implemented animated tiles

Spessi:
I'd recommend to use the "very-large" demo map (you have to adjust the camera to look at 0/0 (I didn't want to change the lookAt value for all maps in the demo application).

Open for discussion:
* Only Tiles are animated. Are animated object and image layers are needed? 
* Is there a better way to get the camera viewport?
* Should we store the absolute tile position in each TiledTile? So we wouldn't need to create a new Rectangle for each tile in the "is in viewport"-check.
* I've moved the VBO and IBO construction into the GroupRenderDetails constructor because I think it's more appropriate there (it also allows to use SetVertices without recreating a new vbo in every call)

### DIFF
--- a/Source/Demos/Demo.TiledMaps/Game1.cs
+++ b/Source/Demos/Demo.TiledMaps/Game1.cs
@@ -83,6 +83,7 @@ namespace Demo.TiledMaps
             var keyboardState = Keyboard.GetState();
             var mouseState = Mouse.GetState();
 
+            _mapRenderer.Update(gameTime);
             if (keyboardState.IsKeyDown(Keys.Escape))
                 Exit();
 
--- a/Source/MonoGame.Extended/Maps/Renderers/DynamicGroupRenderDetails.cs
+++ b/Source/MonoGame.Extended/Maps/Renderers/DynamicGroupRenderDetails.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Xna.Framework.Graphics;
+using MonoGame.Extended.Maps.Tiled;
+
+namespace MonoGame.Extended.Maps.Renderers {
+    class DynamicGroupRenderDetails : GroupRenderDetails
+    {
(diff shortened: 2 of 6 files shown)


Spessi:
@LithiumToast Thank you for your feedback! I've moved the List<VertexPositionTexture> and the array into the private class scope. To only update the animated tiles, I've added a new List which holds the animated TiledTilesetTiles (not TiledTile). The list is created once on map loading.

craftworkgames:
Pretty good.

There seems to be a problem when zooming in on animated tiles. They stop animating at close range for some reason.

Also, please try to use more descriptive variable names and respect the code style. `gd` is not enough. It's super annoying for me to have to go through the code afterwards and fix everything up.

I don't expect everything to be perfect but I'd like to maintain a readable code base as much as possible. After all, we spend much more time reading the code than writing it.

Spessi:
> gd is not enough. It's super annoying for me to have to go through the code afterwards and fix everything up.

Well, I didn't introduce "gd" as variable name. I wanted to be consistent and use the same variable names like they were called before. But in the future I'll rename them if I think they are too indescriptive.
Could you give me an example for bad code style, please? Maybe I just need to adjust my ReSharper settings. Thank you.

craftworkgames:
> Well, I didn't introduce "gd" as variable name. I wanted to be consistent and use the same variable names like they were called before.

My apologies. Although, this is a perfect example of why it's so important. When we let things like this get by for too long, before you know it, everyone is doing it.

> But in the future I'll rename them if I think they are too indescriptive.

Yes, please do. 

> Could you give me an example for bad code style, please? Maybe I just need to adjust my ReSharper settings. Thank you.

The best example is the actual code itself. 95% of the code in the rest of the library follows the same style, if you notice something out of place, feel free to fix it.

Default ReSharper settings are usually fine. 

Again, sorry for blaming you. I just happened to notice it in this PR. It's been a while since I've really properly looked at some of the new code coming in.

I'm just about to do a big clean-up commit.

LithiumToast:
@craftworkgames I didn't have the time to do a full review of this before you merged it, I'll look for naming problems and inconsistencies in future reviews

PR #304, opened 2016-11-13
Include IShapeF object as part of the TiledObject

HyperionMT:
I noticed we're not parsing out TiledObjects into their associated Shapes and instead just storing the bounding rectangle.  I've updated the TiledMapReader and TiledObject to now store the shape for easy access.  This required creating 2 new shapes: EllipseF and PolylineF.  I realize that having both an EllipseF and CircleF is a little redundant but CircleF had a few methods I wasn't ready to implement on EllipseF.


craftworkgames:
Okay. Great :)

PR #316, opened 2016-11-27
Create Graphics library.

LithiumToast:
- Add `MonoGame.Extended.Graphics` library by copying existing code as it is over.
- Use `Profile111` for all PCL libraries.
- Fixed code that would not compile in `Demo.InputListeners/Game1.cs`.


craftworkgames:
Cool man. I'll merge this tonight.

What's the deal with `Profile111`?

LithiumToast:
`Profile111` allows for consumption on all platforms including Xamarin iOS / Android with the exception of Silverlight. It's also equivalent to .NET Standard 1.1

craftworkgames:
> Profile111 allows for consumption on all platforms including Xamarin iOS / Android with the exception of Silverlight. It's also equivalent to .NET Standard 1.1

I was pretty sure I already selected Android and iOS when I set it up, that's why I was wondering what `Profile111` brings to the table.

According to [this table](http://danrigby.com/2014/05/14/supported-pcl-profiles-xamarin-for-visual-studio-2/) `Profile111` supports Windows Phone 8.1 whereas `Profile7` does not.

    Profile 7   (.NET Framework 4.5,   Windows 8) (netstandard 1.1)
    Profile 111 (.NET Framework 4.5,   Windows 8, Windows Phone 8.1) (netstandard 1.1)
    
It's good to have a better understanding of how it works. I'll keep it in mind when creating more of them. It's also great to know it meets .NET Standard 1.1

PR #317, opened 2016-11-28
DynamicBatch rework

LithiumToast:
-Renamed `DynamicBatch2D` to `DynamicBatchRenderer2D`.
-Refactored `GeometryBuffer` into `GraphicsGeometryData`. Vertex and index buffers are now the responsibility of `BatchCommandDrawer` class.
-Added `GraphicsGeometryBuilder` which is to be inherited from to build the geometry of various shapes and whatnot. There is currently only `SpriteBuilderPositionColorTextureU16` which has the responsibility of building a single sprite with `VertexPositionColorTexture` for vertices and `ushort` for indices.
-Added the ability to enqueue draw commands with different primitive types into the dynamic batch renderer. This is important as drawing shapes in the near future will use line primitive type for stroke and triangle primitive type for fill.


LithiumToast:
I'm not sure exactly why my previous commits that were already merged are shown here 😕

craftworkgames:
> Renamed DynamicBatch2D to DynamicBatchRenderer2D

I noticed you've renamed this thing a few times. Would you like to discuss the naming?  The word "dynamic" has always felt a little off to me, although, I'm not sure what else to call it either.

I'll merge these changes in now.

craftworkgames:
I tried out the Batching demo and for some reason the FPS is capping out at 60 frames for me. No idea why.

LithiumToast:
Oh, that's because I had fixed time step enabled. I was comparing between renderers at 60 FPS in release mode and they both are on par.

craftworkgames:
Ha, I actually saw that line, but I overlooked it. Must've been tired.

PR #322, opened 2016-12-12
Tiled library

LithiumToast:
-Seperates out Tiled code into library `MonoGame.Extended.Tiled`.
-Only changes are from ReSharper's cleanup tool.


craftworkgames:
What do you mean by "remove test project"?

Is the test project split out also?

Sorry, I'll get around to merging this soon.

LithiumToast:
Uh, I was trying to have separate test projects like you had going for the GUI, but then I realized there is no reason not to have everything in one project for tests.

craftworkgames:
> Uh, I was trying to have separate test projects like you had going for the GUI, but then I realized there is no reason not to have everything in one project for tests.

Yeah, you're right. There's no strong reason to split out the test projects. Although, I'd still like to do it just to keep things organized that way.

For now I'll just leave it as is and have a think about the pros and cons of both approaches.

PR #330, opened 2017-01-09
JSON.NET & .nuspecs

LithiumToast:
-Upgrade JSON.NET to latest stable for all projects.
-Add some experimental NuGet packages. @craftworkgames You will probably need to double check the `.nuspec` files are correct and do work correctly.


craftworkgames:
Seems to work well. Thanks :)

PR #331, opened 2017-01-17
Tiled Rework

LithiumToast:
Pretty large... but here it is.

The content pipeline unfortunately doesn't allow one to many relationship for content files. It's impossible to split the meta data of a .tmx file (XML) from the rendering code and only load one but not the other. By doing the geometry processing on the content pipeline side however, future work to merge tilesets into one larger texture atlas will be fairly straightforward and won't require any change to runtime code. (Creating this image for the larger texture atlas will have to be done using the content pipeline.)

-Renamed all Tiled classes to use convention `TiledMap...` for both runtime and content pipeline libraries. This includes replacing `FullMapRenderer` with `TiledMapRenderer`.
-Added support for drawing layers of a Tiled map seperately in `TiledMapRenderer`. Also allows to draw a layer at a certain Z depth.`IsVisible` and `Opacity` can also be changed at runtime for a `TiledMapLayer` and be reflected correctly when rendering.
-Refactored code not to use LINQ excessively (there was quite a bit generating garbage). This also reduces the processing time when building or loading maps.
-Re-wrote content pipeline code to process the geometry for the models of a `TiledMapLayer` to render at runtime. The content reader simply just reads the necessary binary data at runtime, nothing more.
-Animated tileset tiles only cycle through texture coordinates at runtime now instead of copying around vertex data and texture coordinates.
-Removed support for `TiledMapObjectLayer` rendering. I don't think it should be the intent of this module to say how people will organize their game objects or entities, if they even want to use Tiled for this purpose even.
-Changed the origin location for isometric Tiled maps; `(0,0)` is the top of the first tile. see http://clintbellanger.net/articles/isometric_math/
-Updated the `TiledMaps` demo to showcase graphics device metrics and camera space origin.
-Added some things missing from the Tiled map format including `TiledMapObjectLayer.Color` and `TiledMapObjectLayer.DrawOrder`.
-Temporarily disabled tests for Tiled maps until code is settled. (To be fair there wasn't many tests to begin with.) I did however add the test library projects for Tiled code, content pipeline and and runtime.

EDIT:

-Oh yes, forget to mention that content pipeline Tiled code has moved to `MonoGame.Extended.Content.Pipeline.Tiled` library. Developers will need to add 
`MonoGame.Extended.Content.Pipeline.Tiled.dll` to their references in their content project file.
Also, the content pipeline Tiled library is not attached to any NuGet package (yet).


craftworkgames:
Absolutely fantastic work! Quite possibly the most wanted contribution to the library.

I'll merge it soon (currently not in front of a computer)

LithiumToast:
Thanks! I know it is large, but it would be great if you could review all of it when you get the time _before_ you merge it. That way I can make necessary changes.

LithiumToast:
So I noticed something strange going on with memory allocations here. I'm using `ReadOnlyCollection<T>` to wrap most arrays or lists to make them appear immutable to the API. However, doing a `for each` iteration over a `ReadOnlyCollection<T>` causes an enumerator to be allocated on the heap. Devil's in the details, but it appears [`ReadOnlyCollection<T>.GetEnumerator()` returns `IList<T>.GetEnumerator`](https://referencesource.microsoft.com/#mscorlib/system/collections/objectmodel/readonlycollection.cs,aafef79f3fced6c2). 

```
        public IEnumerator<T> GetEnumerator() {
            return list.GetEnumerator();
        }
```

> [Even though we’re still doing a foreach over a List, when the list is cast to an interface, the value type enumerator must be boxed, and placed on the heap.](https://blogs.msdn.microsoft.com/etayrien/2007/03/16/foreach-garbage-and-the-clr-profiler/)

> When doing a foreach over a Collection<T> an enumerator WILL be allocated.
> When doing a foreach over most other collections, including as arrays, lists, queues, linked lists, and others:
>     if the collections are used explicitly, an enumerator will NOT be allocated.
>     if they are cast to interfaces, an enumerator WILL be allocated.
Thus, garbage is generated when using `for each` over `ReadOnlyCollection<T>` in the form of boxing the enumerator with `IEnumerator<T>`.

Perhaps the solution here is to create a special collection class that fulfills the purpose of `ReadOnlyColletion<T>` while not boxing the enumerator.

craftworkgames:
Hmm.. I was going to say to have an `IReadOnlyCollection<T>` instead of `ReadOnlyCollection<T>` on the interface but I'm guessing that doesn't solve the problem either.

> Perhaps the solution here is to create a special collection class that fulfills the purpose of ReadOnlyColletion<T> while not boxing the enumerator.

I think you're right. A custom class is probably the way to go here. Although, I don't think it's worth waiting for. I'm going to merge these changes now.

Btw.. what tools to do use to profile the memory allocations?

craftworkgames:
> Thanks! I know it is large, but it would be great if you could review all of it when you get the time before you merge it. That way I can make necessary changes here by appending commits.

This is not really practical. Even if I could review *all* of the code in one hit, the changes I'd like to make are going to be discovered by actually using the code rather than reading it.

I'd much rather have a series of smaller PR's rather than one big one that stays open for a long time.

PR #332, opened 2017-01-19
Fix Tiled Problems

LithiumToast:
-Add the `MonoGame.Extended.Content.Pipeline.Tiled.dll` to the files for the `MonoGame.Extended.Tiled` NuGet package so it can be used to build Tiled map assets.
-Remove dependency on `MonoGame.Extended.Graphics` for `MonoGame.Extended.Tiled`.
-Fixed all demos that use Tiled maps.
-Updated the platformer demo demonstrating drawing things between layers.

screen shot 2017-01-19 at 8 36 29 am


craftworkgames:
Looks great.

I think you're right. It does seem more consistent calling it `Size2`.

To be honest, I sometimes wonder if it was a mistake even introducing size at all. We could just drop it completely in favor of using `Vector2` again.

LithiumToast:
Oh damn `Size2` change wasn't suppose to be in this PR... 

I know it is personal preference, but I actually dislike how XNA uses `Vector2` to mean 3 possible things: a point (position), a direction, or a size. I think doing that makes the code less readable. Not everyone has an education is linear algebra, so a special size type such as `Size2` is probably more intuitive to use and think about then a mathamatical vector type such as `Vector2`. Though game developers should absolutely have education in linear algebra that is often not the case for some developers, especially beginners. `Size2` and `Vector2` can be implicitly casted in both directions anyways. Code like this will work fine:

```
Size2 size = new Vector2(5, 10);
Vector2 size = new Size2(15, 5);
```

Issue #333, opened 2017-01-20
[Particles] Particle System is not data driven


jeffgamedev:
The particle system as exhibited in the demo is great for creating dynamic graphics quickly, but in it's current state, all particle effects are hard coded. It would be nice to have an editor that the user can preview and generate JSON to load in the game later, instead of having to change C# code every time you want to tweak a particle.

libGDX has such a particle editor: https://github.com/libgdx/libgdx/wiki/2D-Particle-Editor

craftworkgames:
I plan to work in this after I've finished (nothing is ever finished ;) ) the GUI system.

I was thinking, maybe we can create a particle editor with the new GUI system perhaps.

jeffol:
Sounds like a good plan! The particle editor can be a simple project:
I'd guess the major features would be:

- Adjust / change particle settings

Adjust particle settings

jeffgamedev:
Particle System has been data driven for a while, and external editor is pending release.

Issue #100, opened 2016-01-30
Entity Component System


LithiumToast:
What do you guys think of adding an [Entity Component System](https://en.wikipedia.org/wiki/Entity_component_system)?

craftworkgames:
I wrote an [entity component system](https://github.com/craftworkgames/Astrid.Framework/tree/master/Astrid.Framework/Components) once and part of an old (failed) game framework. It was really fun to write and works pretty well.

Is it a good fit for MonoGame.Extended?  Probably, but we'd have to think carefully about how it fits in with the rest of the framework. I wouldn't want to force it upon people that they have to use an ECS.

For example, in an ECS you usually have some sort of transform component (position, rotation, scale) and a separate sprite component (textureRegion, color, origin, visibility).

The way we have it now; those 2 concepts are combined in the `Sprite` class. I suppose what we could do is keep the ECS code in a separate namespace and make sure all the component classes are named appropriately. Like, `SpriteComponent` for example. 

This kind of thing has crossed my mind a couple of times. An ECS is certainly a nice way to avoid the deep and wide inheritance hierarchies.

cupstudio:
Component is a great pattern, and I've tackled it myself once as an exercise.  The trickiest part for me was communication among the components themselves.  Use a global message queue?  Dependent components get copies of each other?  The entity itself sends and receives messages among its components?  Have all attributes live inside the Entity to be manipulated directly?  There are a lot of options out there with + and - to each, so I'd love to see how someone more experienced handles it.

I think a very light component system with Interfaces and / or base abstract classes could be doable, just make it really easy to extend.

craftworkgames:
> The trickiest part for me was communication among the components themselves.

Yeah, this is always the challenge. I remember trying lots of different designs in my system. From my vague recollection I recall doing something like this that kinda worked pretty well.

   var transform = entity.AddComponent(new TranformComponent());
   var physics = entity.AddComponent(new PhysicsComponent(transform, shape, mass));
   var sprite = entity.AddComponent(new SpriteComponent(transform, textureRegion));

The idea was to essentially *glue* components together with a 3rd component. In this example, the physics component moves the sprite component around using a transform component. The cool thing about this was that you could attach a physics component to anything that could be transformed (e.g. sprites, camera, particle emitters, etc).

Btw.. I'm not advocating this as the way we need to do it in MonoGame.Extended. I just want to start a discussion about the pros and cons of different approaches.

[...]

PR #334, opened 2017-01-20
Entity-Component-System Redesign

robertdhernandez:
I took a look at the upcoming Entity-Component-System (ECS) and found it to be extremely confusing to use. I went in and modified the framework to adhere more inline to the ECS design pattern. I based it largely on [this article](https://www.gamedev.net/resources/_/technical/game-programming/understanding-component-entity-systems-r3013).

It isn't complete, but it's enough of a framework to start building off of. It's still missing a lot of its events functionality, which is what really glues it all together. The old ```ParticleEmitterSystem``` is completely broken, which are their own issue (#333). And a few demos are broken because of the changes, but I have modified the EntityComponentSystem demo to demonstrate the new API. It can use some more testing, but I wanted to get the idea.

This might be a rather drastic change out of the left field, but I found the existing upcoming entity-component-system API to be very lacking and confusing. Hopefully this new system is more straight-forward than the existing one.


LithiumToast:
Couple things I noticed:
- Transform data is separated into it's own component. There was a discussion about this before. Unity for example does this, but is it really necessary? Are there examples of entities which don't have a transform component? If so, then yeah this probably makes sense. If not, then why bother with the abstraction? Why not just move the transform data directly to the entity? There might be some good reasons to decouple the two but it needs some rational in that case.
- Systems no longer process components? Entities attach systems? What is going on here? The purpose of ECS is that each system does processing with a set of components for each entity. Components don't have to be stored directly with the entity, but they should be accessible from the entity in some fashion.

LithiumToast:
Also the following are relevant:
- #100
- http://gameprogrammingpatterns.com/component.html

craftworkgames:
Yep. I agree with @LithiumToast on this one.

I'm all for discussing and experimenting with new designs but I can't pull this into `develop` in it's current state.

The thing about Entity Component Systems is that they are more concept than implementation. There's loads of blog posts around about the design and implementation of an ECS. Most of them share the same ideas but can have wildly different implementations.

    _logoEntity.AttachComponent(spriteCollection);
    _logoEntity.AttachSystem(new SpriteBatchSystem(GraphicsDevice, _camera, spriteBatch));
    _logoEntity.AttachSystem(new RotatorSystem());

The thing that really threw me about this approach is the way systems are attached to entities? Maybe I've missed something, but I've never seen it done this way before. As far as I can tell, doing it this way would have a serious impact on memory and performance. I am curious about the thought process behind it though?

Don't get me wrong. The current ECS implementation has a lot of room for improvement but we need to consider how it fits with the rest of the library.

robertdhernandez:
I've been doing a little more research into the entity component system model, and found this to be very inefficient. I was also building it off of my understanding of an entity component system from another game that uses it a way similar to this. I do plan on revisiting this topic.

LithiumToast:
If you want to talk through it let me know or open an issue about it. Having a good ECS implementation would be great 😄

KinoshitaShimizu:
Hmm, I do also implement my own ECS (a bit like Unity does), though ended up naming all the items the same (GameObject, redundantly). I wonder if I could help on this, though, I believe my paradigm is a bit different than everyone here. For example, I like easily serializable-deserializable items, so an editor can use them properly... but I've read some classes and though I can serialize them for XNA, I don't seem to be able to deserialize them like that.

Currently I thought in adopting the ECS you use, because else I have to wrap everything around (I think I'd have to do so anyways). If it is possible I'd like to help on this matter, if still requiring help.

PR #337, opened 2017-01-21
Update Entity.cs

z4kk:


z4kk:
_nextEntityId as static field of Entity

craftworkgames:
Okay, so there's a number of problems.

 1. You've submitted 2 separate PR's (this one and #336) for changes that are closely related. These should really be a single PR. You can keep adding commits to a PR even after it's been submitted.

 2. They both broke the build. All PR's must compile before I'll accept them.

 3. Most importantly, you haven't explained *why* you've suggested this change. Telling me *what* you've changed has very little value. I can already see that in the diff. What I really want to know in the comments is *why* you think this implementation is better than the old one.

I'm not trying to be mean. I don't want to discourage you from contributing. But dealing with PR's is already quite time consuming. I have to draw a line somewhere.

PR #338, opened 2017-01-23
Entity-Component-System Redesign

robertdhernandez:
I took another shot at making an entity component system, based on there [blog posts](http://t-machine.org/index.php/2007/09/03/entity-systems-are-the-future-of-mmog-development-part-1/) which define the system properly. 

I have only updated Demo.EntityComponentSystem and Demo.Platformer to demonstrate the new system.


LithiumToast:
What's the state with #336 and #337?

craftworkgames:
I haven't had the chance to take a proper look at this yet. Rest assured it is on my to-do list.

craftworkgames:
There's definitely some interesting ideas in this but it's quite different. I'm not ready to pull a big change like this into `develop`. We can keep discussing it as an issue though.

 The PR is getting rather old and there's far to many changes to cherry pick the merge. Closing.

PR #340, opened 2017-01-28
Graphics refactor

LithiumToast:
-Fixed `TiledMapRenderer` bug with state changes by removing `Begin`/`End` methods pair.
-Removed `Renderer`.
-Renamed `DynamicBatchRenderer2D` to `Batcher2D`. Performance improvements on `Batcher2D`.
-New `Batcher` base class; to be derived in similar fashion as `Batcher2D` for future `UserInterfaceBatch` (or `UIBatch`) to support batching with scissor rectangles.


craftworkgames:
Great work as usual :)

PR #345, opened 2017-02-23
TiledMapRenderer allow custom effect

### LINKED ISSUE #343: [Tiled] Tiled Maps no longer support Effects
After the Tiled Map rendering was refactored, removing it from Sprite Batch operations, it is no longer possible to render the tiled map with custom Effects (fx shaders loaded through the Content Manager).

LithiumToast:
Fixes #343 

Custom effects can be passed  to `Draw` overloads of `TiledMapRenderer`. Custom effects do however require to implement `ITiledMapEffect` to get and set the active texture and alpha transparency for a custom effect.

### DIFF
--- a/Source/MonoGame.Extended.Tiled/Graphics/ITiledMapEffect.cs
+++ b/Source/MonoGame.Extended.Tiled/Graphics/ITiledMapEffect.cs
@@ -1,0 +1,11 @@
+﻿using Microsoft.Xna.Framework.Graphics;
+
+namespace MonoGame.Extended.Tiled.Graphics
+{
+    public interface ITiledMapEffect : IEffectMatrices
+    {
+        EffectTechnique CurrentTechnique { get; }
+        Texture2D Texture { get; set; }
+        float Alpha { get; set; }
+    }
+}
--- a/Source/MonoGame.Extended.Tiled/Graphics/TiledMapDefaultEffect.cs
+++ b/Source/MonoGame.Extended.Tiled/Graphics/TiledMapDefaultEffect.cs
@@ -1,0 +1,19 @@
+﻿using Microsoft.Xna.Framework.Graphics;
+
+namespace MonoGame.Extended.Tiled.Graphics
+{
+    public class TiledMapDefaultEffect : BasicEffect, ITiledMapEffect
+    {
+        public TiledMapDefaultEffect(GraphicsDevice device) 
+            : base(device)
+        {
+            TextureEnabled = true;
+        }
+
+        public TiledMapDefaultEffect(BasicEffect cloneSource) 
+            : base(cloneSource)
+        {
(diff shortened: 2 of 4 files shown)


craftworkgames:
Nice work. It would be good to include this in the demo at some point.

@jeffol I'd love to see what you're using this for. Screenshot? :)

jeffol:
Hey @craftworkgames looking to use this for some simple shader effects like black and white.

bwshader

Currently, I can pass an Effect object I loaded through the ContentManager into the Sprite Batch and it renders my sprites in black and white.

bwshaderload
bweffectbatch

With @LithiumToast 's current implementation, he says it's possible to use Effects now, but they can't simply be passed into the Tiled map renderer and just work. He says this isn't possible, but I disagree. We used to have a tiled map renderer that we could easily just pass an effect object and it renders the map with it. I'm willing to use the new way to get the effects into the map renderer, but I don't think it's very usable and easy to figure out. I've been unable to get it to work myself and I've asked @LithiumToast to include an example of an fx file being loaded and used with the tiled map rendering class.

I've attached the basic black and white shader (fx) file I'm using.

[bwshader.zip](https://github.com/craftworkgames/MonoGame.Extended/files/800540/bwshader.zip)

LithiumToast:
I'll re-iterate what I said in the Gitter chat in a more clear and condensed version:

- `TiledMapRenderer` needs to somehow communicate two pieces of information to the fragment (pixel) shader: (1) the texture to sample from so there is an image on the screen, and (2) the transparency value for each layer, `TiledMapLayer.Opacity`. 

For the texture, there are two options: (1) Use an `EffectParameter` to pass a `Texture2D` to the underlying shader which `Effect` does not support out of the box (need to sub-class), or (2) set `GraphicsDevice.Textures[0]` to the correct `Texture2D` *after* `pass.Apply()` and *before` `GraphicsDevice.DrawIndexedPrimitives`. 

We currently use option 1 for the texture which I'm fine to switch to option 2, however... The pixel shader still needs the `TiledMapLayer.Opacity` information for each layer which has to be an `EffectParameter`; it is the only way to pass an arbitrary value from runtime code to the shader. So either way... Passing a vanilla `Effect` class may work for games which don't use `TiledMapLayer.Alpha` using texture option 2, but for games which do use `TiledMapLayer.Alpha` *and also do use custom effects*, there would be a bug introduced by using a vanilla `Effect` class using texture option 2 for rendering. Namely, `TiledMapLayer.Opacity` would not work as expected. My solution was to simply have an interface, `ITiledMapRenderer` which exposes two properties: `Texture` and `Alpha` which serve as the "glue" to communicate those two pieces of information from `TiledMapRenderer` to a custom `Effect`, which in turn is responsible for passing that information to the shader using both option 1 of using an `EffectParameter` for the texture and the opacity value.

PR #346, opened 2017-02-23
Geometric Primitives Cleanup

LithiumToast:
Before I can draw them using `Batcher2D` I need fix up the shapes. This one is for `CircleF`. I added tests.


LithiumToast:
I'll fix this up soon.

LithiumToast:
Cleaned up `RectangleF` . Added a couple new methods. Need to add tests next.

craftworkgames:
I thought we agreed to leave the shapes in the `Shapes` namespace?

LithiumToast:
We agreed they are not shapes but geometric primitives?

craftworkgames:
image

Communication is hard 😄 It didn't help that I was typing those messages on my phone while I was out of the house I guess.

To be honest I don't really care *that* much if we move them. I can see arguments both ways. I was just a bit surprised when you moved them after we talked about it.. lol.

The thing is, there's no "right" way to do this stuff. I've worked on many projects over the years. When you get 5 programmers in a room to discuss how something should work, or how it should be done, you end up with 5 different ideas and nobody can agree on just one way.

> We agreed they are not shapes but geometric primitives?

It think this is where we started misunderstanding each other. When they where in the `Primitives` folder I didn't really click that you where thinking about *geometric primitives*. My first thought was to think about [primitives](https://en.wikipedia.org/wiki/Primitive_data_type). Oops.

A geometric primitive is certainly one way to think about it. They certainly fit the [wikipedia definition](https://en.wikipedia.org/wiki/Geometric_primitive). Although, it can also be argued that they are still shapes. Even a straight line can be [considered a shape](http://gamedev.stackexchange.com/questions/34225/can-a-straight-line-be-called-a-polygon).

Anyway, hopefully we're starting to get on the same page now. Sorry for the confusion.

I still don't like `Primitives` as a namespace. You obviously don't think they belong in `Shapes`. So what are we going to do?  Should we move them into the root namespace or not?

My preference was actually to not change anything and leave the `Shapes` namespace as is. You obviously feel pretty strongly about it though. So if you really want to do this we'll do it.

craftworkgames:
You know what. Let's just go with what you've done. It's pretty good.

LithiumToast:
It's probably best just to match XNA style, since that is what people are use to with MonoGame, and that is what we are extending. XNA has `Rectangle` (2D, integer fields), `BoundingBox` (AABB in 3D space, floating-point fields), `BoundingSphere`, `Plane`, `Ray` (3D space), etc. They all live in `Microsoft.Xna.Framework` namespace; the core namespace. So this would mean new geometric primitives should probably live in the `MonoGame.Extended` namespace; which is was what I originally had done. Just existing `RectangleF` and other geometric primitives are under `MonoGame.Extended.Shapes`. 

Technically they are shapes, but they are "primitives" in the sense that they are structures in C#. Now technically `BoundingBox` is a bounding volume "primitive" while `Rectangle` is a geometric "primitive". For our purposes they are both just "primitives".

Also it's unfortunate that the XNA team didn't include 2D primitives from the start... When introducing 2D primitives the naming can get confusing between 2D and 3D. Other game engines and frameworks that deal with primitives for computer graphics such as OpenCV, Urho3D and even .NET Core, use various naming conventions... I purpose the naming convention along the lines of `{Name}{2|3}{I|F}`. `2` is for 2D which is what we are already familiar with `Vector2` for example, I already added `Point2` and `Size2`. Similarly `3` is for 3D. However when it is clear that the primitive can only belong in a 2D or 3D context just drop this suffix, for example `RectangleF` is clearly a 2D primitive, while `Plane` is clearly a 3D primitive. `I`, or `F` are for what type are the fields. This is what we are already use to with `RectangleF` and if we were to introduce `RectangleI` it would be equivalent to `Rectangle` from XNA which all the fields are integers. This naming scheme allows more primitives to be added later while conveying the necessary information in the full name of each primitive.

craftworkgames:
Yeah, you're right. You've obviously put a lot of thought into this. Thanks for explaining the naming conventions so clearly.

The only thing I wouldn't do is the `I` suffix since MonoGame already has `Rectangle` let's just say that it's `F` for `float` fields and no suffix for `int` fields.

I noticed after the merge that there's only 2 classes left in the `Shapes` namespace. Should we just move them to the root like that others and get rid of `Shapes` completely?

LithiumToast:
I think `Rectangle` is named that way in XNA because a clash between a non-existent floating-point version and the integer version didn't exist. Again, annoying that XNA didn't just add a `RectangleF` to begin with... I don't think integer primitives are going to be that useful anyways so that's fine if there are one or two like that.

PR #347, opened 2017-03-02
[Tiled] Utility methods for dealing with tile bounding boxes

Lignum:
This PR adds several methods that help with dealing with bounding boxes of tiles, such as:
-  `TiledMapTile.GetBounds(int width, int height)`: Returns a `Rectangle` encompassing the tile, with `width` and `height` being the dimensions of the rectangle. Will return absolute coordinates.
- `TiledMapTile.GetBounds(TiledMapTileLayer layer)`: Helper overload for the previous method, taking `width` and `height` from `TiledMapTileLayer.TileWidth` and `TiledMapTileLayer.TileHeight`.
- `TiledMap.GetBounds(Predicate<TiledMapTile> includePred)`: Returns the bounding boxes of all tiles in the map matching the given predicate.
- `TiledMap.GetBounds(Predicate<TiledMapLayer> includePred)`: Same as previous, but returns bounding boxes of tiles in _layers_ that match the predicate.

Along with these methods, I've changed the `Demo.TiledMaps` project to demonstrate them. I've modified some of the layers in the test maps to have a `Solid` property, which in a real game would mark the tiles in these layers as unpassable. Solid tiles can thus be shown by pressing the B key, which uses the new `TiledMap.GetBounds` method.

Additionally, I've added a `TiledMapTileLayer.ForEach(Action<int, int, TiledMapTile> f)` method, along with a `TiledMapTileLayer.ForEach(Action<TiledMapTile>)` overload that simplifies iteration over all tiles in a layer, which I wrote to simplify the implementation of the other methods. Though this may be out of scope of the PR, so I'm happy to remove it if needed.


craftworkgames:
Sorry I haven't had a chance to look at these yet. They do look interesting but I'll need to take a closer look at them before merging.

Lignumm:
Alright, that's fine, no pressure 😄 .

craftworkgames:
@Lignum I've raised an issue #359 to discuss incorporating these kinds of changes into the library. I think we could really use some improvements in this area but I'd like to discuss the bigger picture first.

PR #349, opened 2017-03-03
Tiled Fix

LithiumToast:
- Fixes a namespace issue with a conflict between `MonoGame.Extended` library and `MonoGame.Extended.Tiled` library. This is what is preventing people from using the latest code.
- Removed `MonoGame.Framework.dll` dependency. Use Portable Class Library NuGet package instead. There is a bug with the NuGet package currently however with 3.6 where SDL.dll has to be copied to the bin folder manually for projects such as demos. This will be fixed with MonoGame soon.
- Upgrade all content pipeline libraries to be decoupled from core libraries such as `MonoGame.Extended`, `MonoGame.Extended.Graphics` and `MonoGame.Extended.Tiled`. This ensures that content pipeline only needs the necessary code for building content.
- Include custom effect example in demo as requested by @jeffol.


craftworkgames:
All these changes are great but I'm a bit concerned about the upgrade to 3.6.

I noticed a few people having troubles upgrading. I wonder if it would be worth waiting until 3.6.1? Is there a reason we need 3.6 at the moment?

Secondly, what do you mean by decoupling the content pipeline libraries from the core libraries? Is it really a problem that the content pipeline depends on both?

LithiumToast:
The immediate problem for 3.6 is already fixed for us in [pull/5542](https://github.com/MonoGame/MonoGame/pull/5542). I was having a heck of a time writing the custom shader code so I tried to mimic MonoGame's shaders, and just to be consistent, I matched version to version so there wouldn't be any problems in that regard. (There were some bugs fixed for effects in 3.6) I do however agree that having 3.6 as part of this PR is bad taste but I don't have the time to split it properly.

Decoupling the content pipeline is easy; the content pipeline does not reference the runtime libraries. The content writer is supplied with a string of the runtime type and content reader instead of using reflection to get that same information. The advantages is that debugging the content pipeline is simpler because only the content pipeline code needs to be debugged. It's often the case that errors can rise from building the content when building the project.

craftworkgames:
>  The content writer is supplied with a string of the runtime type and content reader instead of using reflection to get that same information.

Ah cool. I didn't realize that was the only coupling between them.

Okay. Let's do this.

craftworkgames:
I ran into a bunch of problems with the demos so I updated them to match the MonoGame 3.6 DesktopGL project template. They now copy the required DLLs automatically.

I've also added a readme to the Demos folder to explain it a bit.

PR #357, opened 2017-03-22
Fix project files so they can build from Linux

cra0zy:
Stuff done:
 - Use MG.Content.Pipeline.Portable for MG.Extended.Content.Pipeline project
 - Fix links to native libraries so they point to a location that works with both Windows and Linux


craftworkgames:
Great. Thanks @cra0zy 

It's always nice to know someone is running the project on Linux.

PR #362, opened 2017-03-30
- Fix Tileset.FirstGlobalIdentifier

### LINKED ISSUE #361: [Tiled] Bug loading TMX with multiple tile sets
Hello. I'm having a problem when building some tmx with the pipeline tool.
I have a map with multiple tileset files. 

If i use tiles from the first tileset, building process works great. The problem appers when i use some tiles from the second one.
If my guesses are right, TiledMapProcessor (line 245) is not taking in account "firstgid" attribute of tileset.

```
<tileset firstgid="1" source="Tileset 1.tsx"/>
<tileset firstgid="1025" source="Tileset 2.tsx"/>
```

sin titulo
[Maps.zip](https://github.com/craftworkgames/MonoGame.Extended/files/878263/Maps.zip)

Thank you!

brogine:
Fixes #361.
Check if anything else is needed please.

### DIFF
--- a/Source/MonoGame.Extended.Content.Pipeline/Tiled/TiledMapImporter.cs
+++ b/Source/MonoGame.Extended.Content.Pipeline/Tiled/TiledMapImporter.cs
@@ -55,6 +55,10 @@ namespace MonoGame.Extended.Content.Pipeline.Tiled
                     using (var file = new FileStream(tilesetFilePath, FileMode.Open))
                     {
                         map.Tilesets[i] = (TiledMapTilesetContent)tilesetSerializer.Deserialize(file);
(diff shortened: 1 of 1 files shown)


craftworkgames:
Okay, so the difficultly with these kinds of changes is properly testing them.

The first thing I did was create a new map in the demo [`level08.tmx`](https://github.com/craftworkgames/MonoGame.Extended/blob/develop/Source/Demos/Demo.TiledMaps/Content/level08.tmx) to reproduce the bug. Once I confirmed that I was getting the same error reported in #361 I pulled your changes.

Unfortunately, the changes broke a bunch of other maps in the demo so I've now reverted them.

To fix this issue properly we need to make sure all maps in the demo are working as expected.

PR #368, opened 2017-04-04
Fix #361

### LINKED ISSUE #361: [Tiled] Bug loading TMX with multiple tile sets

LithiumToast:


LithiumToast:
This build fails because `level08.tmx` is a staggered map.

Issue #365, opened 2017-04-01
BitmapFont memory usage.


LithiumToast:
- Use `StringBuilder` for `SpriteBatchExtensions.DrawString` and `BitmapFont.MeasureString`.
- Change `BitmapFont.GetGlyphs` to not return a new array.

![capture2](https://cloud.githubusercontent.com/assets/519592/24582370/883e53bc-16e3-11e7-89b3-2e3da5fa285e.PNG)

craftworkgames:
I had a little look at this tonight and I don't think it's going to be simple.

The only way I could think to not return an array from `GetGlyphs` is to change it to an `IEnumerable<>` and use `yield return`. Unfortunately, that doesn't seem to make the memory allocation problems go away, and in fact I think it might make it worse. Did you have some thing else in mind?

As for using `StringBuilder` I think this could be worth it but it looks like it could be quite a deep refactor so I haven't attempted this yet.

I've actually been thinking about this issue on and off since you posted it and I'm not sure if we can fix it by making changes to `BitmapFont` or any of the existing `DrawString` methods. At least, none that make sense anyway (unless you have something else in mind).

However, I have been tossing around one solution which is to introduce a new type. Some kind of `BitmapFontLabel` of sorts.

The idea is to create labels from fonts and hold the "state" in the label object to avoid repeatedly allocating memory. Something like this:

    _label = bitmapFont.CreateLabel("Hello World!");

On creation of the label we could make the calls to `GetGlyphs` and `MeasureString` and store the result. Then we can update *only* when the text changes:

    _label.Text = "Hello New World!";

This would significantly reduce the memory allocations in most cases since the label text tends to only change on game events like loosing health and rarely changes every frame.

Of course, for this to work correctly, we'd also need to provide some new overloads of `DrawString` that takes one of these `BitmapFontLabel`'s and can render it without allocating any memory.

    _spriteBatch.DrawString(_label); 

What do you think?

LithiumToast:
For use with StringBuilder see what I did [here](https://github.com/craftworkgames/MonoGame.Extended/blob/develop/Source/MonoGame.Extended.Graphics/Batcher2D.cs#L210). In most cases just replacing the parameter type `string` with `StringBuilder` works. The code for methods just have to be duplicated for each `string` and `StringBuilder` parameter. 

[...]

Issue #139, opened 2016-03-12
[BitmapFont] Loading BitmapFont at runtime


codecat:
Is this possible? Right now you need to process any `BitmapFont` you want to use through the pipeline, but we'd like to be able to load them at runtime, much like you can load `Texture2D` using `Texture2D.FromStream()`.

LiljebergXYZ:
Definitely a much needed feature :+1:

craftworkgames:
The main downside would be having the core library taking a dependency on `XmlSerializer`. I've tried really hard not to have any extra dependencies in the core library.

That said, it may not actually be an issue in this case. We'll just have to check if XML de-serialization is supported by PCL's without too much pain.

AngeloG:
You mean a dependency on `System.Xml`? That doesn't sound too bad to me.

craftworkgames:
Yeah. It [looks like](https://msdn.microsoft.com/library/gg597391(v=vs.100).aspx) it should be fine. I just wasn't sure before.

Scellow:
Yes please, i want to stay away from that Content annoying thing

craftworkgames:
>  i want to stay away from that Content Pipeline annoying thing

I used to think the Content Pipeline was pretty annoying also but there are some benefits to using it. The most obvious reason is that the content is optimized before loading into your game.

However, one other often overlooked benefit is that your game doesn't have to take any dependencies on libraries that may or may not work on other platforms. This means you can process your content on Windows and be sure that it will work on Android and iOS for example.

So I guess what I'm saying is that we may add support for loading some file formats at runtime, but we'll be very selective. The last thing I want is implementing something in the core library that only works on Windows.

Luckily, portable class libraries enforce this by default.

AngeloG:
My experience of library dependencies has been the other way around - missing dependencies or errors during compile-time of content. This used to not be an issue with XNA, but sadly, Monogame has not implemented the support properly the way that XNA has done it. (A custom project that builds the content as-is, without needing the content pipeline tools, or a content pipeline list of contents. You'd just add the files to the project and it worked. It was pretty good. Right now it's a PITA unless Monogame decides to fix it.)

Scellow:
It can be good when your game is ready to be published, but it's a nightmare when you are prototyping and testing/editing lot of assets..

craftworkgames:
[...]

PR #370, opened 2017-04-05
Added ability to create nine patch regions in a texture atlas.

rds1983:
Also added method that checks whether a region exists in the texture atlas.

I develop UI library for the MonoGame: https://github.com/rds1983/Myra/
Our projects share a lot of common(bitmap fonts, texture atlases, etc). Therefore I decided to move Myra on top of MonoGame.Extended. It would require some adjustments of the MonoGame.Extended. That PR is one of it. I hope it gets through.

### DIFF
--- a/Source/MonoGame.Extended/TextureAtlases/TextureAtlas.cs
+++ b/Source/MonoGame.Extended/TextureAtlases/TextureAtlas.cs
@@ -93,6 +93,27 @@ namespace MonoGame.Extended.TextureAtlases
         }
 
         /// <summary>
+        /// Determines whether the texture atlas contains a region
+        /// </summary>
+        /// <param name="name">Name of the texture region.</param>
+        /// <returns></returns>
+        public bool ContainsRegion(string name)
+        {
+            return _regionMap.ContainsKey(name);
+        }
+
+        /// <summary>
+        /// Internal method for adding region
+        /// </summary>
+        /// <param name="region">Texture region.</param>
+        private void AddRegion(TextureRegion2D region)
+        {
+            var index = _regions.Count;
+            _regions.Add(region);
+            _regionMap.Add(region.Name, index);
+        }
+
+        /// <summary>
         ///     Creates a new texture region and adds it to the list of the <see cref="TextureAtlas" />' regions.
         /// </summary>
         /// <param name="name">Name of the texture region.</param>
@@ -107,9 +128,27 @@ namespace MonoGame.Extended.TextureAtlases
                 throw new InvalidOperationException($"Region {name} already exists in the texture atlas");
 
             var region = new TextureRegion2D(name, Texture, x, y, width, height);
-            var index = _regions.Count;
-            _regions.Add(region);
-            _regionMap.Add(name, index);
+            AddRegion(region);
+            return region;
+        }
+
+        /// <summary>
+        ///     Creates a new nine patch texture region and adds it to the list of the <see cref="TextureAtlas" />' regions.
+        /// </summary>
+        /// <param name="name">Name of the texture region.</param>
+        /// <param name="x">X coordinate of the region's top left corner.</param>
+        /// <param name="y">Y coordinate of the region's top left corner.</param>
+        /// <param name="width">Width of the texture region.</param>
+        /// <param name="height">Height of the texture region.</param>
+        /// <param name="thickness">Thickness of the nine patch region.</param>
+        /// <returns>Created texture region.</returns>
+        public NinePatchRegion2D CreateNinePatchRegion(string name, int x, int y, int width, int height, Thickness thickness)
+        {
+            if (_regionMap.ContainsKey(name))
+                throw new InvalidOperationException($"Region {name} already exists in the texture atlas");
+
+            var region = new NinePatchRegion2D(new TextureRegion2D(name, Texture, x, y, width, height), thickness);
+            AddRegion(region);
             return region;
         }
 


craftworkgames:
The changes seem reasonable but I am curious; What is the use case is for this? Why do to need to add nine patches to the texture atlas? How / when do you get them back out?

Interesting GUI project you've got there. I've been working on our own GUI package in MonoGame.Extended lately. It's always good to have a few choices 👍

rds1983:
Thing is some texture atlases' formats have nine patches. Like LibGDX.
I.e. if you take a look at this file: https://github.com/rds1983/Myra/blob/master/Source/Myra/Resources/default_uiskin.atlas
It has following sprite definition:
border-dark-blue
  rotate: false
  xy: 330, 164
  size: 3, 3
  split: 1, 1, 1, 1
  orig: 3, 3
  offset: 0, 0
  index: -1

'split:' means it is nine-patch.
So if one day you would add LibGDX texture atlases import, it would need to be able to store nine patches.

craftworkgames:
Yeah okay. I'm just wondering if we also need a similar `GetRegion` that returns a `NinePatchRegion2D` or something.

Anyway, I'll merge this now and we can keep going if other ideas come up. Good work :)

rds1983:
Thanks. It would be quite easy to determine whether the returned TextureRegion2D is nine patch using `is` operator.
Though I think it would be cool idea to add virtual method Draw in TextureRegion2D class and override it in NinePatchRegion2D.
So end user can do following:
```c#
var region = textureAtlas.GetRegion("myRegion");
region.Draw(...)
```
And wouldn't need to care whether the retrieved region is nine patch.

craftworkgames:
> Thanks. It would be quite easy to determine whether the returned TextureRegion2D is nine patch using is operator.

Yep. That's true.

Just to see how it might work I just added an overload to `GetRegion` that allows the type of the region returned to be specified. It's backwards compatible with the old way but provides a little more syntax sugar if you know your region types.

            var atlas = new TextureAtlas("atlas", texture);
            atlas.CreateRegion("normal", 100, 100, 100, 100);
            atlas.CreateNinePatchRegion("nine", 0, 0, 100, 100, new Thickness(1));

            var ninePatch = atlas.GetRegion<NinePatchRegion2D>("nine");
            var textureRegion = atlas.GetRegion("normal");

> Though I think it would be cool idea to add virtual method Draw in TextureRegion2D class and override it in NinePatchRegion2D.

Actually we have something like that already in the [`TextureAtlasExtensions`](https://github.com/craftworkgames/MonoGame.Extended/blob/develop/Source/MonoGame.Extended/TextureAtlases/TextureAtlasExtensions.cs).

        public static void Draw(this SpriteBatch spriteBatch, TextureRegion2D textureRegion, Rectangle destinationRectangle, Color color, Rectangle? clippingRectangle = null)
        {
            var ninePatchRegion = textureRegion as NinePatchRegion2D;

            if (ninePatchRegion != null)
                Draw(spriteBatch, ninePatchRegion, destinationRectangle, color, clippingRectangle);
            else
                Draw(spriteBatch, textureRegion.Texture, textureRegion.Bounds, destinationRectangle, color, clippingRectangle);
        }

This method handles the situation when you've got a `NinePatchRegion2D` stored as a `TextureRegion2D`. I use this extensively in the GUI package and it works really well.

> region.Draw(...)

We've had problems coupling rendering code to the model code so having `Draw` methods on texture regions is something we want to avoid.

rds1983:
I see. Well, according to [this](http://stackoverflow.com/questions/686412/c-sharp-is-operator-performance) using `as` has almost same performance as using enum for checking type.

I've another question. I've noticed that your implementation of BitmapFont lacks kernings. I would want to fix it. Would you accept such PR?

craftworkgames:
> I've another question. I've noticed that your implementation of BitmapFont lacks kernings. I would want to fix it. Would you accept such PR?

Sure I guess. `BitmapFont` currently renders fonts *exactly* the same as `SpriteFont` (at least for the fonts I've tested) so just be careful to keep that in mind. I'm not sure if any of our current fonts have kerning data in the `.fnt` file. I'm curious if it would change the way they currently look?

If you're going to make any significant changes like this though, I'd like to keep the demos up-to-date with each new feature we add. So the bitmap font demo should show off kerning somehow.

I'm currently in the middle of refactoring the demos quite a bit but I'm sure we can co-ordinate our efforts.

rds1983:
>  I'm not sure if any of our current fonts have kerning data in the .fnt file. I'm curious if it would change the way they currently look?
This one does have kernings: https://github.com/craftworkgames/MonoGame.Extended/blob/develop/Source/Demos/Demo.Features/Content/Fonts/impact-32.fnt
So some fonts look would definitely change their appearance.

How about temporarly adding a boolean field UseKernings(initially set to false) to the BitmapFont? Then eventually that field will be removed and kerning will be rendered.

While there, I'd suggest to review [this](https://github.com/craftworkgames/MonoGame.Extended/blob/develop/Source/MonoGame.Extended/BitmapFonts/BitmapFontExtensions.cs)
As it seems to me that every DrawString call does allocations(GetGlyphs). Probably the only way to fix it is to copy/paste GetGlyphs code to DrawString. But I think that worth it.

LithiumToast:
@rds1983 See #365

rds1983:
@LithiumToast good to know, thanks.

RUSshy:
@rds1983 I like your lib because you load raw assets, if you move to MonoGame.Extended i'll loose that feature :'( (for modding)

rds1983:
@RUSshy  not really. I am gonna keep my system of asset management.
As I also have medically confirmed allergy to the MonoGame Content Pipeline. :)

Actually, considering #139, it may go other way around, and I may move my code of loading FNT files(and also of LibGDX texture atlases) in the run time to the MonoGame.Extended. Of course, if MonoGame.Extended team won't mind.

What I'll lose is FNA support. However so far I haven't heard of anyone using my lib for the FNA.

PR #372, opened 2017-04-06
Added kernings support to the BitmapFont.

### LINKED ISSUE #371: Add kernings to BitmapFont
This ticket is continuation of our discussion here #370.

rds1983:
Fixes #371 
Following our discussion I've added static property UseKernings(initially set to false) to Bitmap that controls whether kernings are applied during the rendering.
Also I've slightly updated the StarWarrior demo. Now you can turn on/off that property by pressing 'K' key. The difference is visible only in Debug mode(where additional text is drawn).


craftworkgames:
That looks good mate.

I decided to turn on kernings by default but we'll leave the flag there just in case someone wants to turn it off.

To be honest, it's such a subtle change I doubt anyone will even notice unless they're paying very close attention.

rds1983:
Well, frankly, I didn't notice any different either. :)
On the other hand, I know there are folks with designer taste who scowl every time when they see unkerned text.

craftworkgames:
I guess it depends on the font used. Some fonts would probably look a lot better.

PR #376, opened 2017-04-11
Add new ViewportAdapter : ScalingCenteredViewportAdapter

Lysfith:
This PR add a new viewport adapter : ScalingCenteredViewportAdapter.

This viewport adapter allows to scale content and keep the ratio of virtual size, independently of the window size.

Some examples:
example_1
example_2
example_3

### DIFF
--- a/Source/MonoGame.Extended/MonoGame.Extended.csproj
+++ b/Source/MonoGame.Extended/MonoGame.Extended.csproj
@@ -115,6 +115,7 @@
     <Compile Include="Transform.cs" />
     <Compile Include="ViewportAdapters\BoxingViewportAdapter.cs" />
     <Compile Include="ViewportAdapters\DefaultViewportAdapter.cs" />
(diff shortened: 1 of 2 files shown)


craftworkgames:
@Lysfith Hey, sorry I missed this PR.

How is this different from the Boxing Viewport Adapter?

Lysfith:
Hello,

In fact, it's not different, why i don't see it before :)

Sorry.

craftworkgames:
No worries.

There is actually a viewport adapter improvement we haven't got if you're interested?  I'd like to have an option on the boxing adapter to keep the ratio pixel perfect. Do you know what I mean?

So like, currently if you scale up a low resolution like 100x100 to 250x250 you get a ratio like 2.5x2.5 which stretches the pixels. It often looks better if the ratio is exactly 2x2 with black bars on all 4 sides.

Issue #377, opened 2017-04-14
[Tiled] TiledMapTilesetTile might need objects


musbah:
Hello, it has been a while since I've last looked at extended and the new changes seem amazing. Anyway I've been going through the Tiled section and I noticed that there might be something that I could do before that I can't anymore. 

I did my best to go through everything and make sure it didn't exist, but I could be wrong and if I am then I apologize. 

Anyway, I've was working with the collision editor of tiled on tileset tiles and it doesn't seem like I can get them anymore using extended. Let me try to explain it better, I know that I could just use a Tiled Object layer and add my polygons on there. But there is a feature in tiled that lets you add an object layer to every single tileset tile. 

So basically instead of using the map object layer and adding the same polygons over the tiles that use the same tileset over and over again, I could just add it to the tileset tile once and the code will be able to "multiply" it on the map. This is something that I used to be able to do with extended but not anymore (at least I think so).

Anyway I don't think it would be hard to add it myself, I was just wondering if I am correct in assuming that it can't be done right now. I've gone through all of the tiled files and it doesn't seem to be possible at the moment.

LithiumToast:
If you attach an example of a `.tmx` map that has the missing functionality we can take a closer look.

musbah:
I'm basically talking about the object groups inside of the tile in a tileset, there's basically no way of getting the information (at least I can't find it). 

An example would be the polyline inside of that group object:
`<tileset firstgid="1" name="tileset1" tilewidth="64" tileheight="64" tilecount="260" columns="20">
    <tile id="201">
   		<objectgroup draworder="index">
    		<object id="2" x="10" y="63">
     			<polyline points="0,0 5.25,-7 6.75,-18.5 13,-27.5 21,-34.5 31.5,-26.75 36.75,-18 37.25,-7.75 42.5,0.25"/>
    		</object>
   		</objectgroup>
  	</tile>
</tileset>`

(sorry tried to format it, but the github code thing isn't working)

PR #378, opened 2017-04-15
Added the ability to get tileset tile objects.

musbah:
Basically, I added the ability to get the objects of every tileset tile. (Tiled Collision editor for the tileset).
For more details look at issue [#377](https://github.com/craftworkgames/MonoGame.Extended/issues/377)

I tested it pretty extensively since I'm using it myself. Everything seems to be working correctly.

By the way, do you think I should add an `if(objectCount > 0)` to `TiledMapTilesetTile.cs` ? I don't think it needs one, since it would just be 0, but I guess I could be wrong. Adding it really seems useless to me, but what do you guys think?


craftworkgames:
Thanks @musbah

PR #379, opened 2017-04-20
Fix #365

### LINKED ISSUE #365: BitmapFont memory usage.

LithiumToast:
See #365

### DIFF
--- a/Source/Demos/Demo.Features/Demos/InputListenersDemo.cs
+++ b/Source/Demos/Demo.Features/Demos/InputListenersDemo.cs
@@ -122,10 +122,10 @@ namespace Demo.Features.Demos
             }
 
             var textInputY = 14 * _bitmapFont.LineHeight - 2;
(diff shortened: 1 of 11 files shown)


craftworkgames:
Very impressive implementation and use of enumerators.

Do you think we can reduce the amount of code duplication between `StringGlyphEnumerator` and `StringBuilderGlyphEnumerator`.

LithiumToast:
Not really. `StringBuilder` and `String` are different.

craftworkgames:
Merged.

> Not really. StringBuilder and String are different.

I'm absolutely certain there's a way to reduce the code duplication. Obviously it's not super important but I think it's worth doing at some point. I'll add it to my to-do list.

Issue #352, opened 2017-03-12
[Tiled] Blank Tiles are Ignored and TiledMapTileLayer.TryGetTile() Returns Wrong Tile


JayJeckel:
The TiledMapProcessor.CreateTilesIn*Order() methods ignore blank tiles, ie those with global identifiers of 0, when created the TiledMapTileLayer.Tiles arrays and, in any case where there are blank tiles on a tile layer, that results in the tiles being stored out of index order in arrays that are shorter than expected. As a result this causes TryGetTile() to return null when the given x, y coords do have a tile and to return a tile when the given coords to not have a tile.

For example. I have a map with a width/height of 10/10 and it has two tile layers. The first layer is filled by a tile with gid of 1 and the second layer is empty except for the top row and bottom right corner which each have gid tile 2.

The first issue is that if I call TryGetTile() with the x=0 and y=1 (the first tile in the second row which doesn't have a tile), I will get back the tile for the bottom right corner, ie x=9, y=9, and gid=2. The second issue is the reverse, if I call TryGetTile() with x=9 and y=9, I will get null instead of the actual tile.

The first issue appears to have an easy fix, just add a check to verify that the x and y of the tile being returned matches the x and y that was passed in and return null if that isn't the case. Though this is unnecessary if the fix to the second issue covered below is used.

The second issue seems to be fundamental to how the system is implemented, in that blank tiles are not added to the Tiles array at all. This may be a great idea to save space for products that have many extremely large maps, but as something that is applied out of the gate it seems like a touch of premature optimization. My solution was to remove the 'if (tileLayerData[dataIndex].GlobalIdentifier == 0) continue;' line from each of the four TiledMapProcessor.CreateTilesIn*Order() methods. This caused pretty much all of the demo projects to fail, but I traced it back to the TiledMapProcessor.CreateTileLayerModels() method and adding 'if (tile.IsBlank) { continue; }' as the first line in that method's foreach loop seemed to fix those errors and everything compiled correctly.

I am fairly new to MonoGame.Extended, so if there is some other reason that blank tiles were not being stored in the layer's Tiles array, please let me know, but to me being able to accurately check the correct tiles on all tile layers seems more useful than dumping blank tiles and complicating the ability to check them.

LithiumToast:
[...]

PR #380, opened 2017-04-23
Fix TryGetTile

JayJeckel:
Change TiledMapTileLayer to store blank tiles so that TryGetTile returns the correct TiledMapTile instances (#352).


craftworkgames:
Looks good to me.

Thanks 👍

Issue #381, opened 2017-04-27
[Particles] Triggered/drawn at LayerDepth


toddgibson:
I think it would be wonderful to have an overload method to Trigger that allows for specifying layer depth for drawing...

public void Trigger(Vector2 position, float layerDepth);

I have need of this in my current project. I may make an attempt to implement it, but I've not delved into the guts of your wonderful project thus far. :)

craftworkgames:
This seems like a good idea. I haven't looked into what's involved but I assume it would be relatively straightforward.

I've got quite a lot on my plate at the moment but perhaps someone else what's to submit a PR?

toddgibson:
I actually got this implemented and working last night, it was pretty easily accomplished.  I'll submit a PR for it tonight.

PR #382, opened 2017-04-27
add LayerDepth to Particles

toddgibson:
This work is in regards to Issue #381

### DIFF
--- a/Source/MonoGame.Extended.Particles/Particle.cs
+++ b/Source/MonoGame.Extended.Particles/Particle.cs
@@ -16,6 +16,7 @@ namespace MonoGame.Extended.Particles
         public Vector2 Scale;
         public float Rotation;
         public float Mass;
+        public float LayerDepth;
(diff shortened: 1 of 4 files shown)


craftworkgames:
I've just had a quick glance over this on my phone. It looks good so far but I think there's a small change to make.

All public methods should default the layerDepth to zero. This way it's a non breaking change.

toddgibson:
It should already be non-breaking since I added overloaded methods.  This way if you don't specify layerDepth it will work as normal (by using the existing methods).  But I can still make the change if you want.

craftworkgames:
Yeah, I think it's worth doing just to reduce the code duplication.  

I believe you can just add another commit to this PR to make the changes. If you don't get around to it by tonight I don't mind just fixing it up myself. Maybe I'll try github's online editor.

craftworkgames:
Great thanks :)

PR #384, opened 2017-04-29
DirectX feature demos & shader fix

LithiumToast:
Silly mistake in the shader `.fx` file caused this problem on the community forums: 

http://community.monogame.net/t/having-a-problem-with-drawing-a-tiled-map/9081/11

Along with the fix I added a DirectX version of the feature demos.


craftworkgames:
There's something weird going on with the project changes you made.

When I first ran it I got null reference exception. I figured it had something to do with references so I checked and the DX project didn't appear to be referencing MonoGame DX as you'd expect.

I tried fixing it up with NuGet Package Manager in Visual Studio but it's acting weird. When I change the references in either of the `WindowsOpenGL` or `WindowsDirectX` projects, they both end up with the *same* references.

Did you make a copy of the project maybe? Perhaps it's got a duplicate GUID in the project file or something?

LithiumToast:
Oh, it probably has duplicate IDs.

LithiumToast:
Ok GUIDs are fixed and references are fixed.

craftworkgames:
Much better. Thanks.

PR #385, opened 2017-04-29
Fix nested transform WorldMatrix calculation

johngirvin:
Fixes issues discussed in forum thread:
http://community.monogame.net/t/entities-nested-using-transformcomponent2d-parent/9084

Correctly calculates the WorldMatrix for Transform and TransformComponent2D that have parents (any depth) who are rotated and/or scaled.

### DIFF
--- a/Source/MonoGame.Extended.Entities/Entities/TransformComponent2D.cs
+++ b/Source/MonoGame.Extended.Entities/Entities/TransformComponent2D.cs
@@ -4,7 +4,7 @@ using Microsoft.Xna.Framework;
 namespace MonoGame.Extended.Entities
 {
     [EntityComponent]
-    public class TransformComponent2D : EntityComponent
+    public class TransformComponent2D : EntityComponent, IMovable, IRotatable, IScalable
     {
         private TransformFlags _flags = TransformFlags.All; // dirty flags, set all dirty flags when created
         private Matrix2D _localMatrix; // model space to local space
@@ -180,8 +180,8 @@ namespace MonoGame.Extended.Entities
         {
             matrix = Matrix2D.CreateScale(_scale) *
                      Matrix2D.CreateRotationZ(_rotation) *
-                         Matrix2D.CreateTranslation(_position);
-            }
+                     Matrix2D.CreateTranslation(_position);
+        }
 
         [Flags]
         private enum TransformFlags : byte
--- a/Source/MonoGame.Extended.Entities/Entities/TransformComponent2D.cs
+++ b/Source/MonoGame.Extended.Entities/Entities/TransformComponent2D.cs
@@ -148,6 +148,7 @@ namespace MonoGame.Extended.Entities
 
             RecalculateLocalMatrixIfNecessary();
             RecalculateWorldMatrix(ref _localMatrix, out _worldMatrix);
+
             _flags &= ~TransformFlags.WorldMatrixIsDirty;
         }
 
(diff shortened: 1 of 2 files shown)


craftworkgames:
I just took a quick look at this and I noticed the wheels are rotating backwards in the Scene Graphs demo.

rotating-backwards

Could you please look into this?

I'm not too concerned where the fix is applied, as long as it's explainable. If we need to update the demo that's fine, I just wanted to bring it to your attention in case it's a bug in the code.

As with all features, we need some way to prove that they are working and continue working. Currently the demos are our best tests.

Thanks.

LithiumToast:
The "fix" was described in the forums post.

```
```
                    var texture = ...;
                    var sourceRectangle = ...;
                    var color = ...;
                    var position = WorldPosition;
                    var rotation = -WorldRotation;
                    var scale = WorldScale;
                    var origin = ....;

                    spriteBatch.Draw(texture, position, sourceRectangle, color, rotation, origin,
                        scale);
```

See http://stackoverflow.com/questions/3527531/what-is-this-rotation-behavior-in-xna

**Comment by craftworkgames** [comment]
> The "fix" was described in the forums post.

What code is that? Where does it come from?  Is it from `SceneNode.cs`?

I'm fine if the code needs to change, I'm just wondering why it isn't included in the PR?

**Comment by LithiumToast** [comment]
It's from the discussion on the forums...

**Comment by craftworkgames** [comment]
> It's from the discussion on the forums...

Yes, I'm aware of the discussion.

**Comment by LithiumToast** [comment]
It's just pseudo-code that demonstrates what works with `SpriteBatch` when using `Transform2D` or `TransformComponent2D`.

**Comment by craftworkgames** [comment]
I'm not exactly sure what you're expecting me to do?  Do you want me to fix up the demo?  

> SceneNode's rendering is also weird... It tries to do some of what Transform2D is doing.

What should we do about that? Do we need to fix up the `SceneNode` class or do you think we should get rid of the whole Scene Graphs package entirely?

My point is, this PR is not really complete. Once we'd sorted out these problems (or at least figured out the next step) I'll merge it.

**Comment by johngirvin** [comment]
Changing SceneNode.cs : 76 to `spriteBatch.Draw(..., -rotation, ...);` makes the wheels rotate correctly in the demo. If this is an acceptable fix I'll add a commit for it.

Otherwise, we need to get to the bottom of why using a negated rotation angle in Transform/TransformComponent2D instead of in the Draw call as above gives strange results.

Child nodes rotate in the opposite direction to their parents as shown below (_carNode has a rotation of +15deg applied):
```
protected internal override void RecalculateLocalMatrix(out Matrix2D matrix)
{
    matrix = Matrix2D.CreateScale(_scale) *
             Matrix2D.CreateRotationZ(-_rotation) *
             Matrix2D.CreateTranslation(_position);
}

spriteBatch.Draw(texture, position, sourceRectangle, drawable.Color, rotation, drawable.Origin, scale, drawable.Effect, 0);
```
image

```
protected internal override void RecalculateLocalMatrix(out Matrix2D matrix)
{
    matrix = Matrix2D.CreateScale(_scale) *
             Matrix2D.CreateRotationZ(_rotation) *
             Matrix2D.CreateTranslation(_position);
}

spriteBatch.Draw(texture, position, sourceRectangle, drawable.Color, -rotation, drawable.Origin, scale, drawable.Effect, 0);
```
image

craftworkgames:
> Changing SceneNode.cs : 76 to spriteBatch.Draw(..., -rotation, ...); makes the wheels rotate correctly in the demo. If this is an acceptable fix I'll add a commit for it.

Yeah, I guess this is good enough. I don't really understand it but chances are we'll get rid of the SceneGraphs package anyway once we have a replacement demo using transforms instead.

Issue #383, opened 2017-04-28
Support UWP / Windows 10


Odonno:
I wanted to use the Monogame.Extended package in my project but I can't install the NuGet package.

![image](https://cloud.githubusercontent.com/assets/6053067/25547510/bbabfb84-2c67-11e7-95b4-22cdfc64f21f.png)

craftworkgames:
yeah, we currently don't officially support UWP but it should be possible to get it working if you build from source. 

It'd be nice to support it if we can but I'm not sure it's possible without breaking other platforms in the process. Maybe there is a way to do it, I'm not sure. The hard part is testing all the other platforms to make sure they still work too.

If you do get it working I'd love to know about it.

claudiuslollarius:
I am facing the same issue and think the issue is caused by how you create your nugget packages.

In the nuspec files, I see lines like:
`<file src="..\MonoGame.Extended.Tweening\bin\Release\MonoGame.Extended.Tweening.dll" target="lib" />

Following [the convention for multi-targeting nuget packages](https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks), by changing to something like:

`<file src="..\MonoGame.Extended.Tweening\bin\Release\MonoGame.Extended.Tweening.dll" target="lib\portable45-net45+win8+wpa81" />

should fix this.

claudiuslollarius:
Created pull request #387 to fix issue. No actual code needed to be changed and should not impact functionality on other platforms.

LithiumToast:
Perhaps we should use .NET standard now.

> Warning
PCLs are not recommended
Although PCLs are supported, package authors should support netstandard instead. The .NET Platform Standard is an evolution of PCLs and represents binary portability across platforms using a single moniker that isn't tied to a static like like portable-a+b+c monikers.

craftworkgames:
Yep. We definitely should got to .NET standard.

It's just not something I've really looked into whole lot. The last time I checked I was under the impression that it wouldn't work because MonoGame would also need to be .NET standard (which I figured wasn't going to happen anytime soon). [Maybe?](https://github.com/MonoGame/MonoGame/issues/5339)

In a more [recent conversation](https://github.com/VelcroPhysics/VelcroPhysics/issues/3) I had with @Genbox (Velcro Physics formerly Farseer) I realized that .NET standard is actually PCL compatible, so it might not be as difficult as I first thought.

[...]

PR #387, opened 2017-05-08
Fixed nuspec definitions

claudiuslollarius:
Modified nuspec definitions to conform to [nuspec convention to target multiple platforms](https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks) as described in issue #383

### DIFF
--- a/Source/NuGet/MonoGame.Extended.Animations.nuspec
+++ b/Source/NuGet/MonoGame.Extended.Animations.nuspec
@@ -28,7 +28,7 @@
     </dependencies>
   </metadata>
   <files>
-    <file src="..\MonoGame.Extended.Animations\bin\Release\MonoGame.Extended.Animations.dll" target="lib\portable45-net45+win8+wpa81" />
+    <file src="..\MonoGame.Extended.Animations\bin\Release\MonoGame.Extended.Animations.dll" target="lib\portable-net45+win8+wpa81" />
     <!-- A readme.txt will be displayed when the package is installed -->
     <file src="readme.txt" target="" />
   </files>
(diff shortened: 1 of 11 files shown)


LithiumToast:
@claudiuslollarius Shouldn't the target be `portable-net45+win8+wpa81` instead of `portable45-net45+win8+wpa81`?

claudiuslollarius:
Not so sure: look inside:
```MonoGame.Extended\Source\MonoGame.Extended\packages.config
and you'll see even Newtonsoft JSON and MonoGame itself use
```portable45-net45+win8+wpa81
as platform moniker

LithiumToast:
Hmm. https://docs.microsoft.com/en-us/nuget/schema/target-frameworks says otherwise.

claudiuslollarius:
@LithiumToast Ok, updated the monikers to conform to the document you linked.

craftworkgames:
Great work. My only concern is that we inadvertently break another platform. I mean, it should work but you don't really know until it's tested.

I can test Android but I don't have an iOS or Linux rig setup.

craftworkgames:
Okay, after doing a little reading I'm pretty convinced this will be fine. 

I'll merge now and mention it on the gitter channel. If we have anyone complain we'll deal with that then.

claudiuslollarius:
@craftworkgames I see you merged the pull request.

Could you please rebuild and publish updated NuGet packages? I would really need to use your library in my project.

Thank you in advance

craftworkgames:
@claudiuslollarius Do you have a problem using the [pre-release packages](https://github.com/craftworkgames/MonoGame.Extended/wiki/How-to-use-the-pre-release-NuGet-packages)?

I wasn't planning on doing an official NuGet release for quite some time yet.

PR #391, opened 2017-05-23
[Docs] Packages root page, overview of packages.

robert-wallis:
I'm planning on adding sub-pages under this root `packages` category for each library.  Like the `User Guide` tab at the top of [mkdocs](http://www.mkdocs.org/user-guide/writing-your-docs/).

This is mostly just an overview of what is in each package.  Like the README.md.


robert-wallis:
Turns out that's not how mkdocs works, there are no root-pages.  So I'm going to move this to the index.md

robert-wallis:
Latest Commit f3198fd contains the BitmapFonts page from the wiki and the images that are missing.  Checked in images because, they are missing already from the wiki, and they aren't that big.

craftworkgames:
Great work mate. I've pushed those changes [live](http://craftworkgames.github.io/MonoGame.Extended/)

Ideally I'd like to get the existing wiki migrated over and then we can update the main readme to show everyone the new docs site.

Oh and it's probably a good idea to have a page explaining how to contribute to the docs ;)

PR #396, opened 2017-05-25
[Content] removing unused ScratchBuffer

robert-wallis:
This scratch buffer is `internal`, it has a 1mb limit, and it's unused.

I think if devs want a scratch buffer, maybe there should be a `ScratchBuffer` class with a static internal byte array.

Recompiled whole solution successfully with it removed.


craftworkgames:
Yep. I agree, it looks like an oversight. I hadn't even noticed this code existed tbh.

PR #403, opened 2017-06-08
Added some command lines to run mkdocs

robert-wallis:
I forgot myself, so I had to look it up on the mkdocs site.  I think it's worth putting in here just to remind that you run it from the MonoGame.Extended folder, and that all pages must be in `mkdocs.yml`

### DIFF
--- a/Docs/Community-and-Support.md
+++ b/Docs/Community-and-Support.md
@@ -19,6 +19,7 @@ Note: When you ask a question, please consider how permanent the answer should b
  - A rough draft is better than none.
  - If we all wrote a little documentation, it adds up to a lot.
 
+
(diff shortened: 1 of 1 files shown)


craftworkgames:
Great.

It might even be worth putting something on the home page of the docs site about contributing to the docs.

robert-wallis:
We could move the Docs section to the bottom of the index page.  Or mabye move all the support to the index page?

craftworkgames:
Yeah, tbh I'm not really sure what's best. I'm thinking maybe split up "Community and Support" and "How to Contribute" and add links to both under the "Getting Started" section.

If you get to it before me you're welcome to do whatever you think feels right.

PR #404, opened 2017-06-08
Formatting code blocks with csharp

robert-wallis:
because C# randomly fails to format correctly.


craftworkgames:
Nice work. Good to see you back on this 😄 

When I saw PR 404 my first thought was.. err.. page not found.

robert-wallis:
Ha ha, so did I :)

PR #422, opened 2017-09-14
Use sprite WorldMatrix for drawing

johngirvin:
Sprite's parent transform is ignored when drawing via SpriteBatch extension method.


craftworkgames:
Hey, nice work. Sorry, I almost missed this one.

PR #427, opened 2017-09-22
Created a new 2.5D camera

Apostolique:
Instead of controlling a zoom value directly, we control the zoom through a Z axis value. The camera moves along the Z axis to get closer or further away from 2D plane layers.

2.5D Camera

Parallax and Parallax zooming is handled automatically based on a layer's zPosition.

Except for the zooming, the Camera25D is used exactly like a regular Camera2D. You can draw a layer by calling GetViewMatrix(). To add parallax, you can pull or push a layer with GetViewMatrix(-5) or GetViewMatrix(5). Negative values push a layer further in the distance. Positive values pull a layer towards the camera (or behind it).

There's one big difference though, before drawing a layer, call IsVisible(0) with the layer's "parallax" value. It returns a bool. A layer is not visible if the camera is "below" it.


craftworkgames:
Hmm.. I'm not sure about this.

Maybe there's a way to achieve the same kind of thing without creating a whole new camera class?  It seems like a lot of repeated code. Maybe extending the existing camera or providing some sort of camera controller maybe?

I haven't had a chance to put some proper thought into it but at first glace it doesn't feel right.

ronelm2000:
This would be a good add if there was a way to tilt the camera and flip 2D assets, but it honestly looks more like just an extension of the Zoom functionality. (Parallax is always nice tho)

Still, if the intention is there... it could use a few more refactoring...

Apostolique:
Indeed, I can most likely add the functionality by extending the existing camera class and remove all the duplicated code.

What do you mean by flip 2D assets? @ronelm2000

ronelm2000:
Since you're using 2D planes on a 3D camera, Sprites should be able to be rotated in the X/Z and Y/Z plane. (just like you can right now in the X/Y plane). You can then be able to Animate it (with a custom Animation class returning float).

I can do a separate merge request for this myself, but not after this one. So I want to know if the 2.5D Camera is just limited to accessing the Z-axis...

Apostolique:
There's no limit to what we can do. Right now though, the camera code is based on the same orthographic camera as the 2D camera. To have rotations around the X and Y axis (right now we only have a rotation around the Z axis since rotations around the Z axis don't change how close something is from the camera), the camera would need to be switched to an actual perspective camera. For a 2.5D game, I feel like having an orthographic camera is best, but maybe I'm wrong about that.

craftworkgames:
I just wanted to let you guys know I haven't forgot about this PR.

Tbh, I'm not really sure if we should pull it in as-is or not. I'm fairly sure we'll use it in some form or another in the future, but only after some careful thinking of what we actually want. Thinking that I haven't found the time to put a lot of thought into 😩 

Anyways, if someone is really keen to get this in I'll merge it, otherwise, I'll just leave it open for discussion.

Apostolique:
I think it's fine to wait. In the current state, this Camera25D is separate from the Camera2D. If we wanted to make a common camera interface to reduce duplicated code, we'd have to put some thought into it (It's not right away obvious what a camera should be.).

Personally I'm already using the Camera25D in my project though.

raizam:
I'm using this implementation as well.
Maybe rather than having 2 cameras classes, this behavior should be merged to the current Camera2D class, and rather than changing the Zoom behavior, a Z position may be added

craftworkgames:
> Maybe rather than having 2 cameras classes, this behavior should be merged to the current Camera2D class.

Yep, that's the plan. 

The only question is whether I should merge this pull request as-is and refactor it later I leave it open until I've got the time to do it properly.

Lately, I've been experimenting with a new branching strategy. The idea is to take (most) pull requests into the `develop` branch even if they are a little rough and at some point later review / refactor the code 
and move it into another branch like `2.0`.

This way, the `develop` branch becomes a place for experimental work in progress stuff (but people can still use it if they want) and we have more stable semantically versioned branches that become proper released packages.

What do you guys think?

raizam:
>  What do you guys think?

I think it's a good thing that the develop branch includes experimental features, and keep the stable branches separated.

In this current case, this is a desired feature but needs additionnal work to get it merged, if you don't have much time available, you should request additionnal changes to @Apostolique

Apostolique:
I can change the code if you want.

craftworkgames:
> I can change the code if you want.

As discussed a few times, it feels like it would be better if we merged these changes into a single camera class (or maybe have a base class that different types of camera's derive from).

I don't have time to do this work right now. So yeah, if you want to do it then go right ahead. Thank you.

> I think it's a good thing that the develop branch includes experimental features, and keep the stable branches separated.

Of course, these decisions are not always so "cut and dry". I've taken 203 pull requests into this project since 2015 and while I appreciate all of the work put into every single one of them, it's not always easy to decide what should go into the library and what shouldn't.

In my experience, when someone makes a pull request I've essentially got the following options:

1. Accept it the way it is (this works great for small changes and bug fixes)
2. Ask the contributor to make some changes (communicating exactly what changes should be made can be quite difficult)
3. Take the pull request and refactor the code (this can be quite time consuming, especially if I need to spend a lot of time just understanding what it does)
4. Take the pull request and let it "evolve" naturally (in other words, defer the refactor until someone complains about it or I end up using that feature myself)

A lot of the time, especially with larger pull requests I've struggled to even know what to do with them. Sometimes I can see obvious problems with the design (and later those same problems turn out to be the ones people complain about) but it can be so time consuming to fix that I can't get to it for months. On the other hand, many of those larger pull requests have been really useful and people really want them in the library.

I've tried all of the above with varying degrees of success. I still struggle with it. 

I'm sure there's things I could be doing better. Knowing what those things are and actually spending the time to do them is also challenging.

So yeah, I've been tossing around the idea of an "experimental" branch approach for a while now. I think it'll help. I'm not exactly sure how to do it yet but I have started playing around in a branch called `2.0` with some new ideas for the library.

Apostolique:
Okay, so I think it's possible for my Camera25D to be merged into the Camera2D. The difference between the two is the way the "zooming" is done.

In the Camera2D, the zoom is like a camera lens. You make everything bigger, but you don't actually move in the 3D space. In other words layers (objects) in the distance stay the same size. Example: https://www.youtube.com/watch?v=JUOF8kEVokc

In the Camera25D the concept of zooming is changed to dollying. In this, the camera actually does move in the 3D space. Example: https://www.youtube.com/watch?v=m3uJlyhXRJA

In other words, the two systems should be able to coexist in a single camera. A side effect of this new camera would be the ability to do the vertigo effect (Push pull): https://www.youtube.com/watch?v=0wbwRQ8AXik
_____
I propose that we remove the concept of parallaxFactor and let the dolly handle that part.

A side effect would be that placing objects further away would make them appear smaller. This might not be the right behavior for 2D games where the art is drawn at the right scale from the start and they only want parallax.

Either we force people to scale their objects themselves in their game world, or I provide a GetViewMatrix function that ignores the dolly "zoom" value (Essentially it does the rescaling on the view matrix behind the scenes.).

Hopefully that makes sense.

raizam:
I think it should be kept very simple, here is what I would do:

1. Keep the current Zoom implementation for legacy.
2. Add a ZPosition property

I've removed the DollyIn/out methods as these are sugar method that really are not useful

Apostolique:
I have merged the two cameras. Some functions should get new names as they are now misleading, namely ```GetZPositionFromZoom()``` and ```GetZoomFromZPosition()```.

craftworkgames:
> I have merged the two cameras.

Great work. Looks like some of the demos are not compiling but that should be easy to fix.

> A side effect would be that placing objects further away would make them appear smaller. This might not be the right behavior for 2D games where the art is drawn at the right scale from the start and they only want parallax.

This is definitely not what we want. I think what we are talking about here is the difference between [Orthographic](https://en.wikipedia.org/wiki/Orthographic_projection) and [Perspective](https://en.wikipedia.org/wiki/3D_projection#Perspective_projection) projection. In a lot of the game frameworks I've used they typically split the camera class into separate [OrthographicCamera and PerspectiveCamera classes with a common base class](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/graphics/Camera.html). 

I've actually been thinking for a while (I'm not the first person to mention it either) that maybe we should follow this convention and split the camera up this way. We could make `Camera2D` obsolete and introduce the 2 new camera types (with common base) in a clean way.

Based on the branching model I've been trying I think the way this would work is to:
1. Take everything we currently have in MonoGame.Extended and put it in a branch called `1.0`
2. Pull these changes into `develop` and refactor the camera classes the way we really want them.
3. The new design becomes branch `2.0` along with a bunch of other things I'd like to change.

Apostolique:
What I have right now is still Orthographic with fake perspective. It could make sense to have the option to set the right scale for a layer. Essentially, choose the distance for a layer to be 1:1 with the asset scale when the camera is at depth 1.

For example, let's say a layer is at depth -10. Stuff drawn at that scale would look pretty small. A parameter could be passed to scale the layer back up. It turns out the math for something at depth -10 to be 1:1 when the camera is located at 1 needs to be scaled up by a factor of 11. (The math is ```factor = -depth + 1```)

Example code if you were to do that manually:

    //Here we have a texture 't'.
    float zPan = -10;
    float scale = 11;
    Matrix superTransformMatrix = cam.GetViewMatrix(zPan);

    s.Begin(transformMatrix: superTransformMatrix);
    s.Draw(t, new Rectangle(0, 0, (int)(t.Width * scale), (int)(t.Height * scale)), new Color(255, 255, 255));
    s.End();

Essentially, we could provide a GetViewMatrix that does that part automatically.

craftworkgames:
I'm going to close this (too far behind and will be a merge hell).

I'm planning on re-doing the camera system in the 2.0 branch. When I get to it I'll take these changes and everything we talked about into account.

Apostolique:
Sounds good! If you have questions about specific parts of the code, I can help.

craftworkgames:
@Apostolique hey mate.

We spoke this morning on your stream. Does this PR have the latest version of your camera class?

It think the best approach is that I grab a copy of the latest version and take a closer look. I'll consider if it should go into the `develop` branch or the `2.0` branch. (ideally we only want to make breaking changes to the API in each major release).

After a quick eyeball over the code, I think maybe their might be a bug? For example in these 2 methods you changed it from returning a `Vector2.One` to `0`. Maybe I'm missed something but shouldn't these methods return `1`?

```
         private Matrix GetVirtualViewMatrix()
         {
-            return GetVirtualViewMatrix(Vector2.One);
+            return GetVirtualViewMatrix(0);
         }
 
         public Matrix GetViewMatrix()
         {
-            return GetViewMatrix(Vector2.One);
+            return GetViewMatrix(0);
         }
```

More importantly, to properly evaluate the code I'm going to need to see it in the context of some sort of demo. I don't mind setting this up but maybe you could give me some guidance on how to do the new parallax stuff.

Lastly, the name of the camera class has bothered me for a while. I'm planning on renaming `Camera2D` to `OrthographicCamera` and maybe introducing a new `PerspectiveCamera` in the `2.0` branch. This will be more consistent with [other libraries](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/graphics/Camera.html). Although, this is not set in stone. I've seen other libraries have a `Camera` class with a `Projection` property.

If you have any other thoughts let's discuss :)

LithiumToast:
@craftworkgames The projection matrix should definitely be handled by the camera if you are going to rename the classes as the name implies the projection.

Apostolique:
@craftworkgames The code in this pull request is outdated. Here is newer code: https://gist.github.com/Apostolique/475b3827fd3094fdbec09b3e899b5db0

I'm looking into a way to re-add a Vector as a parameter option to the GetVirtualViewMatrix and GetViewMatrix to get a parallax effect. Maybe we don't need those though if we don't mind breaking changes to the API. Or we do as you say and rename the camera classes.

So the reason I changed Vector.One to 0 in my camera is that the parameter defines a depth level and not a parallax value. (The depth math does end up doing a parallax effect for free as a side effect though.)

PR #431, opened 2017-10-10
Added support for external renderers.

perneky:
With this Nuclex GUI can use visualizers, defined in an application. This allows people to develop custom controls, without compiling MG.EX (like from a nugget package).

One just has to add the binary, containing the visualizer, before the call to GuiManager.Initialize.


craftworkgames:
Great 👍

perneky:
Thank you.

PR #438, opened 2017-11-17
[GUI] Added support for icons on buttons

rafaelalmeidatk:
Allows the option to use images on buttons instead of only text:
Button example


craftworkgames:
Nicely done. I'm glad someone is giving the GUI system some attention.

I often think about getting back to the GUI system myself but other priorities have taken a front seat.

PR #440, opened 2017-11-21
ObjectPool Enumeration Bug

jason2211:
There were several issues with enumeration of the ObjectPool.

1. _headNode was never getting assigned
2. _tailNode's NextNode field was being set to itself causing infinite loop during enumeration
3. The GetEnumerator method was not including the _headNode in the results.  It was starting with the second node in the pool.


craftworkgames:
Thanks Jason, clearly this was implemented without any kind of testing.

Issue #428, opened 2017-10-03
[GUI] The layout control should be hovered as well when its child control is hovered


wutipong:
What's happening is, suppose there's a GuiLayoutControl which have Hovered style and contains children controls. If the mouse is moved to the layout control's child, the layout control loses its hovered status (even if the mouse is still inside the layout control). Only the child displays with hovered style. 

I think, in this case, both layout controls and child control should become hovered, in contrast to the child alone like what it is currently. 

There are two ways to implement this I believe, one is to make the parent become hovered once the hovering control is identified, another way is the exact opposite, the control becomes covered first and then it makes its children become hovered as well. Can't say exactly what is preferred here.

Anyway, I have played with it a little bit (using the latter approach). Here's my code https://github.com/wutipong/MonoGame.Extended/tree/cascade_ptr_event .

AristurtleDev:
Moving forward, MonoGame.Extended will be supporting GUM as the Gui component.  Please see https://www.monogameextended.net/blog/monogame-extended-gum/ for more information.

PR #441, opened 2017-11-22
[GUI] Propegation changed also adding in fix for Issue #428

tspayne87:
Changing Propegation to continue down the line if the event function returns true instead of false.  Also, added in some changes to deal with Issue #428, since the propegation could handle it pretty easy.

example


craftworkgames:
Wonderful :)

PR #443, opened 2017-12-17
[GUI] Binding Context / Control Style Fixes

tspayne87:
Both the binding context and control style were being saved in a global since.  This pull request should address this issue, and allow for multiple binding contexts for two separate screens.  Also the disable control style was only reverting one of the styles back to the previous state once disabled was turned off, this has also been addressed.

example


craftworkgames:
Great 👍

PR #444, opened 2017-12-26
Fix Demo.SpaceGame aiming and spaceship control

jonmseaman:
I was learning how to use MonoGame.Extended from the demos and noticed that the aim for SpaceGame was off. I went ahead and fixed it for myself.
It looked like the person who wrote had their axis backwards, since in MonoGame increasing y value means the sprite will move down the screen.

I wasn't sure if this was fixed in 2.0, as when I downloaded 2.0, this demo wasn't working.

Thanks


craftworkgames:
Looks like these changes where also included in the other PR but there's no conflicts so (probably) no harm in merging it anyway.

I'm pretty sure this demo had the axis around the right way at some point. I guess things have changed somewhere.

I haven't had a chance to fix up the demos in 2.0 yet. It's on my to-do list.

Issue #402, opened 2017-05-29
[NuclexGui] Change colour of GuiControlLabel


Nosrick:
Pretty much as the title suggests, it'd be nice to be able to change the colour of the labels. If it's possible, I'm not sure how to do it.

jonmseaman:
@Nosrick I was thinking about implementing this, as I could use the functionality as well. 
Would it suffice to be able to specify a frame/style with the label control? You would then just have to edit the skin file to set up some new styles to be used, and possibly some new fonts if you wanted to use a different font.

I could also set up a color property possibly... I'll think about it some more.

jonmseaman:
@Nosrick The pull request shown above should solve your issue if you are still interested. You can have labels with different styles, which allows for different colors to be used on different labels.

PR #454, opened 2018-01-10
Feature label styling

jonmseaman:
image

Added ability to add custom styles to labels where each label can have its own style.
Solution for #402 
 
There are also a few unit tests.


jonmseaman:
Also contains a few unit tests

craftworkgames:
Looks great. Thanks

Just so you know I've been making some significant changes to the ECS in the `2.0` branch. 

I'm also considering splitting out the Nuclex Gui at some point into it's own repository (not sure about this yet).

PR #460, opened 2018-01-31
Added XML docs to NuGet

jonmseaman:
Note that this does cause warnings to be generated for missing XML docs.


craftworkgames:
Great. Let's do it 👍

PR #461, opened 2018-02-03
Implemented GuiResizableControl

Mason11987:
Implemented GuiResizableControl, and made GuiWindowControl inherit from it.

GuiResizableControl is a control that can be resized.

GuiResizableControl calls SetCursorForResize(_resizingSides), which can be overridden to provide platform-specific mouse cursor implementation depending on where on the edge of the control your mouse is.

### DIFF
--- a/Source/MonoGame.Extended.NuclexGui/Controls/Desktop/GuiResizableControl.cs
+++ b/Source/MonoGame.Extended.NuclexGui/Controls/Desktop/GuiResizableControl.cs
@@ -1,0 +1,247 @@
+﻿using MonoGame.Extended.Input.InputListeners;
+using MonoGame.Extended;
+using System;
+using Microsoft.Xna.Framework.Input;
+
+namespace MonoGame.Extended.NuclexGui.Controls.Desktop
+{
+    [Flags]
+    public enum ResizingSides
+    {
+        None = 0,
+        Top = 1,
+        Right = 2,
+        TopRight = 3,
+        Bottom = 4,
+        BottomRight = 6,
+        Left = 8,
+        TopLeft = 9,
+        BottomLeft = 12,
+
+    }
+
+    [Flags]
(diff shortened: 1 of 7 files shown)


craftworkgames:
Could you please take a look at why the build is failing. Thanks.

Mason11987:
Ah, forgot to include the project file change.   Added another commit, should resolve it.

Source/MonoGame.Extended.NuclexGui/Controls/Desktop/GuiResizableControl.cs
```diff
@@ -0,0 +1,247 @@
...
+
+namespace MonoGame.Extended.NuclexGui.Controls.Desktop
+{
+    [Flags]
+    public enum ResizingSides
+    {
+        None = 0,
+        Top = 1,
+        Right = 2,
+        TopRight = 3,
```
rafaelalmeidatk: The sequence here should not be with powers of two? Like 0, 1, 2, 4, 8, 16 and so on?

Mason11987: Those represent combinations of flags, Top Right is Top & Right.

This allows you to say `resizingDirection.HasFlag(TopRight) if you'd like, instead of "resizingDirection.HasFlag(Top) && resizingDirection.HasFlag(Right)".  

According to this: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum

"CONSIDER providing special enum values for commonly used combinations of flags."

That's what those represent.

rafaelalmeidatk: You're right, I did not know this guideline. Thank you for the knowledge.

craftworkgames:
Great. I'll merge this shortly.

PR #462, opened 2018-02-03
Bug - Fixed default bounds of Screen desktop

Mason11987:
Screen Desktop defaulted to being width of 1, and height of 1, instead of width of 100%, and height of 100%


craftworkgames:
I'll merge this tonight.

PR #465, opened 2018-02-05
Apply kerning to the current glyph.

SenpaiSharp:
Kerning value 'amount' wasn't being applied to the '_currentGlyph.Position.X' in either MoveNext() methods, only '_positionDelta.X'. Both are needed for proper positioning.

### DIFF
--- a/Source/MonoGame.Extended/BitmapFonts/BitmapFont.cs
+++ b/Source/MonoGame.Extended/BitmapFonts/BitmapFont.cs
@@ -187,7 +187,10 @@ namespace MonoGame.Extended.BitmapFonts
                 {
                     int amount;
                     if (_previousGlyph.Value.FontRegion.Kernings.TryGetValue(character, out amount))
+                    { 
                         _positionDelta.X += amount;
+                        _currentGlyph.Position.X += amount;
+                    }
                 }
 
                 _previousGlyph = _currentGlyph;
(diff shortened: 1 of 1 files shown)


craftworkgames:
Looks reasonable but I don't have any obvious way to verify this.

Do you have a screenshot of the a being rendered with and without kerning?

When I originally wrote the BitmapFont stuff I used to render the font in another application (like [Paint.NET](https://www.getpaint.net/) or something and compare the output). Can you do the same with this?

SenpaiSharp:
Paint.NET does not use kerning with its default text tool (not sure about any plugins), a proper word processor would be better to check against. I'll use OpenOffice Writer as an example, though there will be minor differences due to sampling and such done by BMFont. The fonts I will demonstrate with is Komika Text, the latter which you can find here:

https://www.dafont.com/komika-text.font

examples

The top example is 36 point in OpenOffice.
The second example is BitMap Font in 48 px with height match as drawn with my change applied.
The third is the same as the above with the change applied.
The fourth is the first on top of the second (changed).
The fifth is the first on top of the third (unchanged).

SenpaiSharp:
For extreme demonstration purposes, here is manually changing the kerning of 84,101 to -15 from -4 (moving e closer into T) with my change.

fiexed

and without 

unfiexed

craftworkgames:
Thanks for the great explanation 😄

PR #466, opened 2018-02-06
Implemented Size3 and Point3.  Fixed bug in Point2.

Mason11987:
Implemented Size3 and Point3.  Fixed bug in Point2.

The documentation listed Point2, but code took in Vector2.  Changed to Point2


craftworkgames:
We haven't really focused much on the 3rd dimension in MonoGame.Extended but I guess this is a pretty good place to start.

PR #467, opened 2018-02-08
Remove duplicate ligne

sebgeelen:


craftworkgames:
Maybe they just wanted to be doubly sure that entity wasn't null 😉

sebgeelen:
Better safe than sorry 😄

PR #473, opened 2018-02-16
Fix #468 - RemoveComponents() null reference

### LINKED ISSUE #468: RemoveComponents() null reference
`RemoveComponents()` will throw a null reference exception if the game has a `EntityComponent` that is not attached to an `Entity`.

This is because `AddComponent` uses `componentType.Index` to determine which index on the `_entitiesToComponentBag` to insert an `Entity`. However if a `EntityComponent` exists but is not attached to an `Entity` yet then the component will have a `componentType.Index` value but will not have populated `_entitiesToComponentBag`.

This is not a problem until `RemoveComponents()` which iterates over `_entitiesToComponentBag` using a reversed for loop on every index but not all indexes are guaranteed to be populated.

Example of the problem:
`EntityComponent` with index [4] is not yet attached to any `Entity`.
Thus reverse for loop on the below = null reference
```
[0] Count = 1
[1] Count = 1
[2] Count = 1
[3] Count = 1
[4] null
[5] Count = 1
```
Solution seems to be just adding a null continue check to skip that loop iteration if there are no components

```
   internal void RemoveComponents(Entity entity)
        {
            Debug.Assert(entity != null);

            MarkEntityToBeRefreshed(entity);

            for (var i = _entitiesToComponentsBag.Count - 1; i >= 0; --i)
            {
                var components = _entitiesToComponentsBag[i];
                
                if (components == null)
                    continue;

                object component;
                if (!components.TryGetValue(entity, out component))
                    continue;

                components.Remove(entity);
                (component as IPoolable)?.Return();
                (component as IDisposable)?.Dispose();
            }
        }
```

Elthial:


craftworkgames:
Great thanks :)

PR #483, opened 2018-03-15
Scale Particles without Maintaining Aspect Ratio

mstauber:
Basically added a ScaleX and ScaleY, and a MaintainAspectRatio switch to ParticleReleaseParametes. Pretty simple change that adds some additional functionality to the particle system that shouldn't have any performance impacts.

Example:

```csharp
            _particleEffect = new ParticleEffect
            {
                Emitters = new List<ParticleEmitter>
                {
                    new ParticleEmitter(textureRegion, 500, TimeSpan.FromSeconds(2.5),
                        Profile.Ring(150f, Profile.CircleRadiation.In))
                    {
                        Parameters = new ParticleReleaseParameters
                        {
                            Speed = new Range<float>(0f, 50f),
                            Quantity = 1,
                            MaintainAspectRatioOnScale = false,
                            ScaleX = new Range<float>(10f, 60f),
                            ScaleY = new Range<float>(10f, 50f),
                            Rotation = 0
                        }
                    }
                }
            };
````

image
(Video:https://i.imgur.com/Aqx0Xzy.gifv)

### DIFF
--- a/Source/MonoGame.Extended.Particles/ParticleEmitter.cs
+++ b/Source/MonoGame.Extended.Particles/ParticleEmitter.cs
@@ -196,8 +196,17 @@ namespace MonoGame.Extended.Particles
                 _random.NextColor(out particle->Color, Parameters.Color);
 
                 particle->Opacity = _random.NextSingle(Parameters.Opacity);
-                var scale = _random.NextSingle(Parameters.Scale);
-                particle->Scale = new Vector2(scale, scale);
+                
+                if(Parameters.MaintainAspectRatioOnScale)
+                {
+                    var scale = _random.NextSingle(Parameters.Scale);
+                    particle->Scale = new Vector2(scale, scale);
+                }
+                else
+                {
+                    particle->Scale = new Vector2(_random.NextSingle(Parameters.ScaleX), _random.NextSingle(Parameters.ScaleY));
+                }
+                
                 particle->Rotation = _random.NextSingle(Parameters.Rotation);
                 particle->Mass = _random.NextSingle(Parameters.Mass);
                 particle->LayerDepth = layerDepth;
--- a/Source/MonoGame.Extended.Particles/ParticleReleaseParameters.cs
+++ b/Source/MonoGame.Extended.Particles/ParticleReleaseParameters.cs
@@ -14,6 +14,9 @@ namespace MonoGame.Extended.Particles
             Scale = new Range<float>(1f, 1f);
             Rotation = new Range<float>(-MathHelper.Pi, MathHelper.Pi);
             Mass = 1f;
+            MaintainAspectRatioOnScale = true;
+            ScaleX = new Range<float>(1f, 1f);
+            ScaleY = new Range<float>(1f, 1f);
         }
 
         public Range<int> Quantity { get; set; }
@@ -23,5 +26,9 @@ namespace MonoGame.Extended.Particles
         public Range<float> Scale { get; set; }
         public Range<float> Rotation { get; set; }
         public Range<float> Mass { get; set; }
+        public bool MaintainAspectRatioOnScale { get; set; }
+        public Range<float> ScaleX { get; set; }
+        public Range<float> ScaleY { get; set; }
+
     }
 }


mstauber:
Example:

```csharp
            _particleEffect = new ParticleEffect
            {
                Emitters = new List<ParticleEmitter>
                {
                    new ParticleEmitter(textureRegion, 500, TimeSpan.FromSeconds(2.5),
                        Profile.Ring(150f, Profile.CircleRadiation.In))
                    {
                        Parameters = new ParticleReleaseParameters
                        {
                            Speed = new Range<float>(0f, 50f),
                            Quantity = 1,
                            MaintainAspectRatioOnScale = false,
                            ScaleX = new Range<float>(10f, 60f),
                            ScaleY = new Range<float>(10f, 50f),
                            Rotation = 0
                        }
                    }
                }
            };
````

image

craftworkgames:
I've thought about doing this before but I'm not entirely convinced it's a useful change. Can you think of a useful real world particle effect that would benefit from this type of scaling?

I don't have a problem with adding features like this. I just want to put a bit more thought into it than "just because" that's all.

As for the performance impact. The first thing that popped into my head is [branch prediction](https://jacksondunstan.com/articles/2576). Although, you're probably right, since this is a [predictable pattern](http://igoro.com/archive/fast-and-slow-if-statements-branch-prediction-in-modern-processors/).  I don't think this will be a concern, but I wanted to bring it to your attention because sometimes little changes like this can have surprising effects.

Lastly, I noticed you added `ScaleX` / `ScaleY` and left `Scale`. This could be a slightly confusing API change. Then there's the flag thing. I dunno, it just feels a little messy.

mstauber:
> I just want to put a bit more thought into it than "just because" that's all.

Thank you, I appreciate that.

This is a digital effect I threw together with three emitters all using the same 1px texture in the demo:
https://i.imgur.com/TumMvyr.gifv

I would say most textures probably wouldn't benefit much from this ability. Though I think it could probably be combined with shaders to create some neat effects pretty easily. I'll see if I make an example for you.

As for implementation, 

I agree that the actual API change is not ideal and needs some thought into the design. 
I'm not sure how tolerable this project is to potentially breaking API changes, so I'm not exactly sure what a better solution might look like. At least I wanted to be sure you had something to test and play with if you wanted to.

rafaelalmeidatk:
This feature is indeed useful for create effects with one-pixel textures, I did a lot of that on real games.
About the implementation: you can replace the current `Scale` that is a `float` with a `Vector2` and switch the mode with `MaintainAspectRatioOnScale`. Or maybe you don't even need the `MaintainAspectRatioOnScale`, just a check if the `X` is different from `Y`.

mstauber:
That's a much simpler solution. 

The `ScaleInterpolator` will still work since it's already typed to `Vector2`. But yeah it will break the API.

craftworkgames:
> About the implementation: you can replace the current `Scale` that is a `float` with a `Vector2` and switch the mode with `MaintainAspectRatioOnScale`.

Yeah, this seems like the cleanest API design to me. Although, I think we should just call it `MaintainAspectRatio` and drop the "on scale" bit. I think that's implied (I can't think of another way the aspect ratio could be affected?)

> I'm not sure how tolerable this project is to potentially breaking API changes

You're right, this is a breaking change. In the past people have been quite understanding of breaking API changes and this is a relatively minor one (it causes a compile time error that's easy to fix e.g. `0.5f` becomes `new Vector2(0.5f)` in most cases).

Alternately we could make this change in the `2.0` branch. There's already a *lot* of breaking changes in there (and a lot more to come) which is both a good and bad thing depending on your perspective. If you're already heavily invested into `1.0` then it might be quite painful to go this way.

> That's a much simpler solution. You wouldn't need to check if X is different than Y or have a flag. You can just scale by the provided Vector2. (1,1), (1.3,1.3), (5,5), etc all maintain aspect ratio.

This option looks great on paper but I think I'd have to see it to fully wrap my head around it. There's something about it that's making my spidey senses tingle.

stefanrbk:
> There's something about it that's making my spidey senses tingle.

Same here. I looked at the implementation of `Range<>` and I don't believe `Range<Vector2>` is valid. `Range<>` requires a `IComparable<>`, which `Vector2` is not. It doesn't make sense for 1 `Vector2` to be greater than or less than another `Vector2`, which `Range<>` requires.

Without breaking the API in an even bigger way, the separate `ScaleX` and `ScaleY` would have to be used as independent `float`s if we don't want to change how `Range<>` works or the supporting classes to change the whole signature of `ParticleReleaseParameters`.

mstauber:
If `2.0` is much more tolerant of breaking changes.... another option could be to create something like an `IParticleReleaseParameters`, with methods like `Vector2 GetNextScale()`, `int GetNextQuantity()`, `HslColor GetNextColor()`, etc... providing a default implementation. Allowing developers to provide their own would hand over a lot of control.

craftworkgames:
You know, maybe we were overthinking this before.

I'll merge this now and we'll see how it goes with the implementation as is.

PR #485, opened 2018-03-18
Fix GUI children drawing

rafaelalmeidatk:
The background of the dialogues was being drawn below the controls of other dialogs.

Before the PR:
image
After:
image


craftworkgames:
Great thanks.

I'm planning on fixing many of these issues in `2.0` but there's no reason why we can't also have them fixed in `develop`.

rafaelalmeidatk:
Should I work directly on `2.0` instead of `develop` to catch your chances and avoid conflicts and already-done-work?

craftworkgames:
I'm in the middle of giving the GUI system a complete overhaul in 2.0.

For the moment I've completely disabled dialogues and windows so any changes you make in that area will probably get refactored when I get around to it anyway.

Once things are more stable you're welcome to join me.

PR #487, opened 2018-03-18
Work on GUI

rafaelalmeidatk:
I hope you don't bother if I do this on a single PR, creating a new branch for every fix is getting out of control.

### DIFF
--- a/Source/MonoGame.Extended.Gui/Controls/GuiButton.cs
+++ b/Source/MonoGame.Extended.Gui/Controls/GuiButton.cs
@@ -120,6 +120,11 @@ namespace MonoGame.Extended.Gui.Controls
 
         private void RecalculateIconPosition()
         {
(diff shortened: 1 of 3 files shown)


craftworkgames:
I'm in the middle of revamping the entire GUI system in the `2.0` branch.

That said, I think we'll leave the GUI system in `1.0` pretty much as-is for anyone currently using it. Bug fixes and non-breaking API changes are okay. 

So these changes might be okay to keep in `1.0` but until I get `2.0` stable I'll a little reluctant to merge this PR for now.

rafaelalmeidatk:
You are right, I will wait for the `2.0` before doing any more changes. I won't be using the UI code for some time so I am ok with these changes being only in my branch. Albeit I still think that the `LateDraw` is useful, we can think about it at a later time. Can the PR be closed?

PR #488, opened 2018-03-19
Add OnAddedToEntity and OnRemovedFromEntity methods to components

rafaelalmeidatk:
This allows the components to safely use `Get` on the entity without getting a `NullReferenceException`.

### DIFF
--- a/Source/MonoGame.Extended.Entities/EntityComponent.cs
+++ b/Source/MonoGame.Extended.Entities/EntityComponent.cs
@@ -80,6 +80,9 @@ namespace MonoGame.Extended.Entities
             _returnToPoolDelegate = null;
         }
 
+        public virtual void OnAddedToEntity() { }
+        public virtual void OnRemovedFromEntity() { }
+
         public override string ToString()
         {
             return $"Entity: {Entity}";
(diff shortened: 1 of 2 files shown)


craftworkgames:
I'm not convinced about this change for a couple of reasons:

1. Components are meant to be lightweight. In an ideal world, they would only be bags of data without any behavior. 
2. In the `2.0` branch I'm experimenting with removing the `EntityComponent` class entirely. The advantage of this is that *any* class can become a component.

For example, take a look at the [component](https://github.com/junkdog/artemis-odb/blob/master/artemis-core/artemis/src/main/java/com/artemis/Component.java) class from artimis-odb. It's literally nothing. Lightweight.

In [EntityX](https://github.com/alecthomas/entityx) the components are literally simple `struct`s. They don't even have a concept of a base component type.

stefanrbk:
That actually doesn't sound like a bad idea for `2.0`. I'm playing with the ECS and that makes sense to move away from inherited components.

If I remember from your original post, @rafaelalmeidatk, you wanted a way for entities to talk to each other. I did more research, and I am actually moving away from the event driven system I originally thought of as I found a better way. `EntityProcessingSystem`s. They are what manipulate your entities with your own logic, and, TBH, if you have 2 different kinds of components you need to talk to each other for changes, they should probably be combined into 1 component anyway.

PR #490, opened 2018-03-22
Added rotation to DrawRectangle and FillRectangle

msimpsoneu:


craftworkgames:
What origin do they rotate around?  The top left corner?
It might we worth considering adding an origin parameter as well.

msimpsoneu:
Good point, agreed. I fixed it on FillRectangle but didn't get to DrawRectangle and unfortunately life has gotten in the way of me working on it since then. For now I'll just close the pull request but if I get the chance I'll fix DrawRectangle or simply push the FillRectangle code I fixed earlier.

craftworkgames:
@msimpsoneu Yeah, no problems. I've got it on my to-do list to go over these methods at some point for the 2.0 release (it might not be anytime soon though).

It seems to me there's room for improvement to how we handle drawing shapes in Extended. These `SpriteBatch` extension methods are convenient but they also have a handful of downsides. Maybe it's worth discussing options in an issue.

In any case, I'll keep rotation in mind when we're considering what to do with this stuff.

Issue #230, opened 2016-08-05
[Tiled] Combine TiledMap-Rendering with Entity-Rendering


McSora:
Hey all,

While map-rendering is quite evolving at the moment I'd like to discuss a topic i came across while working the the old tiledmap-renderer.
I'm speaking of a possibillity to integrate entities (character, npcs, dynamically placable structures and so on) in the rendering process of the map.

Of course rendering one map at once is awesome! 
But facing facts, when it comes to oldschool 2D - 4Dir RPGs where you're working with maps with trees or rocks on it which in one case have to be drawn above the player and in another case are drawn beneath the player, the method of rendering single layers is not allways satisfying.

You can think of a tree which is .... 5 squares in height.
A character is about 1.5 squares in height and is colliding 0.5 squares in front of tree.
Let's say we have 2 layers. A bottom layer on which we are placing the bottom of the tree,
and a top layer with the rest of the tree.

We then render as follows:
```
spritebatch.Draw(bottomLayer,...);
spritebatch.Draw(character,...);
spritebatch.Draw(topLayer,...);
```
in this scenario the player would disappear behind the tree just fine.

now think of a character that is taller ... let there be godzilla with a total of 4 squares in height, standing in front of our tree. the bottom of the tree is drawn behind godzilla, the top of the tree in front of him although the tree should disappear completely.
Same goes for a placable structur like a den or something... maybe if you feature some building possibilitires in your game. 
in these cases single-layer-rendering doesn't fit anymore :(

To be honest I dont have a ready-to-use concept for this problem myself...
ie. it would be on the verge of madness to let the map handle a list of entities which it would then ... kind of order by their y-axis ...
I think it could be some higher level class which holds the map and the entities to combine rendering, but now that the map is drawing itself indipendently in the new renderer ... this could also turn out to be tricky.

Maybe it would be best to hear some other voices on the topic :)
Could also be i am thinking in the wrong direction here :D

LithiumToast:
This problem is pretty trivial. Use the Z value to give a depth which position things accordingly. If the depth is related to the height of the object everything will work correctly.

Here is an image explaining the Z value ordering with a scene graph. The image was taken from Apple's documentation [here](https://developer.apple.com/library/ios/documentation/GraphicsAnimation/Conceptual/SpriteKit_PG/Nodes/Nodes.html).
[...]

PR #491, opened 2018-03-27
ECS Split Layer Processing

stefanrbk:
I have a need in my project to `Update` or `Draw` the different system layers separately to allow processing between layers (primarily for layered `TiledMapLayer`s and `sprite`s).

What I currently draw:
* some map layers
* all sprites
* all other map layers

But I will need to be able to draw the following:
* some map layers
* some sprites
* some map layers
* all other sprites
* all other map layers.

The only other alternative I could think of to make this work is to somehow make the `TiledMapRenderer` work with `spritebatch` to allow for z-ordering.

### DIFF
--- a/Source/MonoGame.Extended.Entities/SystemManager.cs
+++ b/Source/MonoGame.Extended.Entities/SystemManager.cs
@@ -95,10 +95,24 @@ namespace MonoGame.Extended.Entities
             ProcessLayers(gameTime, _updateLayers);
         }
 
+		internal void Update(GameTime gameTime, int layer)
+		{
+			Debug.Assert(layer < _updateLayers.Length);
+
+			ProcessLayer(gameTime, _updateLayers, layer);
+		}
+
         internal void Draw(GameTime gameTime)
         {
             ProcessLayers(gameTime, _drawLayers);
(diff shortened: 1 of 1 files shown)


craftworkgames:
I'm a little confused by your changes.

I can see that you've changed some private and internal methods but I don't see were these methods are being called?  How do these changes actually help your situation?

stefanrbk:
I just realized this probably isn't the best way to do what I'm trying to do. 
What I'm trying to do is split when my NPCs are rendered based on what floor they are on. Think of walking under a bridge, then walking across the bridge. The bridge is in the tiled map and needs to be rendered before or after the sprites depending on if the player is supposed to be under or over the bridge. The sprites I can handle with a layer property in a `TransformContent` for the sprite, but I'm not sure how to use it to render all layer 0 sprites, pause sprite rendering in the `EntitySystem` to render more map, then render layer 1, etc.
With this, I was trying to make the ECS be able to call specific `EntitySystem` layers independently.

LithiumToast:
@stefanrbk See https://github.com/craftworkgames/MonoGame.Extended/issues/230

PR #501, opened 2018-05-16
Texture bug fix

stefanrbk:
When I released #480 to fix some texture alignment issues, I failed to check the variable types for the math being done. With my original fix by removing the `+ 0.5f` overdraw, the math changed from `float` division to `int` division. I'm now recasting for `float` division as it should be!

My bad!!


craftworkgames:
Great fix thanks.

I haven't had a chance to properly test it but it's such a small change I trust that it'll work okay. At some point I'm going to create a new Tiled map demo and I'll check it out then.

PR #502, opened 2018-05-16
Remove bug in TextureAtlas.RemoveRegion

stefanrbk:
When removing a `TextureRegion2D` from a `TextureAtlas`, referencing the `TextureRegion2D`s by index works, but referencing by name points to the old index. The only way I could think of making the current code work would be to rebuild the `Dictionary<string, int>` every time, or atleast the parts which are affected by the removal. Instead of rebuilding the internal `Dictionary<string, int> _regionMap` every time an entry is removed, I changed the signature to `Dictionary<string, TextureRegion2D>`. This eliminates the need to translate the name into a `List<TextureRegion2D>` index in the first place.

What do you guys think?

### DIFF
--- a/Source/MonoGame.Extended/TextureAtlases/TextureAtlas.cs
+++ b/Source/MonoGame.Extended/TextureAtlases/TextureAtlas.cs
@@ -35,7 +35,7 @@ namespace MonoGame.Extended.TextureAtlases
             Name = name;
             Texture = texture;
             _regions = new List<TextureRegion2D>();
-            _regionMap = new Dictionary<string, int>();
+            _regionMap = new Dictionary<string, TextureRegion2D>();
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Apostolique:
This looks like a fine fix for now. Your indentation seems broken in some places though.

Also, is there a reason we would ever access a texture atlas by index? We might just want to remove the _regions variable. Perhaps we should change that for 2.0?

stefanrbk:
The only reason I can think of is if you have pattern you want to use in an animation and you don't want to hard code each frame? IDK, I'm grasping at straws haha
And sorry about indentation. I'll check it and resubmit. I think I played with my tab settings :sweat_smile:

craftworkgames:
This fix seems like the way to go.

> Also, is there a reason we would ever access a texture atlas by index? We might just want to remove the _regions variable. Perhaps we should change that for 2.0?

Perhaps not. When I merge this I'll take a closer look and try to figure out if it really makes sense. It certainly adds a bit of complexity to I'm in favor of removing it if it's not really needed.

I'm just sorting out our build machine and a few other things and then I'll get around to merging these pull requests.

craftworkgames:
Hey guys,

I merged this tonight and made some small changes to the way it stores the "regions by index".

I almost figured out a way to write unit tests for it but unfortunately, those tests won't run on the build machine because MonoGame throws exceptions if "real" audio and graphics devices are not connected to the machine 😞 Maybe one day there'll be some hope for this.

PR #504, opened 2018-05-22
QuadTree Collision Library

jonmseaman:
I implemented a quad tree collision library. Please let me know what you think. I hope to be able to make some of the improvements listed below soon. 

I know I'm surprising you with a new API and a bunch of changes. Sorry about that.

# In this PR
1. Implements a QuadTree for collision detection using the IActorTarget API. This class is heavily tested by the unit tests.
2. Added CollisionComponent which does basic collision detection.
3. Added a demo featuring the CollisionComponent to Demo.Features
4. Some comments added to other classes in the collision library.

## What could be improved.
1. Continuous Collision - Currently, there is only basic collision. It would be very useful to have continuous collision detection.
2. Allow circles and oriented rectangles - Only Axis Aligned Bounding Boxes are currently supported.
3. Improved CollisionComponent API - If the quad tree is changed to use a Transform2D, it might be easier to use with the ECS system since they may use that to track location of entities.


craftworkgames:
I like what you've done here but I'm getting ready to merge the `2.0` branch back into `develop` so I'm going to hold off on merging anything major just yet.

Apostolique:
I think QuadTree should be written as Quadtree instead.

Issue #495, opened 2018-04-25
[Content] .NET Core Content Pipeline


WardBenjamin:
Since Nuget packages are stored in `~/.nuget/packages` instead of the `./packages` directory for .NET Core projects, the MGCB needs to reference a different path for the `MonoGame.Extended.Content.Pipeline` file - this just needs to be added to the documentation at http://craftworkgames.github.io/MonoGame.Extended/installation/.

I haven't figured out how to make this work with environment variables yet, but this path works for me when hardcoded: 

```/reference:/home/wardp/.nuget/packages/monogame.extended.content.pipeline/1.0.617/tools/MonoGame.Extended.Content.Pipeline.dll```

I'll be opening an issue on the main MonoGame repository about this as well since the solution isn't portable across different platforms/installations - I'll link it momentarily.

What a blast from the past.... _Last contributed on Feb 5, 2016_

craftworkgames:
Now that `develop` is running .NET Core it's a good time to take a look into this issue. I've run into it a couple of times myself.

I think we need to come up with a new (hopefully easier) process for dealing with the MonoGame Content Pipeline tool. It's always been a bit of a pain having to reference the DLL from the packages folder anyway.

I'd appreciate some thoughts on this?

WardBenjamin:
In lieu of a proper solution, such as making the MGCB NuGet-aware, I see two easy options, both of which are somewhat annoying for the end user:

* Manually copy the DLL file from the NuGet cache to a folder relative to the MGCB file, such as inside the content folder
* Do the same as above, but symlink it instead - this has the advantage of not having to copy the DLL every time there's a package update

I'm in favor of the second solution (and am currently using it), but it would have to be very clearly documented.

craftworkgames:
> Do the same as above, but symlink it instead - this has the advantage of not having to copy the DLL every time there's a package update

That's quite a clever idea. I'm not sure if we'll actually be able to automate it but still, it's clever.

> In lieu of a proper solution, such as making the MGCB NuGet-aware

It's entirely possible that we could aim for a "proper" solution rather than some dirty hack. Let's not rule out the possibility of making the MGCB NuGet aware.

Alternately, maybe there's something entirely different we could do. Like build our own content pipeline system?  I mean, we could still use the MGCB compiler to create the XNB files but maybe we can do away with the crappy GUI and have something that builds the content automatically from the command line.

[...]

PR #509, opened 2018-06-05
[WIP] Tiled Content Overhaul

stefanrbk:
I'll start by putting this up here.

This first change here removes the need for `Content.Load<Texture2D>` calls in the reader by passing an `ExternalReference`. It effectively does the same as the previous method, but removes all path manipulation to do it.

`Texture2D` in tilesets shouldn't be referenced more than once, so the `Texture2D` in tilesets are loaded into the tileset directly, not loaded separately.


craftworkgames:
Looking great mate.

I'd really like to restore or create a new Tiled Maps demo that shows off the features. It's also an important part of testing to see if we've broken anything. Unfortunately writing unit tests that involve parts of MonoGame has proven to be quite difficult but having a demo is quite useful.

When I get around to creating the demo I'll merge these changes. Unless you're interested in doing one?

stefanrbk:
I can look into making a demo. I threw some assets together just to test the changes as I was going. I'm gonna have to investigate why TeamCity didn't build... it would build and run on my machine...

stefanrbk:
I'm not the best when it comes to figuring out compiler stuff, but when I build the whole solution after cleaning the solution, my some of the projects will fail to build on the first attempt. The errors I usually get are that a file is locked by another process (generally involving the pipeline tool) or an asset fails to process (probably the pipeline tool trying to build a file before monogame.extended.content.pipeline.dll finishes building). Everything seems to build just fine after the first build fails.

craftworkgames:
> an asset fails to process (probably the pipeline tool trying to build a file before monogame.extended.content.pipeline.dll finishes building). Everything seems to build just fine after the first build fails.

Ah yeah. I've been trying to sort this one out. It's been a real pain so far and I haven't solved it yet. It's related to #495

stefanrbk:
> Ah yeah. I've been trying to sort this one out. It's been a real pain so far and I haven't solved it yet. It's related to #495

Yeah, that is definitely a pain. I guess it's a good thing I'm overhauling our content pipeline .dll in 1 swoop then!

stefanrbk:
I've been thinking about something to do for a demo, but I can't even think of where to start. The only real changes are backend in building the content files. My next goal is going to tackle implementing the additions to the .TMX format I added, now that will need a demo/full tutorial!
haha

craftworkgames:
Great news. @jonmseaman has been working on getting our demos back up and running 😄 

So soon we'll have the old Tiled demo back up and then we can make sure all the test maps load okay.

stefanrbk:
Sounds good to me!! 😂

stefanrbk:
After merging `develop` into my branch to test with the new demos, something is in fact broken somewhere...

stefanrbk:
WOO! I fixed it! I made a change in the Tiled importers and processors to actually throw unhandled errors if there is an issue building any Tiled-related asset as this will stop the build and show as an error in the VS Error List window. I should now be 98% complete with the Tiled overhaul. Only thing which is left before I fully sign off that the PR is done is I noticed a graphics glitch on level07 with transparency. I want to see if it does it for this PR or if it is an issue already in `develop`.

Source/MonoGame.Extended.Tiled/TiledMapTilesetReader.cs
```diff
@@ -0,0 +1,154 @@
...
+		{
+			if (existingInstance != null)
+				return existingInstance;
+
+			return ReadTileset(reader);
+		}
+
+		public static TiledMapTileset ReadTileset(ContentReader reader)
+		{
+			var texture = reader.ReadObject<Texture2D>();
```
stefanrbk: Guys, I need some help here... When loading the textures, there are some weird transparency things going on... Transparent black (`Color(0, 0, 0, 0)`) is rendering properly, but transparent white (`Color(255, 255, 255, 0`) is rendering the white. I'm not too experienced with the rendering system yet to know where even to look for where this inconsistency in alpha rendering is coming from (or even if it is our problem or MonoGame's). Anyone have any idea?

Apostolique: This should help you: https://microsoft.github.io/Win2D/html/PremultipliedAlpha.htm

Specifically this sentence:

> In this world, RGB and alpha are linked. To make an object transparent you must reduce both its RGB (to contribute less color) and also its alpha (to obscure less of whatever is behind it). Fully transparent objects no longer have any color at all, so there is only one value that represents 100% transparency: RGB and alpha all zero.

stefanrbk: I completely agree, but these aren't my assets 😛
What doesn't make sense is the texture transparency works as expected when using `Content.Load<Texture2D>()` and `ContentProcessorContext.BuildAsset<Texture2DContent>()` but not `ContentProcessorContext.BuildAndLoadAsset<Texture2DContent>()`. I have tried to pass `"PremultiplyAlpha"` to the processor for the texture, but that doesn't seem to help.

stefanrbk:
Transparency issue is "fixed". Here is some background as the in line comments are difficult to see from the issue earlier. 100% transparency should definitely be `Color(0, 0, 0, 0)`, but if someone uses `Color(255, 255, 255, 0)` it shouldn't affect the rendering of the texture between if the `Texture2D` is included in the `TiledMapTileset` .xnb file or built as its own .xnb file. Effectively, the content processor should zero out the color when the `Texture2DContent` is premultiplied in the content processor. Instead, it appears the opacity is being lost when a non-black color has 0 opacity. I removed all references to `ContentProcessorContext.BuildAndLoadAsset` and replaced with `ContentProcessorContext.BuildAsset` when it comes to `Texture2DContent`. This means the `Texture2D` will be built separately and not be included in the .xnb file for the tileset. I don't like this solution as it then requires 2 .xnb files per tileset (one for the `Texture2D` and one for the `TiledMapTileset`), but building the files together doesn't preserve the 100% transparency where needed.

PR #513, opened 2018-06-14
Rectangles now collide with circles

jonmseaman:
I added support and tests for collisions which include both rectangles and circles. I tested the collisions and they are working with the correct penetration vectors as far as I can tell.

My only regret for this implementation is that there are now `ICollisionActor` and `IActorTarget` which are basically the same, but one uses `Shape` instead of `RectangleF`. This was done because of the way CollisionGrid was using `RectangleF`, and I needed a generic shape.

Let me know if anything should be changed or improved. Thanks


craftworkgames:
Looks good.

The collisions stuff has always been pretty lightweight. I'll probably revisit it sometime before we release 2.0. Although, there's still plenty of other things on the to-do list before I get to it.

PR #515, opened 2018-06-17
Fix circlef contains

### LINKED ISSUE #511: CircleF.Contains() uses incorrect formula.
The CircleF.Contains() method simply subtracts the point provided from the center of the circle and then checks that the absolute values of the resulting value's x and y are less than or equal to the radius of the circle. However this does not work for telling if something is actually within the circle. Say you have a circle centered at 16,16 with a radius of 16. If you were to check the point 0,0 which is not within the circle but still fulfills the check then you get an incorrect response. I know that using the Math.Sqrt() method isn't very efficient but using the actual distance formula would result in the correct response. Without using the distance formula the CircleF is pretty much treated as a square for the purposes of the Contains() method as it currently is.

jonmseaman:
Fixes #511

### DIFF
--- a/Source/MonoGame.Extended/CircleF.cs
+++ b/Source/MonoGame.Extended/CircleF.cs
@@ -261,8 +261,11 @@ namespace MonoGame.Extended
         /// </returns>
         public static bool Contains(ref CircleF circle, Point2 point)
         {
-            var distance = circle.Center - point;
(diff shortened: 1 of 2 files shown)


craftworkgames:
Great work.

Oh look at that. Lot's of tests that need to be converted to xUnit eventually.

PR #518, opened 2018-06-19
Change to .NET BitVector32

stefanrbk:
Changed to the new .NET version of `BitVector32`. The main difference is that it works with `int` values instead of `uint` like our implementation, and you have to use `.Data` instead of accessing the internal data with a cast.

### DIFF
--- a/Source/MonoGame.Extended/MonoGame.Extended.csproj
+++ b/Source/MonoGame.Extended/MonoGame.Extended.csproj
@@ -21,4 +21,10 @@
     <PackageReference Include="MonoGame.Framework.Portable" Version="3.6.0.1625" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="System.Numerics.Vectors">
+      <HintPath>..\..\..\..\..\..\..\Program Files\dotnet\sdk\NuGetFallbackFolder\microsoft.netcore.app\2.0.0\ref\netcoreapp2.0\System.Numerics.Vectors.dll</HintPath>
+    </Reference>
+  </ItemGroup>
 </Project>
--- a/Source/MonoGame.Extended/MonoGame.Extended.csproj
+++ b/Source/MonoGame.Extended/MonoGame.Extended.csproj
@@ -18,6 +18,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Remove="Collections\BitVector32.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
     <PackageReference Include="MonoGame.Framework.Portable" Version="3.6.0.1625" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
   </ItemGroup>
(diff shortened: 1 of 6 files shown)


stefanrbk:
Oops, I forgot to fix the tests...

Source/MonoGame.Extended/MonoGame.Extended.csproj
```diff
@@ -17,8 +17,18 @@
...
+  </ItemGroup>
+
   <ItemGroup>
     <PackageReference Include="MonoGame.Framework.Portable" Version="3.6.0.1625" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="System.Numerics.Vectors">
+      <HintPath>..\..\..\..\..\..\..\Program Files\dotnet\sdk\NuGetFallbackFolder\microsoft.netcore.app\2.0.0\ref\netcoreapp2.0\System.Numerics.Vectors.dll</HintPath>
```
craftworkgames: This hint path is interesting? It references a DLL on the machine rather than a package on NuGet?

Source/Tests/MonoGame.Extended.Tests/Collections/BitVector32Tests.cs
```diff
@@ -171,22 +172,22 @@ public void SectionIncorrectSize()
...
-            Assert.True(bitVector[uint.MinValue], "UInt32.MinValue");
+            var bitVector = new BitVector32(-1);
+            Assert.True(bitVector[0], "Int32.MinValue");
         }
 
         [Fact]
         public void TestSectionIndexer()
         {
-            var bitVector = new BitVector32(uint.MaxValue);
+            var bitVector = new BitVector32(-1);
```
craftworkgames: -1 is the same as MaxValue?

craftworkgames:
Looks good. 

Most of these changes will be irrelevant when I merge the new ECS back into `develop`.

Source/Tests/MonoGame.Extended.Tests/Collections/BitVector32Tests.cs
stefanrbk: Since we are using `int` instead of `uint` we have [Two's compliment](https://en.wikipedia.org/wiki/Two%27s_complement) to deal with. The intent of the original code was for all 32 bits to be 1's, so -1 is opposite to 0.

Source/MonoGame.Extended/MonoGame.Extended.csproj
stefanrbk: I let VS add the reference as I didn't know where it was... Apparently I'll have to look at that haha

stefanrbk:
This could always be used with some sort of quest or state system for saving, so I might clean up the tests to show that off

Source/Tests/MonoGame.Extended.Tests/Collections/BitVector32Tests.cs
WardBenjamin: All 32 bits being 1's is equal to int.MinValue - negative one is 0b1000....0001

stefanrbk: `int.MinValue` failed the test. `int.MinValue` resulted in `10000000_00000000_00000000_00000000`

Source/MonoGame.Extended/MonoGame.Extended.csproj
stefanrbk: So, now that I'm looking at this, this is just to reference the new .NET `Matrix3x2` and `Vector2` which we aren't using. This is COMPLETELY unneeded... *facepalm*

craftworkgames:
I made the relevant changes when I merged the ECS back into `develop` so this PR can be closed now.

PR #524, opened 2018-07-10
One more PrivateAsset change!

stefanrbk:


stefanrbk:
How did that break the build? 😕

craftworkgames:
> How did that break the build? 😕

It doesn't look like it's your fault. Here's the error:

`Could not copy "C:\TeamCity\buildAgent\work\8dee3b4a1a2bbd93\Source\MonoGame.Extended.Content.Pipeline\Newtonsoft.Json.dll" to "bin\netstandard2.0\Newtonsoft.Json.dll". Beginning retry 1 in 1000ms. The process cannot access the file 'C:\TeamCity\buildAgent\work\8dee3b4a1a2bbd93\Source\MonoGame.Extended.Content.Pipeline\bin\netstandard2.0\Newtonsoft.Json.dll' because it is being used by another process.`

Not sure what's going on here. I'll reboot the server and re-run the build. Hopefully it doesn't keep happening.

Oh and just so you know, you should be able to take a look at broken builds yourself if you want by following these steps:

1. Click the "Details" link on the failed check to got to the TeamCity page.
2. Click on "Log in as guest".
3. Click on "Build Log".
4. Find the error (it can be a bit fiddly looking through the logs to find the "real" error sometimes).

If you can't login as a guest or something else doesn't work for you please let me know.

craftworkgames:
Ah okay, so I ran the build again and it produced a different error this time.

`The type or namespace name 'Pipeline' does not exist in the namespace 'Microsoft.Xna.Framework.Content' (are you missing an assembly reference?) [C:\TeamCity\buildAgent\work\8dee3b4a1a2bbd93\Source\Tests\MonoGame.Extended.Content.Pipeline.Tests\MonoGame.Extended.Content.Pipeline.Tests.csproj]`

I think this one might be real. The problem with making `PrivateAssets` is that projects that reference them don't see the private stuff. In this case, it's one of the unit test projects.

To reproduce this issue locally you should be able to run the cake build yourself (just run `build.ps1` if you don't have cake installed).

That said, I'm not entirely certain this change is actually the final change we need to make. I think there might still be some problems referencing the package in Content Pipeline. I've got it on my to-do list to explore this more but right now I'm quite tired and I'll be heading to bed soon.

stefanrbk:
Hahaha sounds good. I'll close and we can reopen later.

PR #527, opened 2018-07-16
Update TiledMapProcessor.cs

Raflos10:
Added support for build parameters for images from Tiled.

### DIFF
--- a/Source/MonoGame.Extended.Content.Pipeline/Tiled/TiledMapProcessor.cs
+++ b/Source/MonoGame.Extended.Content.Pipeline/Tiled/TiledMapProcessor.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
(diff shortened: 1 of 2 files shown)


stefanrbk:
This appears to undo PR #521. I would look [here](https://github.com/craftworkgames/MonoGame.Extended/pull/527/files#diff-11581d3ffb9618112ab0565b60a7b499L31) and make sure you pull the color from the tileset/image layer so transparency can be passed from the tileset or map file and not have to duplicate it with ColorKeyColor.

Raflos10:
It should still take the color from the image. I just have the image content passed to a method where the entire data dictionary is made, including the transparency color. 
I figured it was better than copy-pasting the whole thing when the code is more or less the same for both tilesets and image layers.

craftworkgames:
I'm missing quite a bit of context when it comes to this issue. 

Do you think you could provide a demo that shows an example of the problem we're trying to solve here?

I haven't been taking a hard line on this in the past but I think about time that all PR's include updates to the demos and / or unit tests to show what the changes actually do.

stefanrbk:
I didn't look close enough. PR #521 is still implemented [here](https://github.com/craftworkgames/MonoGame.Extended/pull/527/files#diff-11581d3ffb9618112ab0565b60a7b499R118)

PR #529, opened 2018-07-18
Fixed the link to Pipeline tool setup

### LINKED ISSUE #528: [Tiled] Missing documentation regarding setting up the content pipeline
Hi, I'm trying to follow the guide here: http://docs.monogameextended.net/Features/Tiled/
But at "Instructions to do so can be found here" I get linked to a 404, could you please fix this? I wouldn't know how to progress otherwise!

stefanrbk:
Fixes #528


craftworkgames:
Thanks mate. I really appreciate help with the docs.

I'm going to merge this now because it's technically better than what we had before.

Although, I think it's possible (and would be better) to use relative URLs in the future. This is so you can [build and run the docs site locally](http://docs.monogameextended.net/Documentation/) using `mkdocs`.

I plan to re-write some of these docs soon so I'll fix up the relative URL then.

PR #548, opened 2018-09-17
fix reflection bug

esengine:
When I want to create an instantiated control by means of reflection, it has an error and requires a constructor with no arguments. So I want to fix it.

Another one is to change the readability of updatecontrol to public, because when I want to use this library, there is a conflict, I have to manually update the state of the control. I can't update the state with guisystem.update. Because it will happen to find an instance of TouchPad.

https://github.com/MarcStan/MonoGame.Framework.WpfInterop/issues/10


craftworkgames:
I'm pretty sure I fixed this bug in the branch I'm working on already, but the fix looks essentially the same. I'll merge this now.

As for making `UpdateControl` public. I guess that's fine for now, although, generally speaking increasing the surface area of the public API can be a problem for refactoring code in the future. There might be a better solution to this problem that we can explore when the time comes.

PR #551, opened 2018-10-07
removed non-existing .nuspec files from solution

vividos:
Hi, here's my first pull request, maybe there are some more following. The .nuspec files were incorporated into the .csproj files and were removed in a commit, but are still in the solution file. This PR fixes that.


craftworkgames:
Great. Nice catch :)

PR #558, opened 2018-10-14
update of project to monogame 3.7

megabobik:


craftworkgames:
Hey thanks for doing this mate.

The build is broken. Could you try running `build.ps1` from your machine please.

megabobik:
script showed 4 tasks completed: restore, build, test, pack.

megabobik:
Maybe, you will make update of monogame lib?

craftworkgames:
> Maybe, you will make update of monogame lib?

Good thinking. I updated to the latest release of MonoGame 3.7 on the build machine.

Unfortunately, there's still some errors coming out in the build.

```
System.IO.FileNotFoundException : 
Could not load file or assembly 'MonoGame.Framework, Version=3.7.0.1708, Culture=neutral, PublicKeyToken=null'. 
The system cannot find the file specified.
```

I'm not sure exactly what's causing this yet. I'll do a little more digging later when I've got some time.

megabobik:
TeamCity show same error. I think you resolve issue locally, if you open VS and download all dependencies by nugget package manager.

megabobik:
After setting reference to ME.Content.Pipeline.dll, new types(tiled, astrid etc.) don't appear in processor and importer lists.

craftworkgames:
> After setting reference to ME.Content.Pipeline.dll, new types(tiled, astrid etc.) don't appear in processor and importer lists.

Oh, this fixes *that* problem. Good to know. I noticed people talking about it on the forums but I had no idea what was causing it.

Haha. I guess the lesson learned here is that it's probably a good idea to lead with the problem we're trying to solve *before* talking about the solution.

craftworkgames:
Oh my, this is worse than I thought. I tried to update the NuGet packages myself to replicate the issue and sure enough I'm getting the same problem.

But it's worse than that. None of the demos are working because they are either throwing a null reference exception or simply exiting immediately when the game runs.

The good news is I *think* I understand why it's happening. It's most likely got to do with how NuGet is resolving the package version.

See, the way a PCL bait and switch is supposed to work is that MonoGame.Extended references the PCL version of MonoGame (in this case `MonoGame.Framework.Portable` version 3.7.0.1708) and then the "real" game references a platform specific version of MonoGame (in this case `MonoGame.Framework.DesktopGL.Core` version 3.7.0.7).

The problem is (I think) when NuGet resolves which version of the DLL to use it sees that the PCL version is higher than the DesktopGL.Core version and uses the wrong one.

So to fix this we either need the DesktopGL.Core version to be the same or higher version than the PCL or to put it another way we need the PCL to be lower than the DesktopGL.Core version.

Given that we don't have direct control over the version of the DesktopGL.Core version (because this package is maintained by @cra0zy over [here](https://github.com/cra0zy/MonoGame/tree/core) then our best course of action right now is to continue to use the older PCL version (3.6).

That shouldn't be a problem because these bait and switch PCL's don't actually contain any implementation anyway and I doubt the public API of MonoGame has changed recently (since it's identical to XNA).

PR #559, opened 2018-10-17
Added loop to handle dynamic -> dynamic body collision resolution

zoeysaurusrex:
- The equality check is a little gross for dynamicBody -> otherDynamicBody, but it works for now.
- Added notes about how I did this.
- Tested it with the current platformer demo

### DIFF
--- a/Source/Demos/Platformer/Collisions/World.cs
+++ b/Source/Demos/Platformer/Collisions/World.cs
@@ -61,7 +61,21 @@ namespace Platformer.Collisions
                 }
             }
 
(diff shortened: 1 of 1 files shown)


craftworkgames:
Cool man. I mean it's not perfect but it's something.

I wish I had more time to complete the Platformer demo but I've always got a long to-do list.

zoeysaurusrex:
Since I'm using the World.cs in my own game for simple collisions (top down 2d which doesn't need much), I'm happy to contribute anything back.

PR #564, opened 2018-10-22
Changed to call initialisation if not done

emersont1:
as per request from http://community.monogame.net/t/screenmanager-wont-loadcontent/11043


craftworkgames:
Great work. Thanks :)

PR #565, opened 2018-10-24
Fix Tile Animation Textures

corykroll:
Noticed when upgrading MonoGame.Extended.Tiled NuGet from 1.0.617 to 1.1.0 that the Tiled tile animation textures were not rendering correctly.  This adjustment (which effectively reverts part of a check-in after 1.0.617) fixes the issue.

### DIFF
--- a/Source/MonoGame.Extended.Tiled/TiledMapTilesetTileAnimationFrame.cs
+++ b/Source/MonoGame.Extended.Tiled/TiledMapTilesetTileAnimationFrame.cs
@@ -21,8 +21,8 @@ namespace MonoGame.Extended.Tiled
         {
             var sourceRectangle = tileset.GetTileRegion(LocalTileIdentifier);
             var texture = tileset.Texture;
-            var texelLeft = sourceRectangle.X / texture.Width;
-            var texelTop = sourceRectangle.Y / texture.Height;
+            var texelLeft = (sourceRectangle.X + 0.5f) / texture.Width;
+            var texelTop = (sourceRectangle.Y + 0.5f) / texture.Height;
             var texelRight = (sourceRectangle.X + sourceRectangle.Width) / (float)texture.Width;
             var texelBottom = (sourceRectangle.Y + sourceRectangle.Height) / (float)texture.Height;
 


craftworkgames:
Hey @corykroll 

Thanks for the PR. There's a few things I want to go over first before merging this change.

1. I don't normally accept pull requests into the `master` branch. We're trying to use [git flow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) these days and I think changes to `master` are technically a called a hotfix. In this case, it probably does make sense to be a hotfix to version 1.1 but tbh I've never really done one before so I'm kinda new to this workflow as well.

2. I had a look at the [same code](https://github.com/craftworkgames/MonoGame.Extended/blob/develop/Source/MonoGame.Extended.Tiled/TiledMapTilesetTileAnimationFrame.cs#L20) in the `develop` branch and as far as I can tell it doesn't have the same fix. It also doesn't seem to have the bug either. So maybe we've already fixed it in another way or perhaps the bug isn't actually in Extended after all?

tiled-animated-mushroom

3. What version / package of MonoGame are you using?  It might be a difference between the DirectX and OpenGL versions that cause the need for the half pixel offset. If that's the case, you might have fixed it for DirectX and broken it for OpenGL. In either case, it might already be fixed in the latest version of MonoGame. This would be worth investigating.

4. Lastly, if I do merge this back into `master` I'm not sure how easy it'll be to generate new NuGet packages. It does look like we've got a cake build (i.e `build.ps1`) so maybe it's not so hard. I'm just sayin' there might be more work involved.

So yeah, I think we need to do a bit more investigation before making this change. In particular we need to know if this is the "right" fix or not. Of course, if this fix works for you and you're happy with it one option is to own fork.

corykroll:
I admit I'm pretty new to this exact process, so I'm happy to adjust my pull request or ultimately fork if what I'm after doesn't fit back into the project.  I'm just happy to help where I can - I love what you guys are building here!

1. I originally thought to look at doing a pull request to the develop branch, but since that looks to be a larger re-factoring and this is more of a hotfix, master seemed like the better choice

2.  I originally noticed this change going from NuGet package 1.0.617 to 1.1.0.  So I forked and downloaded the project, switched to master branch, made sure I was on tag 1.1.0 and built it.  I got the bad animation as the result when I pointed my own project to the build.  I looked into the git history and this check-in look suspicious (https://github.com/corykroll/MonoGame.Extended/commit/2005c2c1617ba02a086530bd38143c45398b67bf).  Figuring that change is what caused my issue, I went about undoing it locally and retesting.  I definitely can see the before and after.  But my environment might be the difference here...

3. I am using MonoGame 3.7.0 currently.  I am building a UWP game on a Windows computer, so I assume it is using DirectX under the hood.  That said, this animation issue that I'm trying to fix came up prior to 3.7.0 though too (I had been installing Development Builds prior to the 3.7.0 release) so I don't think it is that MonoGame runtime specifically.

4. I understand this situation for builds may not be ideal for NuGet packages.  If there is any way I can help with that, I'd be happy to give it a go.  What I'm essentially looking for is a hotfix to the current 1.1 package.  This might also break OpenGL though (admittedly not something I thought of before submitting), and maybe the right solution is a switch in the code based on what is running under the hood to work in all scenarios.  I could look at creating several projects (one that uses DirectX and another OpenGL) to get looked into further.

Ultimately I'm open to whatever you'd like to see done next to move this forward.  Thank for your consideration and feedback on the pull request.

stefanrbk:
This have to do with my PR #501 which fixed a bug I accidently introduced?

stefanrbk:
So, I looked closer and this is what I fixed right after 1.1.0 came out. Please see PR #480 for where I accidentally added the bug when fixing a texture alignment issue, and PR #501 for the fix. 480 deals with a + 0.5f added to most of the rendering code to fix an issue with the rendering code previously where tiles would render with gaps between tiles if the textures came right up to each other. That introduced some artifacts when zooming in and the original issue of the texture gaps was nowhere to be seen, so I removed all the `+ 0.5f`s but didn't realize the part of the code you pointed out just needs to be a float, which I didn't realize the ` + 0.5f` did originally.

craftworkgames:
@stefanrbk Thanks for taking a closer look at this, although, I'm still a little confused. Are you saying we should merge this PR or not?

I just tried to reproduce the issue with the gaps when zooming in and it doesn't seem to be happening for me on the `develop` branch. Am I missing something?

Just to be clear, if there's a clear and obvious way to reproduce an issue I'm more than happy to merge a PR to fix it. However, simply saying that something is "not rendering correctly" isn't enough to know if the PR is correct or not.

stefanrbk:
> @stefanrbk Thanks for taking a closer look at this, although, I'm still a little confused. Are you saying we should merge this PR or not?

Sorry, I'm saying I already fixed the issue in `develop` and this PR can be closed. This animation bug has been in the code since 1.1 was released back in May and I've seen quite a few PRs and Issues revolving around it. Is there a possibility of doing a "bug fix" release if the next version is still a while out?

craftworkgames:
> Is there a possibility of doing a "bug fix" release if the next version is still a while out?

I'm certainly open to the possibility of doing a bug fix release. Although, I will say that I'm not sure how much work will be involved. Who knows, maybe it'll be easier than I think. 

Anyway, it sounds like this PR isn't quite the right fix for this issue because it's inconsistent with how it's fixed in `develop`. I'll close this PR for now.

PR #567, opened 2018-10-29
Large improvements to Gui. Few add-ons in other places

gktval:
Many changes to the GUI. Most of these changes were help make things more functional. This includes making the gui system thread-safe and adding a slider panel. There were lots of problems and errors with control placement and keeping things aligned within the clipping rectangle. Other small changes were creating 'autofit' in the label control which allows for word wrapping inside its parent container. Also, added a 'hot icon' to the button control to draw a separate image over the button. The stack panel can now wrap items if the number of controls goes beyond the width of the control. Lots of other small bug fixes and amendments were made to help render controls correctly. As these changes were made, I tried to refer back to the demo gui project to make sure that it was still functional in its current state.
Changes to the camera class were to make the viewport public so that it could be reset, thus allowing rotation of the screen on the phone.
Many improvements to the collision grid.
Created a new texture packer class, which allow someone to combine several sprites with different colors into a singe texture. I use this class for drawing the main character, which may have different hair styles, clothing, and skin. Each of these styles also have different colors.
Scaling was added to the bitmap font class so that all the font could be scaled easily.
A method in transform was added to get the position as a delegate.

I can create some demos for the slider panel and the texture packer class if you want. For now, if you want to see the functionality of all these changes, it is easiest just to download my game at http://peregeronline.com/play-pereger
The whole game was built with assistance from Monogame.Extended, which is quite a remarkable project and has greatly helped me along the way. I hope that my commits will contribute in some small way.


gktval:
broken

PR #575, opened 2018-11-30
Fixed ContentReaderExtensions

### LINKED ISSUE #573: [Content]

valarion:
Fixes #573


craftworkgames:
Great work.

It's probably a good idea to also include some of the more trivial test cases as well.. like `[InlineData("resource", "resource")]`. I'm sure it already works but having these extra test cases helps to prevent accidental breakages in the future.

I'll merge this soon.

PR #583, opened 2019-01-16
Update installation.md

autemox:
Added precise location of nuget packages folder to clear up confusion about it being in the solution folder (it is not), and by default windows 10 normal search feature does not check the .nuget folder, thus a search for the dll will find nothing and can lead to a lot of confusion for new users.


craftworkgames:
Great work. Thanks mate.

PR #585, opened 2019-01-23
Remove isLoaded when load screen

KevinYeti:
It seems the two IF conditions is not needed.
When using the function of ScreenManager.LoadScreen(),  the isLoaded variable has already been setted to true.


craftworkgames:
I'm still unsure if this is the "correct" change but I'll merge it anyways.

PR #589, opened 2019-01-30
Notify subscribers before removing components from entity

### LINKED ISSUE #577: Entity can't be retrieved on OnEntityRemoved
If I use this following code on an `EntityProcessingCode`:

    protected override void OnEntityRemoved(int entityId)
    {
        var entity = GetEntity(entityId);
    }

`entity` will be `null`, I don't know what it was or what component it had before being removed. I solved this problem by moving [this line](https://github.com/craftworkgames/MonoGame.Extended/blob/dbc57e0f8acd8d9635c65f11b3ce40b7603670b1/Source/MonoGame.Extended.Entities/EntityManager.cs#L109) to the top of the method. I am not sure if this is the best solution.

rafaelalmeidatk:
Fixes #577

### DIFF
--- a/Source/MonoGame.Extended.Entities/EntityManager.cs
+++ b/Source/MonoGame.Extended.Entities/EntityManager.cs
@@ -98,15 +98,15 @@ namespace MonoGame.Extended.Entities
 
             foreach (var entityId in _removedEntities)
             {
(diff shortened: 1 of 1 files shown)


craftworkgames:
Finally getting around to merging this.

PR #594, opened 2019-02-17
Math folder

lithiumtoast:
- Move appropriate files to a new `Math` folder in the core project.
- Remove `.DotSettings` files from being ignored.
- Allow ReSharper to ignore namespace checks for the `Math` folder by adding `MonoGame.Extended.csproj.DotSettings`.


lithiumtoast:
@craftworkgames This PR does not break anything; the `Shapes` namespace and the code inside is not yet moved.

PR #595, opened 2019-02-26
Some input listener code for migrating games from System.Windows.Forms

ryancheung:
1.  Add Handled property to KeybordEventArgs for migrating games code using System.Windows.KeyEventArgs
2. Let MouseButton enum can be used as bit flags, as the System.Windows.Forms.MouseButton do

### DIFF
--- a/Source/MonoGame.Extended.Input/InputListeners/KeyboardEventArgs.cs
+++ b/Source/MonoGame.Extended.Input/InputListeners/KeyboardEventArgs.cs
@@ -21,6 +21,7 @@ namespace MonoGame.Extended.Input.InputListeners
                 Modifiers |= KeyboardModifiers.Alt;
         }
 
+        public bool Handled { get; set; } = false;
         public Keys Key { get; }
         public KeyboardModifiers Modifiers { get; }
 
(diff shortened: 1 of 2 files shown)


stefanrbk:
Sounds reasonable. Where are those flag values coming from though?

ryancheung:
They come from System.Windows.Form’dll

stefanrbk:
Microsoft and their complicated looking code from decades ago... *rant, rant*

Is there a benefit of using the _exact_ values from Microsoft and not code them to something more readable like:

```
0x0
0x1
0x2
0x4
0x8
0xf
...
```

craftworkgames:
We need to be clear about what problems we're actually trying to solve here?

1. I can see how being able to use bit flags might be useful. However, I can't see anything in your commit that tells me if it actually works?  As far as I can tell, there'll need to be more changes made to make these bit flags actually useful.

2. I agree with @stefanrbk. What is the actual purpose of the values used? Does it matter if they match the ones in `System.Windows.Forms`?

3. Same goes for the `Handled` flag. What should happen when you set it to `true`?  Does it change the behavior of anything?  As far as I can see from this commit, setting the flag wouldn't actually *do* anything.

It would be much clearer for me if you told your story.  What is it that you're doing that drives these changes?

For example, "I'm currently migrating my game from `System.Windows.Forms` and I ran into an issue because (reasons) and I need to change this because (reasons) and so on..."

As it is right now, I can't see a good reason to accept this pull request.

PR #606, opened 2019-05-23
Added event handlers for control events

patrickvandewal:
Hi!

I've added a few control event handlers to the control class so these can be implemented inherited types. 

The use-case for this is that I am building a game which has an inherited type of button where I have to catch specific enter, leave, down and up events on the control to execute actions. 

Cheers
Patrick


craftworkgames:
These `OnPointerDown` methods are already marked `virtual` so they can already be overridden. 
I don't see why we need to add extra events.

PR #612, opened 2019-06-29
Improve collision demo

jonmseaman:
- Fix an issue with actor bounds not being set properly. This makes the collisions in the demo smoother.
- Improved the demo's code quality by adding some comments.


craftworkgames:
The truth is, the collision library in Extended has a lot to be desired. Chances are we'll end up replacing or removing it at some point anyway.

That said, any incremental improvements are welcome.

PR #614, opened 2019-07-01
Fix Clear method

rumkuhgel:
Make sure to clear the array before setting count to zero.


craftworkgames:
Oh right, because otherwise the `Array.Clear` method will have `0` passed in as the `length` parameter.

Issue #617, opened 2019-07-15
TiledMap ArgumentException Unclear As To What Problem Is


manbeardgames:
If you import a Tiled map (.tmx) that contains two or more layers with the same name, an `ArgumentException` is thrown in the [AddLayer method of the TiledMap class.](https://github.com/craftworkgames/MonoGame.Extended/blob/130daf3e5e7dc02e1b2abc6244553314fd147eac/Source/MonoGame.Extended.Tiled/TiledMap.cs#L72)

This happens because `_layersByName` is a Dictionary object using `layer.Name` as the key. When the layer is added, if one already exists, the ArgumentException is thrown.

If you are using MonoGame.Extended from a NuGet package, when this exception is thrown, the only thing you see as the user is 
```
System.ArgumentException: 'An item with the same key has already been added.'
```

Looking at the stack trace for the exception, you can see the AddLayer method mentioned

```
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at MonoGame.Extended.Tiled.TiledMap.AddLayer(TiledMapLayer layer) in C:\TeamCity\buildAgent\work\8dee3b4a1a2bbd93\Source\MonoGame.Extended.Tiled\TiledMap.cs:line 81
```

From an end user perspective though, it is quite unclear as to why this exception is thrown and can be very confusing and frustrating. 

I would like to suggest adding a check before adding the layer to see if the key exists. If it does exist, manually thrown an exception explaining why.  Something like the following:

```csharp
if (_layersByName.ContainsKey(layer.Name))
{
    throw new Exception($"The TiledMap \"{Name}\" contains two or more layers named \"{layer.Name}\". Please ensure all layers have unique names.");
}
```

craftworkgames:
Yeah, I this is a good idea. It's always bothered me that the default `ArgumentException` doesn't tell you what key has already been added. I mean it's gotta be right there when they throw right?

Just a couple of things to consider:

 - Generally speaking it's better practice to throw something a little more specific instead of `Exception`. I would usually suggest [throwing `InvalidOperationException`](https://docs.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#use-the-predefined-net-exception-types) but in this case keeping it as a `ArgumentException` will be a non-breaking change (in case someone is already catching that type of exception).

[...]

PR #618, opened 2019-07-15
Check for existing layer name

manbeardgames:
This PR adds checks for an existing layer by name before adding new layer for a TiledMap (see #617) If layer name already exists, a descriptive exception is thrown instead of the generic ArgumentException.


craftworkgames:
I made some small tweaks. Let me know what you think.

PR #622, opened 2019-07-23
Collision fix

janfokke:
ICollisionActor.Bounds gets called multiple times in CollisionComponent.Update
Which in some circumstances can cause an exception when making a PenetrationVector


craftworkgames:
Hey @janfokke 

Sorry I didn't get to this earlier. I saw it in my email but I've been super busy lately.

I just want to let you know I do appreciate any code fixes as pull requests and I do my best to merge them in as time permits.

However, the collision library in Extended is quite bare bones. It was never really done properly for various (long forgotten) reasons. It really needs to be re-written from the ground up to be what I'd always hoped it would become.

So long story short, it's likely that most of the code in the Collisions library will be thrown out and replaced anyway.

I am curious to know if you're using it in a game though?

PR #639, opened 2019-10-25
Fixed pathing issues creating problems loading Tiled assets

corykroll:
I was having an issue loading Tiled maps, since in some cases the path to the asset contained \\ and / along the same path.  When debugging the pipeline writers, I would see things like this for the path:
"C:\\Users\\Cory\\source\\repos\\MonoGame\\Tools\\Pipeline\\Content/CollisionTileSet.tsx"

Using Path.Combine instead kept the slashes consistent and fixed the issue.

I looked for other placed in MonoGame.Extended with this same pattern and replaced them with Path.Combine.


craftworkgames:
Looks pretty sensible. Thanks :)

PR #641, opened 2019-12-10
TiledMap property multi-line bug fix

### LINKED ISSUE #635: Tiled Importer fails to read maps with multiline string properties
As outlined in the [TMX format documentation](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#property), a string property that contains multiple lines is stored inside the `property` element rather than in the `value` attribute:

> When a string property contains newlines, the current version of Tiled will write out the value as characters contained inside the property element rather than as the value attribute. It is possible that a future version of the TMX format will switch to always saving property values inside the element rather than as an attribute.

The Tiled Importer does not reflect this, causing it to throw an exception due to the `value` attribute being missing.

janfokke:
#635 fix


craftworkgames:
Ah man, sorry I totally forgot to merge this. I'll do it now.

PR #642, opened 2019-12-25
Improved collision performance and Re-added collision demo

janfokke:
Performance Improvements
- `QuadtreeData `now keeps track of its parents. This makes removing it from a `Quadtree `a lot faster.
- `Reset()` is removed and Dirty `QuadtreeData `is now tracked and reset inside the functions that use the Dirty flag.
- `Quadtree.Contents` is now a `HashSet`. This makes removing faster
- `Shake()` now only gets called at the end of update and explicit remove

I tested this with 1000 colliders and the fps went from ~1 to 60

### DIFF
--- a/Source/Demos/Tutorials/Demos/CollisionDemo.cs
+++ b/Source/Demos/Tutorials/Demos/CollisionDemo.cs
@@ -1,0 +1,237 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.Xna.Framework;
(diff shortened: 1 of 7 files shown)


craftworkgames:
@janfokke I really appreciate the work you've been doing lately. Especially the effort to revive the demos again.

What are your thoughts on the future of MonoGame.Extended?  I'm looking for someone to help figure out what it should be in 2020.

PR #647, opened 2020-02-04
Additional Shapes - Issue 629

padesso:
This PR contains an implementation using a PrimitiveBatch in lieu of a ShapeBatch for drawing additional shapes.  These shapes include: 
- Points
- Rectangles
- Filled Rectangles
- Circles
- Filled Circles
- Line Segments
- Polygons
- Filled Polygons

The polygon implementation using the ear clipping method.  I made every attempt to keep the code format consistent and updated the Tutorial Shapes project with an example of usage.


craftworkgames:
Thanks :)

PR #655, opened 2020-02-25
Collison

janfokke:
I made these changes because I had collision performance problems with my own game. 

The most significant change is the layering mechanism where the user can use flags to indicate which layers can collide with each other. I implemented this in a way that is almost completely backward-compatible except for `QuadTree.Query()`. Now you have to provide a `QuadTreeData` instead of an `IShape` because `QuadTree` now uses layers.


Source/MonoGame.Extended.Collisions/QuadTree/QuadTree.cs
```diff
@@ -42,10 +47,11 @@ public Quadtree(RectangleF bounds)
         /// <summary>
         ///     Counts the number of unique targets in the current Quadtree.
         /// </summary>
+        /// <param name="max">Max amount of items to count</param>
         /// <returns>Returns the targets of objects found.</returns>
-        public int NumTargets()
+        public int NumTargets(int max = -1)
```
janfokke: I added the max for performance in `Shake()`

janfokke:
The error 
Could not copy "C:\AzurePipelines\_work\1\s\Source\MonoGame.Extended.Content.Pipeline\Newtonsoft.Json.dll" to "bin\netstandard2.0\Newtonsoft.Json.dll
Doesn't seem to be related to my changes

Issue #601, opened 2019-04-05
[GUI] StackPanel Padding


davidbezdek:
Hello,

If padding is set on stackpanel content is indented only from bottom and right
**Branch:** develop

**Steps to reproduce:**
In demo gui set padding on stack panel
```
var stackTest = new DemoViewModel("Stack Panels",
                    new StackPanel
                    {
                        Padding = 20,
                        Items =
                        {
                            new Button { Content = "Press Me", HorizontalAlignment = HorizontalAlignment.Left, VerticalAlignment = VerticalAlignment.Top },
                            new Button { Content = "Press Me", HorizontalAlignment = HorizontalAlignment.Right, VerticalAlignment = VerticalAlignment.Bottom  },
                            new Button { Content = "Press Me", HorizontalAlignment = HorizontalAlignment.Centre, VerticalAlignment = VerticalAlignment.Centre  },
                            new Button { Content = "Press Me", HorizontalAlignment = HorizontalAlignment.Stretch, VerticalAlignment = VerticalAlignment.Stretch },
                        }
                    });
```
![image](https://user-images.githubusercontent.com/31237584/55631295-01f5dd80-57b8-11e9-8e40-fb3b0b05f768.png)

craftworkgames:
Thanks for the bug report.

lithiumtoast:
Moved to v4 as not sure what the future will be for the GUI project. I already removed the `NuclexGui` due to some problems and being confusing with this GUI project. There are also other GUI libraries out there like [described in this community post](https://community.monogame.net/t/mono-game-ui/10030/4?u=lithiumtoast).

AristurtleDev:
Closing this as MonoGame.Extended is moving to use GUM as the GUI tool.

See this blog post for more information: https://www.monogameextended.net/blog/monogame-extended-gum/

PR #656, opened 2020-02-26
#601 [GUI] Padding fix, Uniform grid padding offset

davidbezdek:
-Fix for issue https://github.com/craftworkgames/MonoGame.Extended/issues/601
-Added offset for UniformGrid


craftworkgames:
Okay

Issue #657, opened 2020-02-27
ArgumentNullException in EntityManager.Update


geirsagberg:
Happens irregularly when destroying entities.

Stack trace:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'item')
  at at MonoGame.Extended.Collections.Pool`1.Free(T item)
  at at MonoGame.Extended.Entities.EntityManager.Update(GameTime gameTime)
  at at MonoGame.Extended.Entities.World.Update(GameTime gameTime)
  at at Microsoft.Xna.Framework.Game.<>c.<.cctor>b__116_1(IUpdateable updateable, GameTime gameTime)
  at at Microsoft.Xna.Framework.Game.SortingFilteringCollection`1.ForEachFilteredItem[TUserData](Action`2 action, TUserData userData)
  at at Microsoft.Xna.Framework.Game.Update(GameTime gameTime)
  at at Microsoft.Xna.Framework.Game.DoUpdate(GameTime gameTime)
  at at Microsoft.Xna.Framework.Game.Tick()
  at at Microsoft.Xna.Framework.SdlGamePlatform.RunLoop()
  at at Microsoft.Xna.Framework.Game.Run(GameRunBehavior runBehavior)
  at at Microsoft.Xna.Framework.Game.Run()
  at BunnyLand.DesktopGL.Program.Main() in C:\Projects\BunnyLand\src\BunnyLand.DesktopGL\Program.cs:19
```

Image: 
![image](https://user-images.githubusercontent.com/594564/75462553-b8d8ea00-5984-11ea-9d7e-c4293e590ad1.png)

Public repo: https://github.com/geirsagberg/BunnyLand
Happens when holding Ctrl to shoot, after a while

geirsagberg:
Maybe do a null check before calling `.Free(entity)`?

geirsagberg:
I narrowed it down to the fact that the same entity is added to the _removedEntities list twice. I saw there is a Debug.Assert to prevent this, but those are stripped out of release builds. I added a PR that instead checks if it is already added, and doesn't add it twice: https://github.com/craftworkgames/MonoGame.Extended/pull/658

geirsagberg:
Reolved via #658.

PR #658, opened 2020-02-27
Don't add entity to _removedEntities if it is already there

geirsagberg:
I ran into a bug where an entity was seemingly added to _removedEntities twice, and because Debug.Asserts are stripped out when running the NuGet package, it wasn't caught until it was attempted to be freed, see issue https://github.com/craftworkgames/MonoGame.Extended/issues/657. I fixed it by adding a check and not adding it twice to the RemovedEntities list.


craftworkgames:
Seems sensible.

PR #661, opened 2020-03-25
Add RectangleF.TopRight

jamieyello:
There was a TopLeft and a BottomRight, but no TopRight or BottomLeft... But there could be. :)


craftworkgames:
Thanks.

PR #671, opened 2020-06-25
removed ownership handover note

janfokke:
Since the project has been handed over to @lithiumtoast, the handover note is no longer relevant.


lithiumtoast:
Thanks!

PR #677, opened 2020-07-05
v4.0

lithiumtoast:
This is an on-going PR for v4; it will be updated as we do more work on v4.

The idea for v4 is to start completely fresh with the latest understandings of how to write efficient and safe C# code in 2020. To do this all the code is wiped to a clean state and then any necessary code is copied and adapted as necessary.

Decisions made moving forward:

- There is 1 core "portable" package called [`Ankura`](https://github.com/craftworkgames/Ankura) for the new API which is detached and replaces MonoGame. (It's called Ankura as a development project and may change later.) The API is a fork of [Veldrid](https://github.com/mellinoe/veldrid) with influences from [FNA](https://github.com/FNA-XNA/FNA), [bgfx](https://github.com/bkaradzic/bgfx), and [sokol](https://github.com/floooh/sokol). The API will have a legacy version to shoehorn the XNA API for existing projects. The reason to detach from MonoGame is that the *development* of certain features/capabilities for using GPUs in C# (such as compute shaders) is often too slow and third parties such as Veldrid have already made significant ground.
- There 1 "desktop" package for the content pipeline called [`Assetyard`](https://github.com/craftworkgames/Assetyard). This effectively replaces the content pipeline.
- There is 1 package for each platform. These packages pull in all the dependencies necessary for the specific platform. E.g. `Akura.iOS`, `Ankura.Android`, etc.
-  Each feature is its own independent NuGet package. This makes it easier for people to use only what they want from MonoGame.Extended without pulling in everything. E.g. `MonoGame.Extended.Batching `, `MonoGame.Extended.TextureAtlas`, `MonoGame.Extended.Tiled`, etc. This repository will remain at is (`MonoGame.Extended`) or see a name change in the future (probably just drop to `Extended`) to house these features. 
- Development is focused on creating a prototype of a game called [`VectorVenture`](https://github.com/craftworkgames/VectorVenture), a 2D neon-shooter, which is similar to the current space game demo. Moving forward the demos/samples will no longer be in this repository and instead be in there own independent repositories. Each of these games is an early prototype (un-finished & un-polished). Plans are to create a prototype for each "genre/sub-genre" of video games.


lithiumtoast:
Update after diving into Veldrid:

- Geometry shaders: are interesting but seem to not be practical for game development. It is known that are inconsistencies of performance across hardware/driver vendors for geometry shaders, Metal flat out right doesn't support geometry shaders, and clever tricks with vertex, fragment, compute, and tessellation shaders make most, if not all, use cases of geometry shaders irrelevant. For these reasons, geometry shaders will not be supported with Ankura.

- Compute shaders: OpenGL Core 3.3 profile does not support storage buffers (SSBO). Storage buffers are introduced in OpenGL 4.3. Apple only supports OpenGL up to 4.1. Since one the primary use cases for storage buffers to store the output of parallel compuations from the GPU from either a compute or geometry shader, no storage buffers makes the case for compute shaders very weak. Without geometry shaders the case for no compute shaders is not acceptable, thus a decision has to be made. The decision is that OpenGL will be available only in a "legacy" graphics profile in Ankura. This would be similar to "Reach" vs "HiDef" in XNA where some capabilities or features are not available in the lower profile. Compute shaders, and other new GPU capabilities not found in MonoGame v3.8, will be usable with the "modern" graphics profile only. This would probably mean that only Vulkan, Direct3D12, and Metal would be the back-end APIs that would support the modern feature set such as compute and tesselation.

- Tessellation shaders: Metal, Direct3D12, and Vulkan have different pipeline models for tessellation. Direct3D12 and Vulkan have very similar pipeline models based around a an extended graphics pipeline which could in theory be abstracted the same way. Metal has a very different pipeline model which is based around a compute pipeline. This has limitations and pitfalls in a direct comparision. However, if a cross-platform higher API is to be exposed for tessellation it would seem that Direct3D12 and Vulkan would need to emulate Metal's pipeline for tesselation or vice-versa. I would rather wait for WebGPU to advance on this topic as their goals of creating an abstraction on top of Vulkan, Direct3D12, and Metal are very much similiar to the goals of Ankura and MonoGame in general. Also if MonoGame is to support the web as a platform target then WebGPU is very much going to be something that will need to be supported anyways. There latest docs can be found at https://gpuweb.github.io/gpuweb/#intro, and their GitHub at https://github.com/gpuweb/gpuweb.

lithiumtoast:
Moving forward, Ankura will generally follow the WebGPU spec unless otherwise specified.

For information about basically any topic such as storage buffers, tesselation, etc, can be found here: https://github.com/gpuweb/gpuweb/search?q=investigation%3A&type=Issues

lithiumtoast:
This project is on hold:

- I decided to use the Zig language (https://ziglang.org) for Ankura. Why? Because I do not find C/C++ fun; I do not want to maintain a C/C++ library. However, the current version of Zig has issues with platform invoke. Expected to be fixed in version 0.8 of Zig. See https://github.com/ziglang/zig/issues/6092

Good news:

- Time will be re-directed to be working with MonoGame. I have a prototype for a new collisions NuGet package in the works. More news soon.

lithiumtoast:
OK the project is back on; I found work-arounds for Zig.

I'm doing the work here: https://github.com/lithiumtoast/Ankura

lithiumtoast:
I'm closing this. I will write a blog post this weekend about the details of what's up with version 4.0.

PR #699, opened 2020-10-24
Fix: FadeTransition glitch

topnik-code:
The FadeTransition had visual glitches when combined with a BoxingViewportAdapter. This happened because the viewport's offset was applied twice.


lithiumtoast:
@topnik-code Thanks.

PR #702, opened 2020-11-09
Adjust DrawLine origin offset for thickness

atom-sgt:
The origin of the rectangle used by DrawLine was hard-coded to 0.5f.  This is correct for the default thickness value, but should account for the specified thickness instead.


atom-sgt:
Further testing revealed problems

PR #708, opened 2020-11-27
Added Editor config

LokiMidgard:
My complete code was full of warnings because I have different standard styles then this Project.

I added a `.editorconfig` that has the style of this source code (at least for the things that were different then mine).

Maybe it would also make sense to activate warnings as errors on build, then ci could check if a pull request don't mix styles.


lithiumtoast:
Could you rebase?

Also it says here that `.editorconfig` only works for Visual Studio 2019 16.8 and later. I guess this is okay, but not really sure if this is a deal-breaker for developers downstream. Technology moves fast; I would rather just accept it and move on.

LokiMidgard:
I used it before. but maybe I had an extension installed. And since it is only a file, it should just be ignored if its not supported.

I'll rebase late

Source/.editorconfig
```diff
@@ -0,0 +1,34 @@
+﻿root = true
```
lithiumtoast, line 1: Could you pleaase move `.editorconfig` file to the folder `src/dotnet/` instead of `Source/`

lithiumtoast:
1 check failed because apparently secrets can't be read for actions of forked repositories: `Secrets are not passed to workflows that are triggered by a pull request from a fork.` This is sad because it means pull request packages can't be deployed to my public feed on MyGet.org or NuGet.org. I'll need to modify the automated workflow to skip the deploy step for pull requests.

PR #710, opened 2020-11-29
ShapeF refactoring

LokiMidgard:
I started converting IShapeF methods to use Generics.

Also added a Analizer to warn about usage of IShapeF.


lithiumtoast:
Thanks for the effort.

I think I would rather just delete `IShapeF` all together. I'll think it over when I move to use Ankura.

LokiMidgard:
I'll keep this open untill then.

*Ankura looks very interresting*

lithiumtoast:
Too much going on in this PR. Feel free to open another one :)

Issue #712, opened 2020-12-13
Tiled TileMapImporter tileset source relative pathing


merthsoft:
Current, if you have a `.tmx` file with a `source` tag that uses a relative path, for example `<tileset firstgid="1" source="../Tilesets/Tileset_16x16_1.tsx"/>`, the importer will error out that the path could not be found, using the full combined path, without resolving the relative pathing. tiled is able to open the map file with no issue. For a fix I believe adding a `Path.GetFullPath` when determining the source will fix it, per my own testing. I can make a PR for this.

lithiumtoast:
Sure a PR is always welcome :)

lithiumtoast:
Fix by #713

PR #713, opened 2020-12-14
Resolve relative paths for tileset sources in tiled map importer

merthsoft:
Fixes issue #712


lithiumtoast:
Thanks 😃 

Could you rebase? I pushed a small commit that should fix the check for deploying.

merthsoft:
actually we should hold off on this. I'm getting some strange results while debugging now that I need to clear up.

merthsoft:
Ok never mind it seems ok! I've merged dev

lithiumtoast:
The last job in the workflow fails because forks don't have access to GitHub secrets. I need to look into this more...

Otherwise, great!

PR #717, opened 2020-12-23
Fix ParticleEmitters always using same random seed

mikeparker:
### Problem
Right now, every `ParticleEmitter` generates the same random spread of particles. This isn't so noticeable with lots of particles but with a low number (e.g. 5 particles in a spurt) it's very noticeable.

This is because the default `FastRandom` constructor initialises with a seed of `1`.

### Fix
This simply initialises the `FastRandom` with a random seed used by generating a new guid. We must use `Math.Abs` because FastRandom only accepts positive integers for the seed.


mikeparker:
Looks like `Profile` class has the same problem. I'm getting identical patterns from profiles.

lithiumtoast:
> Looks like Profile class has the same problem. I'm getting identical patterns from profiles.

If you figure it out I be open for a PR :)

PR #721, opened 2021-01-29
Update TiledMapLayerModelBuilder.cs

kryzp:
Fixed an issue todo with the tiles being about 1/2 pixel offset.


lithiumtoast:
Thanks!

Issue #723, opened 2021-02-08
Entity.Detach() method doesn't invoke EntityChanged until next Update of EntityManager


SjaakAlvarez:
When calling Detach() to remove a component, the entity is not updated until the next Update of the EntityManager.

For example:

When I Detach a component in a EntityProcessingSystem, the component bit is still set in the next EntityDrawSystem Update and the entity is still part of the ActiveEntities which might crash the code.

A quick and dirty solution is to modify EntityManager.OnComponentsChanged to:

```
private void OnComponentsChanged(int entityId)
{
    _changedEntities.Add(entityId);
    _entityToComponentBits[entityId] = _componentManager.CreateComponentBits(entityId);
    EntityChanged?.Invoke(entityId);
}
```

AristurtleDev:
This was resolved in #724

PR #724, opened 2021-02-08
Entity.Detach() method doesn't invoke EntityChanged until next Update of EntityManager #723

SjaakAlvarez:


lithiumtoast:
Thanks!

PR #733, opened 2021-04-01
Include documentation in packages

Apostolique:
This will include the XML docs in the next release.


src/dotnet/MonoGame.Extended.Collisions/MonoGame.Extended.Collisions.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
Emersont1, line 1: It's probably worth trying to undo these changes and then have it squash-merged to minimise chaanges to the code, especially as it seems only a whitespace change

PR #735, opened 2021-06-03
Replace Color.TransparentBlack with Color.Transparent

Pizt0lmnk:
`Color.TransparentBlack` has been removed in Monogame.
https://github.com/MonoGame/MonoGame/commit/c6d916d6983526900869d5e2f189471224f06c14

When using a development version of mgcb-editor (3.8.1.1983-develop) I was unable to build a .tmx file because of the usage of `Color.TransparentBlack` in `TiledMapImageContent.cs`.

I have replaced it with `Color.Tranparent` which is the replacement and has the same value.

### DIFF
--- a/src/dotnet/MonoGame.Extended.Tiled/Serialization/TiledMapImageContent.cs
+++ b/src/dotnet/MonoGame.Extended.Tiled/Serialization/TiledMapImageContent.cs
@@ -29,7 +29,7 @@ namespace MonoGame.Extended.Tiled.Serialization
 		[XmlIgnore]
 		public Color TransparentColor
 		{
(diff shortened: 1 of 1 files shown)


lithiumtoast:
Thanks.

brenjt:
It looks like this change is missing from the latest build. When is the next build supposed to be released? I can't compile tmx files until then

lithiumtoast:
@brenjt Latest builds are automatically deployed to MyGet. E.g. https://www.myget.org/feed/lithiumtoast/package/nuget/MonoGame.Extended/3.9.0-alpha0079.

alialdibouni:
What file did you edit to add `Color.Transparent`?

PR #755, opened 2022-01-19
TryGetTile no longer returns a value if width is outside the map

LokiMidgard:
when calling `TryGetTile` on `TileMapTileLayer` and using a `y`  coordinate greater then the height the method returns `false`.
Doing the same with `x` while `y` is not at maximum will return `true`.

This will check the `x` used and returns false if it is over the width.


lithiumtoast:
Thanks!

PR #770, opened 2022-06-22
Fixed issue when GetState() call rewrites previous states.

KRC2000:
There is a bug In the `GetState()` method of extended keyboard and mouse. 
Every time `GetState()` gets called - previous state get overwritten with current. If done more then one time - previous state gets overwritten and lost.
https://github.com/craftworkgames/MonoGame.Extended/blob/c90365211eaae3ed092d748230f4cbebe490d1b9/src/cs/MonoGame.Extended.Input/KeyboardExtended.cs#L11-L16
The only way to go is to request state only one time, save it in the variable and then use it, which is unhandy.
I think state swap should be performed only once per tick, while allowing multiple `GetState()` calls, so I propose to add `Refresh()` method, that will be called by user at the start of the tick, before any input checks.

### DIFF
--- a/src/cs/MonoGame.Extended.Input/KeyboardExtended.cs
+++ b/src/cs/MonoGame.Extended.Input/KeyboardExtended.cs
@@ -10,9 +10,13 @@ namespace MonoGame.Extended.Input
 
         public static KeyboardStateExtended GetState()
         {
+            return new KeyboardStateExtended(_currentKeyboardState, _previousKeyboardState);
+        }
+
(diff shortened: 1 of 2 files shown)


KRC2000:
Here is PR for according documentation update: [https://github.com/craftworkgames/craftworkgames.github.io/pull/47](url)

lithiumtoast (approved):

lithiumtoast:
Thanks!

PR #776, opened 2022-07-06
Tiled: Support class attribute as alias for object types

carlfriess:
As of Tiled version 1.9.0 the `type` attribute of objects has been replaced by the `class` attribute, as is reflected by the documentation [here](https://doc.mapeditor.org/en/stable/manual/custom-properties/#custom-classes). This change prefers the `class` attribute and falls back to the `type` attribute. For compatibility reasons I haven't updated the name of the `Type` field in [`TiledMapObject`](https://github.com/craftworkgames/MonoGame.Extended/blob/9a539d0517b9299a3e381547d47c6a2771f34bbe/src/cs/MonoGame.Extended.Tiled/TiledMapObject.cs#L22).


lithiumtoast:
Thanks

PR #782, opened 2022-08-09
Fix crash at loading AnimationSprite on Android.

garakutanokiseki:
If I use AnimationSprite on Android, App happened exception 'TitleContainer.OpenStream requires a relative path' at TextureAtlasJsonConverter.
MonoGame's content loader need relative path on Android.
This fix is load content by relative path if happened exception.


lithiumtoast:
Thanks!

PR #784, opened 2022-08-18
Uncommented and converted unit tests to XUnit

toore:


toore:
Noticed some strange behaviour related to matrix calculations and went into the code base. Most of the test cases are commented. I don't know the reason, but for confidence reasons, IHMO I think the tests should be enabled all the time. Before committing any changes to the production code I decided to contribute and make sure some tests can be run.

PR #785, opened 2022-08-27
Improvement: enabled and ported tests to xunit

toore:
For confidence and regression matter, I've enabled (uncommented) and ported some tests to xunit. _IMHO_ the tests should be enabled and running (successfully) all the time.


toore:
_IMO_, I think the failing checks are false positives. Looking at the other PRs, there are more checks in this PR (more targets built).

Is there any possibility to re-trigger this build?

PR #787, opened 2022-08-28
Fix rectangle transform

### LINKED ISSUE #786: Transforming a RectangleF does not calculate bounding rectangle from the rotated original as intended
```RectangleF.Transform``` does not compute the axis-aligned bounding rectangle as cited by the comments. I think that the code was implemented incorrectly by mistake. The desired result of transforming a rectangle is creating a new axis-aligned rectangle, that bounds the calculated rotated one. It will typically grow.

The desired result can be examplified by the following:
| Original rectangle (axis-aligned) | Original rotated ~45 degrees (CW) and computed bounding rectangle (also axis-aligned) |
|-|-|
|image|image|

The results of the current ```RectangleF.Transform```, is that the new rectangle's center point is correctly computed, but the new extents of the rectangle is typically way off. (The rotated extent is not computed according to documentation, and the translation matrix information is multiplied in, when extents don't actually need that info since they are relative the center point).

The internal method ```PrimitivesHelper.TransformRectangle``` is used by a ```Transform``` method in both ```BoundingRectangle``` and ```RectangleF``` so both types probably have the same problem.

toore:
Resolves #786.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 
 # Rider
 .idea/
+*.user
 
 # macOS
 .DS_Store
--- a/src/cs/MonoGame.Extended/Math/PrimitivesHelper.cs
+++ b/src/cs/MonoGame.Extended/Math/PrimitivesHelper.cs
@@ -65,10 +65,10 @@ namespace MonoGame.Extended
             // Real-Time Collision Detection, Christer Ericson, 2005. Chapter 4.2; Bounding Volumes - Axis-aligned Bounding Boxes (AABBs). pg 86-87
 
             center = transformMatrix.Transform(center);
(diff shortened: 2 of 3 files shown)


lithiumtoast (approved):
Page 87 http://www.r-5.org/files/books/computers/algo-list/realtime-3d/Christer_Ericson-Real-Time_Collision_Detection-EN.pdf.

Screen Shot 2022-09-19 at 10 07 53 PM

Indeed I did not have this correct if unrolling the first loop for X and Y (only two dimensions in 2D instead of 3 in 3D), then unroll the second loop. The second loop is addition of `a.r[j]` which means yes, it should be `HalfWidth = xRadius * Abs(M11) + yRadius * Abs(M12)`. Similar for `HalfHeight` but with `i=1`.

PR #789, opened 2022-09-24
Rebuilt effects with mgfxc 3.8.1.303

mattj1:
Required in order for features such as the Tiled map renderer to work with Monogame 3.8.1


lithiumtoast:
Thanks!

PR #795, opened 2022-11-19
OnCollision are called even after objects involved are moved and not intersecting

### LINKED ISSUE #591: Prevent pushing dynamic bodies
Hey guys,

there should be a solution to prevent pushing a dynamic body with an other dynamic body.
They should collide, but not pushable. In many games you need e.a. npcs which moving around a a town, but you must not push them or they should push you. But if you ran into one or they ran into you, the character must stop / flow around.

### LINKED ISSUE #796: CollisionManager reports objects as colliding, even after object(s) has been moved out of collision bounds
_Background:_
When using the `CollisionManager` and updating/moving the bounds of one of the objects out-of-band of the `OnCollision` callback, the bounds used by `CollisionManager` is not updated properly. This will have the effect that `OnCollision` is called without the objects colliding.

_Scenario:_
After a collision has been detected correctly, the collliding object is moved out of bounds of collision, out-of-band of the actual collision management. `CollisionManager` is still reporting the objects as being colliding, since the bounds are cached in `CollisionManager` and not updated properly.

The "opposite" is also failing. `CollisionManager` determines two non-intersecting objects as not colliding. If one of the objects  is moved and the objects intersects, `CollisionManager` will not use the updated bounds in the collision testing.

_Suggested solution:_
Do not cache bounds, always use the **actual** bounds.

toore:
Resolve #796 

Potentially resolve #591 

I changed the algorithm, so that the **actual** bounds are always used. I don't know if this will cause any unwanted side-effects.

A couple of unit tests were added to prove the case.

### DIFF
--- a/src/cs/MonoGame.Extended.Collisions/QuadTree/CollisionComponent.cs
+++ b/src/cs/MonoGame.Extended.Collisions/QuadTree/CollisionComponent.cs
@@ -54,7 +54,6 @@ namespace MonoGame.Extended.Collisions
                     };
 
                     target.OnCollision(collisionInfo);
-                    value.Bounds = value.Target.Bounds;
                 }
                 _collisionTree.Insert(value);
             }
--- a/src/cs/MonoGame.Extended.Collisions/QuadTree/CollisionComponent.cs
+++ b/src/cs/MonoGame.Extended.Collisions/QuadTree/CollisionComponent.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
 using Microsoft.Xna.Framework;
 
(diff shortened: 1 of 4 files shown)


florentmontesano:
I had the exact same issue as https://github.com/craftworkgames/MonoGame.Extended/issues/796 and fixed it by moving the `value.Bounds = value.Target.Bounds;`  statement one line below, to put it out of the loop in `CollisionComponent#Update`.

Your solution is much better since there is no more need to update the actual bounds.

I tested this fix in my game and it looks good.

Please merge! 👍

Gandifil:
Ok, I have check this solution, and will merge it.

@florentmontesano But, I need fully refactor all code about collision system. Remember, I will rewrite 
this in next releases.

PR #799, opened 2023-02-06
Add power of two values to MouseButton

dezoitodemaio:
Add power of two values to `MouseButton` so `MouseButton.HasFlags` can be used.


Gandifil:
Yes, more than one mouse buttons can be pressed at the moment and attribute "Flags" don't setup enum values.
But remember, events and `MouseStateExtended` use this enum as standalone.

PR #801, opened 2023-03-08
Tiled parallax factor

### LINKED ISSUE #800: Tiled doesn't support fields ParallaxFactor for layers.
_Summary:_
Tiled has a feature for creating "background" and "foreground" layers. For it layers has field ParallaxFactor. This settings influences rendering of Tiled Map.

_Proposal:_

- Add additional fields in layer class.
- Add modifying a view matrix in TiledMapRenderer, if ParallaxFactor or offsets vectors are not default.

Gandifil:
Add implements of parallaxFactor for Tiled Map.

1. Reading parallaxFactor from Tiled files.
2. Writing in XNB.
3. Reading from XNB in layers objects.
4. Use parallaxFactor if not One when render Tiled Map.

This should resolve #800


Gandifil:
#800

Issue #803, opened 2023-03-31
Looping SpriteSheetAnimations incorrectly go to a Completed state if the elapsed game time is greater than AnimationDuration


DavidFidge:
In my game I have short two framed animations which I want to loop indefinitely.  Sometimes the time between updates are greater than the total animation duration * 2.  In these situations, line 65 of SpriteSheetAnimation does not subtract enough time off of CurrentTime and thus the animation incorrectly gets set to Completed and the animations stop working.

Below is a basic unit test to demonstrate the issue.  The last two theories (seconds of 4 and 5) fail because they result in CurrentTime still being >= AnimationDuration and thus IsComplete gets set to true and the animations stop playing.

```
        [Theory]
        [InlineData(0, 0)]
        [InlineData(0, 0.9f)]
        [InlineData(1, 1f)]
        [InlineData(1, 1.9f)]
        [InlineData(0, 2f)]
        [InlineData(0, 2.9f)]
        [InlineData(1, 3f)]
        [InlineData(0, 4f)]
        [InlineData(1, 5f)]
        public void Looping_SpriteSheetAnimation_Should_Return_Correct_Frame_And_Not_Complete(int expectedTextureRegionIndex, float time)
        {
            var textureRegion2D1 = new TextureRegion2D("Region 1", null, new Rectangle());
            var textureRegion2D2 = new TextureRegion2D("Region 2", null, new Rectangle());

            var textureRegions = new[] { textureRegion2D1, textureRegion2D2 };

            var spriteSheetAnimationData = new SpriteSheetAnimationData(
                new[] { 0, 1 },
                1,
                true,
                false
            );

            var spriteSheetAnimation = new SpriteSheetAnimation("Test", textureRegions, spriteSheetAnimationData);

            spriteSheetAnimation.Play();
            var gameTime = new GameTime(TimeSpan.Zero, TimeSpan.FromSeconds(time));

            spriteSheetAnimation.Update(gameTime);

            Assert.Equal(textureRegions[expectedTextureRegionIndex], spriteSheetAnimation.CurrentFrame);
            Assert.False(spriteSheetAnimation.IsComplete);
        }
```

The issue also happens when the time between updates is a little over the Animation Duration.  In these circumstances, the CurrentTime will accumulate the residual amount of the result of CurrentTime - AnimationDuration and the animation will eventually incorrectly go to a Completed state once the residual amount causes CurrentTime -= AnimationDuration to still exceed AnimationDuration.

[...]

PR #806, opened 2023-03-31
Fix bugs in SpriteSheetAnimation

DavidFidge:
- Fix looping animations incorrectly completing when an update contains a large elapsed time.
- Ping pong waits until frame delay is complete on re-entry of the first frame before firing Complete rather than firing Complete immediately upon entering the frame.  This makes it consistent with a non ping pong animation which waits for the frame delay of the last frame before it fires Complete. 
- Ping Pong now finishes on the first frame rather than potentially finishing on a different frame if the elapsed time since last update was larger than the time remaining for the animation to become complete. 
- Add unit tests for SpriteSheetAnimation.

### DIFF
--- a/src/cs/MonoGame.Extended/Sprites/SpriteSheetAnimation.cs
+++ b/src/cs/MonoGame.Extended/Sprites/SpriteSheetAnimation.cs
@@ -41,7 +41,7 @@ namespace MonoGame.Extended.Sprites
         public new bool IsComplete => CurrentTime >= AnimationDuration;
 
         public float AnimationDuration => IsPingPong
-            ? (KeyFrames.Length*2 - 2)*FrameDuration
+            ? (KeyFrames.Length*2 - (IsLooping ? 2 : 1))*FrameDuration
             : KeyFrames.Length*FrameDuration;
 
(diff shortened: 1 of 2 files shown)


Gandifil:
> * Fix looping animations incorrectly completing when an update contains a large elapsed time.
> * Ping pong waits until frame delay is complete on re-entry of the first frame before firing Complete rather than firing Complete immediately upon entering the frame.  This makes it consistent with a non ping pong animation which waits for the frame delay of the last frame before it fires Complete.
> * Ping Pong now finishes on the first frame rather than potentially finishing on a different frame if the elapsed time since last update was larger than the time remaining for the animation to become complete.
> * Add unit tests for SpriteSheetAnimation.

Do I correctly understand you? You mean that:

There are 3 frame. Order of rendering must be: 0, 1, 2, 1, 0 (without duplicating frame number 2). For looping: 0, 1, 2, 1, 0, 1, 2, 1 and etc (without duplicating frame number 0 and 2).

Gandifil (approved):

Gandifil:
Resolve #803

DavidFidge:
Thanks for taking the pull request!  That's correct, that's how the original code worked (this pull request just fixes all the bugs with it).

PR #838, opened 2024-02-17
Throw exception in ViewportAdapter constructor if the GraphicsDevice is null

safoster88:
If the viewport adapter is initialized in the constructor of game, GraphicsDevice will be null. 

There is no check for this, so the null ref only occurs when trying to call GetViewMatrix(). 

This PR aims to improve the public API.

### DIFF
--- a/src/cs/MonoGame.Extended/ViewportAdapters/Exceptions/NullGraphicsDeviceException.cs
+++ b/src/cs/MonoGame.Extended/ViewportAdapters/Exceptions/NullGraphicsDeviceException.cs
@@ -1,0 +1,18 @@
+﻿namespace MonoGame.Extended.ViewportAdapters.Exceptions;
+
+using System;
+using Microsoft.Xna.Framework.Graphics;
+
+/// <summary>
+/// Thrown when the <see cref="GraphicsDevice"/> is null
+/// </summary>
+public class NullGraphicsDeviceException : Exception
+{
+    /// <summary>
+    /// Thrown when the <see cref="GraphicsDevice"/> is null
+    /// </summary>
+    /// <param name="message">Exception message</param>
+    public NullGraphicsDeviceException(string message) : base(message)
+    {
+    }
+}
--- a/src/cs/MonoGame.Extended/ViewportAdapters/ViewportAdapter.cs
+++ b/src/cs/MonoGame.Extended/ViewportAdapters/ViewportAdapter.cs
@@ -4,10 +4,17 @@ using Microsoft.Xna.Framework.Graphics;
 
 namespace MonoGame.Extended.ViewportAdapters
 {
+    using MonoGame.Extended.ViewportAdapters.Exceptions;
+
     public abstract class ViewportAdapter : IDisposable
     {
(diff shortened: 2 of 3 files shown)


Gandifil:
Thank you for attention!

src/cs/MonoGame.Extended/ViewportAdapters/ViewportAdapter.cs
```diff
@@ -4,10 +4,17 @@
...
 {
+    using MonoGame.Extended.ViewportAdapters.Exceptions;
+
     public abstract class ViewportAdapter : IDisposable
     {
         protected ViewportAdapter(GraphicsDevice graphicsDevice)
         {
+            if (graphicsDevice is null)
+            {
+                throw new NullGraphicsDeviceException($"Attempted to construct {nameof(ViewportAdapter)} but {nameof(GraphicsDevice)} is null");
```
AristurtleDev, line 15: This should just throw an `ArgumentNullException` since the issue here is that the *argument* is null, not that the graphics device itself is null.

Please use the new method for checking

```cs
ArgumentNullException.ThrowIfNull(graphicsDevice);
```

src/cs/Tests/MonoGame.Extended.Tests/ViewportAdapters/ViewportAdapterTests.cs
```diff
@@ -0,0 +1,35 @@
...
+using Xunit;
+
+public class ViewportAdapterTests
+{
+    [Fact]
+    public void Constructor_ShouldThrowNullGraphicsDeviceException_IfANullGraphicsDeviceIsPassed()
+    {
+        var act = () => new TestViewportAdapter(null);
+
+        Assert.Throws<NullGraphicsDeviceException>(act);
```
AristurtleDev, line 16: After making the suggested change of using `ArgumentNullException` this should be updated to check for that exception type.

src/cs/MonoGame.Extended/ViewportAdapters/Exceptions/NullGraphicsDeviceException.cs
AristurtleDev, line 1: This custom exception is not needed.  The problem being solved is that the graphics device argument for the method is null, not that the GraphicsDevice itself is null.   This can be removed and the exception changed to use `ArgumentNullException` to better communicate to consumers that the argument being null is the issue.

AristurtleDev (changes requested):
Thank you for the PR @safoster88 
Please make the requested changes and then it can be merged.

Cev0li:
Is this resolved? I'd like to take a shot at my first PR if it's not!

kaltinril:
How long do you let PR's sit with no response from the person who created it before they are fixed by someone else, or, canceled/deleted?

AristurtleDev:
> Is this resolved? I'd like to take a shot at my first PR if it's not!

Given that the original PR was asked to be updated by both @Gandifil  and I with no response, if you would like to take this @Cev0li, it's be a good first contribution.

AristurtleDev:
Closed PR due to lack of response from OP

PR #841, opened 2024-04-09
tiled content pipeline bugfixes & new fields

KatDevsGames:
added support for reading type field from TiledMapPropertyContent (Tiled lets you set it...)
added support for repeatX/repeatY fields from Tiled image layers property (renderer still does not support this yet)

value read bugfix @TiledMapProperties.cs:10 - this got introduced when you pulled my previous pr (sorry)

### DIFF
--- a/src/cs/MonoGame.Extended.Collisions/CollisionComponent.cs
+++ b/src/cs/MonoGame.Extended.Collisions/CollisionComponent.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using Microsoft.Xna.Framework;
 using MonoGame.Extended.Collisions.Layers;
 using MonoGame.Extended.Collisions.QuadTree;
+using Newtonsoft.Json.Linq;
 
 namespace MonoGame.Extended.Collisions
 {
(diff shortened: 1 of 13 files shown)


tigurx:
Hi @KatDevsGames ,

Your timing is absolutely impeccable. The RepeatX and RepeatY for the image layer in tiled is actually something I started looking into last night when trying to implement a parallax image layer for my project. Do you have any idea on how to approach this in regard to the renderer? 

Please let me know if I can be of any assistance. I am not exactly sure how to approach it, but wouldn't mind helping tackle that if it is something that proves difficult.

- James

Gandifil:
Hi @KatDevsGames and @tigurx !

I have watched PR. 

- TiledMap: all good. What does do `RepeatX `and `RepeatY`? Duplicating image N times horizontal and vertical? If yes, I don't think implementation of this is too big trouble.
- Collision: thanks for `Contains` method, I forgot about this. But I don't understand `Cast`. I mean it must be on another logic layer. Collision is not physics (but they are linked of course). `CollisionComponent` does not have to move actors and does not know about "hit". May be, for you better to do component which moves actors and asks `CollisionComponent` about collisions. Other question, if you need to update collisions for sequence of object (not all, not one). Is it true?

In future, please, make PR separately by theme.

tigurx:
@Gandifil , For RepeatX & RepeatY -- Yes, it just repeats the image N times along the respected axis like you suggested.

https://github.com/craftworkgames/MonoGame.Extended/assets/16768537/ddffcdb6-4957-4b7b-9a36-2668e6605db1

src/cs/MonoGame.Extended.Collisions/CollisionComponent.cs
```diff
@@ -97,6 +98,50 @@ public override void Update(GameTime gameTime)
...
+        /// <summary>
+        /// Cast an actor's colliders in a direction and check for collisions.
+        /// </summary>
+        /// <param name="actors">The list of actor colliders to cast.</param>
+        /// <param name="move">Normalized vector representing the direction to cast each shape.</param>
+        /// <param name="hitBuffer">Buffer to receive results.</param>
+        /// <param name="distance">Maximum distance over which to cast the Collider(s).</param>
+        public int Cast(IEnumerable<ICollisionActor> actors, Vector2 move, List<CollisionEventArgs> hitBuffer, float distance)
+        {
+            hitBuffer.Clear();
```
kaltinril, line 110: Does it matter that hitBuffer.Clear() isn't called in the 2nd Cast?  What if someone calls the 2nd public method instead, then the hitBuffer won't get cleared right?

src/cs/MonoGame.Extended.Collisions/CollisionComponent.cs
```diff
@@ -6,6 +6,7 @@
 using Microsoft.Xna.Framework;
 using MonoGame.Extended.Collisions.Layers;
 using MonoGame.Extended.Collisions.QuadTree;
+using Newtonsoft.Json.Linq;
```
kaltinril, line 9: Is MonoGame.Extended trying to stay AOT compliant?

This article on MonoGame says that Newtonsoft json are not AOT compliant for console?  Also that Linq should be avoided?  Also, why Newtonsoft's Linq vs System.Linq?

https://docs.monogame.net/articles/preparing_for_consoles.html

src/cs/MonoGame.Extended.Tiled/TiledMapPropertyValue.cs
```diff
@@ -4,26 +4,20 @@ public class TiledMapPropertyValue
...
+
+    public TiledMapPropertyValue(string value, string type, string propertyType)
     {
         Value = value;
+        Type = type;
+        PropertyType = propertyType;
         Properties = new();
     }
 
-    public TiledMapPropertyValue(TiledMapProperties properties)
```
kaltinril, line 21: Was this removed for a reason?

kaltinril (commented):
Looks like Cast is added but not yet used.  I see the bug fix.  I see you also added 4 properties (type, propertyType, repeatX, repeatY).

I noticed you removed a constructor for TiledMapPropertyValue that used to allow passing in a TiledMapProperties instance.  Was this intended?

AristurtleDev:
@KatDevsGames thank you for the original submission.  Due to the prolonged period of no communication after asking for PR to be updated to resolve conflicts, I'm going to close this out.  If you would like to have it reopened and resolve the merge conflicts to get it pushed through, please let us know.

PR #845, opened 2024-04-23
Add usable extension to Camera.cs and also improve documentation a little

MikDal002:
Hello :) It is my first contribution here. I am trying to share things I've found useful :).

### DIFF
--- a/src/cs/MonoGame.Extended/Cameras/Camera.cs
+++ b/src/cs/MonoGame.Extended/Cameras/Camera.cs
@@ -4,6 +4,10 @@ namespace MonoGame.Extended
 {
     public abstract class Camera<T> where T : struct
     {
+        /// <summary>
+        /// Returns position of the camera in real world position.
+        /// Defaults to Vector2.Zero. 
+        /// </summary>
         public abstract T Position { get; set; }
         public abstract float Rotation { get; set; }
         public abstract float Zoom { get; set; }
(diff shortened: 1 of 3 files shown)


Gandifil:
Hi! Thanks for your attention!

It's good, but why you don't add method directly (without extension class)?

AristurtleDev:
Hi @MikDal002, first thank you for the PR submission. 

Looking at these changes, I'm not sure there's a valid use case to add this extension method.  The `Camera<T>` class is a generic, which means any implementation of it is not always going to be a `Vector2` for the `T` type.  The provided implementation in MonoGame.Extended (the `OrthographicCamera`) does happen to use `Vector2` for the generic type, however consumers may choose to use any valid `struct` such as `Vector3`, or `Point`.  

By adding this extension method, we would also need to add additional methods to support each scenario, which means MonoGame.Extended is now anticipating the needs of the consumer which goes against the [Not Future Proof](https://www.monogameextended.net/docs/about/principles/#not-future-proof) principle.  Additionally, this extension method adds very little utility since the consumer can just do `orthCamera.ScreenToWorld(point.ToVector2())` and achieve the same functionality, even though they have to use the `Point.ToVector2()` call, which is all this is doing internally already 

Instead consumers that derive from `Camera<T>` should implement the `T ScreenToWorld(T)` method within the class they create to fit that use-case/scenario.

PR #846, opened 2024-04-28
Fixes incorrect rendering of flipped animation tiles

tigurx:
This PR fixes the incorrect rendering of flipped tiles that are animated. 

The reason this bug occurred was that when the vertex buffer got updated with the next animation frame texture, it pulled the default orientation instead of the flipped version.

I opted to add a dictionary of the flipped vectors to the animation frames. This dictionary only gets loaded with the necessary orientations of flipped vectors needed for rendering. I thought this was a better option than recalculating all of the necessary flipping in the update loop. This adds a bit more memory overhead, but should run faster.

Feel free to reach out if you have any questions. I have some footage of the before and after I can show if you are interested.


src/cs/MonoGame.Extended.Tiled/Renderers/TiledMapAnimatedLayerModel.cs
```diff
@@ -4,15 +4,17 @@ namespace MonoGame.Extended.Tiled.Renderers
...
             : base(graphicsDevice, texture, vertices, indices)
         {
             Vertices = vertices;
             AnimatedTilesetTiles = animatedTilesetTiles;
+            AnimatedTilesetFlipFlags = animatedTilesetTileFlipFlags;
         }
 
         public VertexPositionTexture[] Vertices { get; }
         public TiledMapTilesetAnimatedTile[] AnimatedTilesetTiles { get; }
+        public TiledMapTileFlipFlags[] AnimatedTilesetFlipFlags { get; }
```
AristurtleDev, line 18: Could this be changed to be a `ReadOnlySpan<TiledMapTileFliPFlags>` instead.  This is a get-only property but that does not make the array it returns immutable. 

To do this you would need to add a backing field and return that field as the `ReadOnlySpan`

```cs
private readonly TiledMapTileFlipFlags[] _animatedTilesetFlipFlags;

public ReadOnlySpan<TiledMapTileFlipFlags> AnimatedtilesetFlipFlags => _animatedTilesetFlipFlags;
```

tigurx: I implemented the above. I'm working on changing over the renderer to utilize the span now. I am still pretty new to spans; would this be the fastest way to iterate a span alongside the array of animated tiles?

See code snippet from **TiledMapRenderer.cs** below followed by the two lines where I utilize the span:

```C#
private static unsafe void UpdateAnimatedLayerModels(IEnumerable<TiledMapAnimatedLayerModel> animatedLayerModels)
{
	foreach (var animatedModel in animatedLayerModels)
	{
		// update the texture coordinates for each animated tile
		fixed (VertexPositionTexture* fixedVerticesPointer = animatedModel.Vertices)
		{
			var verticesPointer = fixedVerticesPointer;
			var slice = animatedModel.AnimatedTilesetFlipFlags.Slice(0, animatedModel.AnimatedTilesetFlipFlags.Length);
			for (int i = 0; i < animatedModel.AnimatedTilesetTiles.Length; i++)
			{                       
				var currentFrameTextureCoordinates = animatedModel.AnimatedTilesetTiles[i].CurrentAnimationFrame.GetTextureCoordinates(slice[i]);

				// ReSharper disable ArrangeRedundantParentheses
				(*verticesPointer++).TextureCoordinate = currentFrameTextureCoordinates[0];
				(*verticesPointer++).TextureCoordinate = currentFrameTextureCoordinates[1];
				(*verticesPointer++).TextureCoordinate = currentFrameTextureCoordinates[2];
				(*verticesPointer++).TextureCoordinate = currentFrameTextureCoordinates[3];
				// ReSharper restore ArrangeRedundantParentheses
			}
		}

		// copy (upload) the updated vertices to the GPU's memory
		animatedModel.VertexBuffer.SetData(animatedModel.Vertices, 0, animatedModel.Vertices.Length);
	}
}
```

**The first line splits the span into equal slices:**
var slice = animatedModel.AnimatedTilesetFlipFlags.Slice(0, animatedModel.AnimatedTilesetFlipFlags.Length);

**I am able to access the necessary member when I need to get the texture coordinates while I am iterating through:**
animatedModel.AnimatedTilesetTiles[i].CurrentAnimationFrame.GetTextureCoordinates(slice[i]);

Is there an alternative method of writing this that would be faster than the method above?

Best regards,

James

AristurtleDev: You don't have to slice it, you should be able to just adjust the property as mentioned to be a `ReadOnlySpan<T>` and then use the `for` loop as usual.

AristurtleDev: To clarify, `Span<T>` and `ReadOnlySpan<T>` under the hood are just a struct with a pointer to the array in memory and the length.  So when working with them, you can work with them just like it was a normal array, like in a for loop, etc, but you get the performance advantage that it's using pointers under the hood.

tigurx: Thank you for your explanation. I made the necessary change to make this variable private and read only with a public `ReadOnlySpan<T>` property. I also removed the slice that I had proposed and access the span directly by index within the loop.

AristurtleDev (approved):

Issue #849, opened 2024-05-17
Multiple Test Failures When Running Test


AristurtleDev:
## Description
The tests are failing for multiple projects.  The following is a list of each project and which are currently passing.  In the sections below you can find the specific errors for why the tests are failing

- [x] MonoGame.Extended.Collisions.Tests
- [x] MonoGame.Extended.Content.Pipeline.Tests
- [x] MonoGame.Extended.Content.Pipeline.Tests.Tiled
- [x] MonoGame.Extended.Entities.Tests
- [x] MonoGame.Extended.Gui.Tests (there are no tests here)
- [x] MonoGame.Extended.Tests
- [x] MonoGame.Extended.Tiled.Tests (there are no tests here)

### MonoGame.Extended.Collision.Tests Errors

<details>
<summary>Click to expand</summary>

```sh
aristurtle@aris MonoGame.Extended.Collisions.Tests % dotnet test
  Determining projects to restore...
  All projects are up-to-date for restore.
  MonoGame.Extended -> /Users/aristurtle/Desktop/MonoGame.Extended/src/cs/MonoGame.Extended/bin/Debug/net6.0/MonoGame.Extended.dll
  MonoGame.Extended.Collisions -> /Users/aristurtle/Desktop/MonoGame.Extended/src/cs/MonoGame.Extended.Collisions/bin/Debug/net6.0/MonoGame.Extended.Collisions.dll
  MonoGame.Extended.Collisions.Tests -> /Users/aristurtle/Desktop/MonoGame.Extended/src/cs/Tests/MonoGame.Extended.Collisions.Tests/bin/Debug/net6.0/MonoGame.Extended.Collisions.Tests.dll
Test run for /Users/aristurtle/Desktop/MonoGame.Extended/src/cs/Tests/MonoGame.Extended.Collisions.Tests/bin/Debug/net6.0/MonoGame.Extended.Collisions.Tests.dll (.NETCoreApp,Version=v6.0)
Microsoft (R) Test Execution Command Line Tool Version 17.8.0 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
The active test run was aborted. Reason: Test host process crashed : Stack overflow.
Repeat 726 times:
--------------------------------
   at MonoGame.Extended.Shape.Intersects(MonoGame.Extended.IShapeF, MonoGame.Extended.IShapeF)
--------------------------------
[...]

PR #854, opened 2024-05-17
Update Newtonsoft.Json.dll Dependency

### LINKED ISSUE #853: NewtonSoft.Json.dll Dependency Is Wrong Version
## Description
The `MonoGame.Extended.Content.Pipline` project relies on a `Newtonsoft.Json.dll` that is copied to the bin directory on build.  The .csproj references version 13.0.1 

https://github.com/craftworkgames/MonoGame.Extended/blob/4da89293110149a02c19bf4b179da5114c241f10/src/cs/MonoGame.Extended.Content.Pipeline/MonoGame.Extended.Content.Pipeline.csproj#L34

However the .dll in the directory is version 12.0.0

```sh
aristurtle@aris MonoGame.Extended.Content.Pipeline % monodis --assembly Newtonsoft.Json.dll
Assembly Table
Name:          Newtonsoft.Json
Hash Algoritm: 0x00008004
Version:       12.0.0.0
Flags:         0x00000001
PublicKey:     BlobPtr (0x00005fbf)
```

## Related
- #849

AristurtleDev:
## Description
This updates the `Newtonsoft.Json.dll` dependency in the `MonoGame.Extended.Content.Pipline` project.  The csproj itself references version 13.0.0.1 but the .dll dependency that is being copied on build was version 12.0.0.0

## Related Issues
- Closes #853 
- #849


lithiumtoast:
Is there a possibility of opening an issue to use `System.Text.Json`? It has better support with source generators which is better for AOT scenarios like consoles.

AristurtleDev:
> Is there a possibility of opening an issue to use `System.Text.Json`? It has better support with source generators which is better for performance and AOT scenarios like consoles.

Yep, can do

Issue #587, opened 2019-01-24
Matrix2 get scale incorrect


leftnoodle:
It returns incorrect values and doesn't take into account negative scale values. This can be seen when creating a matrix using CreateFrom with rotation = 0.296706 radians, scaleX =7 and scaleY = 4. The result from get scale is 6.796, 4.338.

I believe the correct calculation should be:

var scaleX = (float)Math.Sign(M11) * (float)Math.Sqrt(M11 * M11 + M12 * M12);
var scaleY = (float)Math.Sign(M22) * (float)Math.Sqrt(M21 * M21 + M22 * M22);

The get rotation also returns incorrect rotation values

craftworkgames:
Thanks for reporting the issue.

To be honest, I didn't write most of that code and I don't use it very often. So it's difficult to tell just from looking at your description what the problem is or if that's the correct solution.

To make this change I suggest we need some unit tests and / or find a good reference for the equations.

lithiumtoast:
I wrote the code. 

@leftnoodle You are correct it does not take into account negative scale values. My proposed solution.

```
public void Decompose(out Vector2 scale, out float rotation, out Vector2 translation)
{
     translation.X = this.M31;
     translation.Y = this.M32;

     float xs = (Math.Sign(M11 * M12) < 0) ? -1 : 1;
     float ys = (Math.Sign(M21 * M22) < 0) ? -1 : 1;

     scale.X = xs * (float)Math.Sqrt(M11 * M11 + M12 * M12);
     scale.Y = ys * (float)Math.Sqrt(M21 * M21 + M22 * M22);

     rotation = (float)Math.Atan2(M21, M11);
}
```

For two dimensions, the code can be chopped to get methods (properties) as the status-quo. However, for three dimensions [it's been discussed why this is not ideal](https://github.com/MonoGame/MonoGame/issues/5553). To keep consistency with MonoGame I propose marking the get properties, such as `Scale`, deprecated and rely on the new `Decompose` method. The similar pattern can be found with [`Matrix`](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Matrix.cs#L1454).

leftnoodle:
I think you will need to do the same as Matrix.cs and divide M21 and M11 by scale.Y and scale.X respectively to obtain the correct rotation?

I'm also not sure that will calculate the correct scale, I could be mistaken but if we take a standard rotation matrix with PI / 4.0 radians then we get:

[...]

PR #870, opened 2024-05-27
Replace `Matrix2` with `Matrix3x2`

AristurtleDev:
## Description
This pull request replaces the `Matrix2` struct with a new `Matrix3x2` struct.  This is not the same as the `System.Numerics.Matrix3x2`. 

## Why not use `System.Numerics`
I would love to have this switched to using the `System.Numerics.Matrix3x2` struct, however at this time it is currently to much work.  MonoGame itself does not use the `System.Numerics` types currently, so implementing that here would mean that conversions would need to be made between `Microsoft.Xna.Framework.Vector2` and `System.Numerics.Vector2` all over the place.  This would not only cause type ambiguity collisions within MonoGame.Extended, but also for consumers as well.

There is an open issue that can be tracked in the MonoGame repository for when MonoGame will switch to using `System.Numerics` https://github.com/MonoGame/MonoGame/issues/7204. When the base MonoGame performs that change, MonoGame.Extended will then be updated to reflect that.

## Issues Resolved
This PR also resolves issue #587 by correctly accounting negative scale.


AristurtleDev:
XML documentation needs to be added before PR is completed.

PR #881, opened 2024-05-31
Remove unnecessary dependencies to SharpDX and Content.Pipeline from the core library

nkast:
- remove unused SharpDX using
introduces in https://github.com/craftworkgames/MonoGame.Extended/pull/840

- remove Content.Pipeline reference from the core library
introduced in https://github.com/craftworkgames/MonoGame.Extended/pull/692


AristurtleDev (approved):
Thanks @nkast

PR #882, opened 2024-06-01
use System ZLibStream

nkast:
ZLibStream was added in net6

https://learn.microsoft.com/en-us/dotnet/api/system.io.compression.zlibstream?view=net-8.0


AristurtleDev (commented):
Thanks @nkast

PR #907, opened 2024-07-02
Add pitch support to Camera & OrthographicCamera

jameskellie:
This is a simple and unobtrusive addition to the Camera & OrthographicCamera classes that adds a pitch variable and implementation. The reason this is needed is because it allows the camera to better create a faux-3D or 2.5D perspective that can be rotated, zoomed, and pitched, as if it were a true 3D camera, while still maintaining that it is fundamentally 2D. The inclusion of this addition does not alter any existing functionality nor would it affect existing projects already using the OrthographicCamera class.


AristurtleDev (approved):
Thanks @jameskellie

PR #925, opened 2024-07-23
fix color extension bug

ValorZard:
Thanks @nkast 

from nkast
> Color in fna doesnt have a constructor that takes a packed uint value, actually it's private
> https://github.com/FNA-XNA/FNA/blob/31dab885817ddaece9758f480027373f0ad74bea/src/Color.cs#L1668
> There is a property PackedValue, (that implements the  IPackedVector<uint> interface.)


AristurtleDev (changes requested):

AristurtleDev (approved):
Thanks @ValorZard

PR #941, opened 2024-09-01
fix(Matrix3x2): decompose method

### LINKED ISSUE #935: Matrix3x2 Decompose function returns incorrect translation
Hello! It looks like the translation returned by Decompose is actually different than the Z vector, since it assigns both X and Y of the translation to M31 rather than M31 and M32 like I would expect. I assume this is a typo

https://github.com/craftworkgames/MonoGame.Extended/blob/develop/source/MonoGame.Extended/Math/Matrix3x2.cs#L301

```csharp
public void Decompose(out Vector2 translation, out float rotation, out Vector2 scale)
{
    translation.X = M31;
    translation.Y = M31;
....
```

Std-Enigma:
fixes #935


AristurtleDev (approved):
Thanks @Std-Enigma , was indeed a typo

PR #945, opened 2024-09-02
Fix bug when using `NinePatch` with `Texture2DRegion`.

### LINKED ISSUE #943: NinePatch does not work correctly with arbitrary Texture2DRegion patches
If the `Mid` region is offset from the top-left of the image by more than the top and left sides of the patch, for example by using middle images from a spritesheet, then the `Padding` calculation will be wrong, and will instead take the position of the `Mid` region within the spritesheet as being the padding from the top-left.

I'll submit a PR to fix this in a bit.

Dwergi:
Resolves #943

### DIFF
--- a/source/MonoGame.Extended/Graphics/NinePatch.cs
+++ b/source/MonoGame.Extended/Graphics/NinePatch.cs
@@ -62,6 +62,9 @@ public class NinePatch
     /// </summary>
     public string Name { get; }
 
+    /// <summary>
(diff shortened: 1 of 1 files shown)


AristurtleDev (approved):
Thanks @Dwergi. I don’t particularly like forcing the user to know that each of the edges must be the same, even if it’s shown in the documentation comments.

This might be something that needs to be reworked from the original code in the future.

PR #946, opened 2024-09-02
`BitmapFont` now always loads from `TitleContainer`.

### LINKED ISSUE #944: BitmapFont.FromStream does not use TitleContainer and breaks KNI web builds
KNI web builds do not support using `File.OpenRead`, but rather require `TitleContainer.OpenStream`. `BitmapFont` does not take this into account. 

I'll have a PR to fix this soon.

Unfortunately, this requires deprecating the `BitampFont.Read(FileStream stream)` method, because that uses the `FileStream.Name` field, which does not belong the `Stream` API.

Dwergi:
Fixes KNI web builds when using `BitmapFont`.

Resolves #944


AristurtleDev (approved):
Thanks @Dwergi

PR #948, opened 2024-09-03
Fix bug when creating a NinePatch using the Texture2DRegion extension method

### LINKED ISSUE #947: Texture2DRegionExtensions.CreateNinePatch creates rectangles that overlap with the middle section
When creating a NinePatch object using the mentioned extension method, the created texture regions are offset into the middle. This is because the patches use the middle section's width/height to calculate the x and y coordinates of the rightmost/bottom-most texture regions. The X should be based on the `image's width - right padding` and the Y should be based on `image's height - bottom padding` instead.

Separate from #943, as that deals with creating the rectangles for rendering.

greenstack:
This PR fixes the creation of a NinePatch object using the `Texture2DRegionExtensions.CreateNinePatch` method. This will resolve #947.


AristurtleDev (approved):
Thanks @greenstack

PR #949, opened 2024-09-17
Add Infinity methods

fixin0:
Infinity methods in float prepared for use in Ray and Vector methods

### DIFF
--- a/source/MonoGame.Extended/Math/Infinity.cs
+++ b/source/MonoGame.Extended/Math/Infinity.cs
@@ -10,13 +10,13 @@ namespace MonoGame.Extended
     public static float Infinity()
     { 
 
-        return BitConverter.ToSingle(BitConverter.GetBytes(0x7E80000), 0);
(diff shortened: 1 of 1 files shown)


AristurtleDev:
@fixin0 Is there a reason we would need these? The .NET BCL already provides these through [`float.PositiveInfinity`](https://learn.microsoft.com/en-us/dotnet/api/system.single.positiveinfinity?view=net-8.0)  and [`float.NegativeInfinity`](https://learn.microsoft.com/en-us/dotnet/api/system.single.negativeinfinity?view=net-8.0)

The IEEE 754 also defines positive infinity to be `0x0x7F800000`, which your implementation is missing the last `0`.

fixin0:
Actually, I was working on something different for this repo, but I couldn't work on it because of my work, so I sent the PR that I currently apply, yes you are right, it is unnecessary with these conditions. I am closing the Pull Request.

PR #961, opened 2024-11-06
Unify and improve FromHex in ColorHelper.cs and ColorExtensions.cs

Ev3yr:
Updated ColorExtensions.FromHex to reference ColorHelper.FromHex rather than using it's own, different, implementation.

Updated ColorHelper.FromHex with a new implementation that makes use of a single uint parse of the hex code and bit shifts, rather than numerous string operations, which is approx. 4x faster. It now also supports conversions of 3-and-4 digit hex codes in the same manner as HTML color codes, where "ABC" would be interpreted as "AABBCC".

### DIFF
--- a/source/MonoGame.Extended/ColorExtensions.cs
+++ b/source/MonoGame.Extended/ColorExtensions.cs
@@ -11,11 +11,7 @@ namespace MonoGame.Extended
     {
         public static Color FromHex(string value)
         {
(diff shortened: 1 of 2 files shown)


AristurtleDev:
Thanks @Eb3yr.

Side note: Looking at it, I'm not sure why there are two separate classes related to color utilities. This is something that needs to be looked into and cleaned up as well.

Going to open an issue for it to track that.

Issue #960, opened 2024-10-28
Matrix 3x2 Invert doesn't work properly


optsing:
Matrix3x2 Invert produces an incorrect result because each next element is computed from the previously modified elements, not the original one.

As a solution, you can create a new matrix instead of mutating the original one or use temporary variables.

Also, Matrix3x2 contains a division operator that simply divides each element of one matrix by an element of the other, which makes no practical sense, better to replace it with a multiplication by an inverted matrix, so A * B / B = A

AristurtleDev:
Resolved in #962

PR #962, opened 2024-11-08
Fix Invert(ref Matrix3x2)

Ev3yr:
Previously Matrix3x2 sequentially mutated the fields of the input ref matrix, so later elements would be computed from modified elements. Fixed by constructing and assigning a new struct to the input matrix.

Additionally added a check for the determinant equalling zero, which would set the reciprocal multiplier to infinity. Returns the Identity matrix in this case, as MM⁻¹=I for non-singular matrices. M⁻¹=I in the singular case would uphold this.

Resolves issue #960

### DIFF
--- a/source/MonoGame.Extended/Math/Matrix3x2.cs
+++ b/source/MonoGame.Extended/Math/Matrix3x2.cs
@@ -562,14 +562,21 @@ public struct Matrix3x2 : IEquatable<Matrix3x2>
     {
         var det = 1.0f / matrix.Determinant();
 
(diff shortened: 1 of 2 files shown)


AristurtleDev (commented):
Thanks @Eb3yr 
Would it be possible to add a unit test for this change before merging?

Ev3yr:
Added a test for two matrices with a positive and negative determinant, and for a matrix with a determinant of zero.

AristurtleDev (approved):

PR #964, opened 2024-11-18
Add compile time exclusion

kaltinril:
This fixes an issue where because Monogame now comes with this extension method, it fails if someone has Monogame and Extended functionality mixed on in their ContentTypeReader class.

https://github.com/MonoGame/MonoGame/blob/8df1e3edb3076b010822d2c9cf019f2f7f5f70a0/MonoGame.Framework/Content/ContentReaderExtensions.cs#L24

Example code that will break, can't have both of these using statements.  To fix I've simply made it so Monogame.Extended doesn't add these methods during compile time.

```csharp
using Microsoft.Xna.Framework.Content;
using MonoGame.Extended.Content;

namespace ContentManager_Extensions
{
    public class ReaderExample : ContentTypeReader<ReaderExampleClass>
    {
        protected override ReaderExampleClass Read(ContentReader reader, ReaderExampleClass existingInstance)
        {
            var graphicsDevice = reader.GetGraphicsDevice();

            return new ReaderExampleClass("exampletexture.png");
        }
    }

    public class ReaderExampleClass
    {
        public string TextureName { get; set; }

        public ReaderExampleClass(string textureName)
        {
            this.TextureName = textureName;
        }
    }
}
```

image

### DIFF
--- a/source/MonoGame.Extended/Content/ContentReaderExtensions.cs
+++ b/source/MonoGame.Extended/Content/ContentReaderExtensions.cs
@@ -9,12 +9,14 @@ namespace MonoGame.Extended.Content
 {
     public static class ContentReaderExtensions
     {
+#if FNA || KNI
         private static readonly FieldInfo _contentReaderGraphicsDeviceFieldInfo = typeof(ContentReader).GetTypeInfo().GetDeclaredField("graphicsDevice");
 
         public static GraphicsDevice GetGraphicsDevice(this ContentReader contentReader)
         {
             return (GraphicsDevice)_contentReaderGraphicsDeviceFieldInfo.GetValue(contentReader);
         }
+#endif
 
         public static string RemoveExtension(string path)
         {
--- a/source/MonoGame.Extended/Content/ContentReaderExtensions.cs
+++ b/source/MonoGame.Extended/Content/ContentReaderExtensions.cs
@@ -9,13 +9,20 @@ namespace MonoGame.Extended.Content
 {
     public static class ContentReaderExtensions
     {
-#if FNA || KNI
-        private static readonly FieldInfo _contentReaderGraphicsDeviceFieldInfo = typeof(ContentReader).GetTypeInfo().GetDeclaredField("graphicsDevice");
+		
+#if FNA
+		public static GraphicsDevice GetGraphicsDevice(this ContentReader contentReader)
+		{
+			var serviceProvider = contentReader.ContentManager.ServiceProvider;
+			var graphicsDeviceService = serviceProvider.GetService(typeof(IGraphicsDeviceService)) as IGraphicsDeviceService;
+				   
+			if (graphicsDeviceService == null)
+			{
+				throw new InvalidOperationException("No Graphics Device Service");
+			}
 
-        public static GraphicsDevice GetGraphicsDevice(this ContentReader contentReader)
-        {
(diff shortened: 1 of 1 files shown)


nkast:
The extension exist on KNI as well. 
https://github.com/kniEngine/kni/blob/02e6accab0615535b6adb15075b013a28ada9a58/src/Xna.Framework.Graphics/Graphics/Utilities/ContentReaderExtensions.cs#L33-L41
So if you include it, it will have the same conflict as you get in MonoGame.

The extension will not work on FNA because ContentReader  doesn't have a graphicsDevice field.
https://github.com/FNA-XNA/FNA/blob/master/src/Content/ContentReader.cs
The reflection will no longer work on MonoGame/KNI either. The field was removed in https://github.com/MonoGame/MonoGame/pull/6913

 The proper way to get a graphicsDevice inside a ContentTypeReader, is through the ServiceProvider.
 ```
   var graphicsDeviceService = contentReader.ContentManager.GetService(typeof(IGraphicsDeviceService)) as IGraphicsDeviceService;
   var graphicsDevice = graphicsDeviceService.GraphicsDevice;
 ```

kaltinril:
Interesting, I'm not trying to add it, I'm trying to exclude it.

It sounds like you're saying it's either in or won't work in eveything (fna, kni, monogame). 

If thats true, I say it should be removed from extended.

This method extension is in monogame.

I don't know if "monogame/kni" is different than "kni". So I'm not sure what you mean in your last statement, as monogame does have this method.

nkast:
You are excluding it from monogame because of a conflct. The same method exist on KNI, so you have to exclude it from KNI as well.

Which leaves FNA, where the reflection will fail.

I think it's ok to remove it, as it didn't work for some time now. Unless it's used from Extended internally.

AristurtleDev:
I could be adjusted to remove reflection so it's compatible with FNA.  This is also a step that will need to be looked at in the future to ensure MGE is AoT compatible once MonoGame removes the BRUTE dependency.

So we would need to adjust it to the service pattern like MonoGame and KNI has implemented like suggested by @nkast , but only for FNA

The `FieldInfo _contentReaderGraphicsDeviceFieldInfo` isn't used anywhere except in that method as a static cache after the first time it's called, so that would need to be removed, then the method updated to

```cs
#if FNA
public static GraphicsDevice GetGraphicsDevice(this ContentReader contentReader)
{
    var serviceProvider = contentReader.ContentManager.ServiceProvider;
    var graphicsDeviceService = serviceProvider.GetService(typeof(IGraphicsDeviceService)) as IGraphicsDeviceService;
           
    if (graphicsDeviceService == null)
    {
        throw new InvalidOperationException("No Graphics Device Service");
    }

    return graphicsDeviceService.GraphicsDevice;
}
#endif
```

Since FNA's goal is to remain 1:1 compatible with the XNA Api, i don't see them adding this extension method in.

kaltinril:
Sounds good @nkast  and @AristurtleDev.  This small change spawned because I was trying to validate and update the documentation for ContentManager Extensions in Monogame.Extended .

https://www.monogameextended.net/docs/features/contentmanager/contentManager-extensions/

Please let me know if there are any other changes you wish, I'm going to notate the section in the Extended docs to indicate that it only applies for FNI.

nkast:
Looks good. Thanks.

AristurtleDev:
Thanks @kaltinril

Issue #929, opened 2024-08-06
AnimatedSprite displays first region of texture atlas before animation plays


aconstas:
Link to docs: https://www.monogameextended.net/docs/features/2d-animations/animatedsprite/

I was following the documentation but I'm seeing this issue where the first region of the atlas displays before the desired animation.  In my code I've defined a sleep animation that does play, but before it plays the first region of the texture atlas (atlas) displays before.

@AristurtleDev: "The problem is how the AnimatedSprite(SpriteSheet, String) constructor is done when you call it by doing _dog = new AnimatedSprite(_spriteSheet, "sleep").

That constructor takes the _spriteSheet and calls the other constructor to set that up. Since AnimatedSprite is derived from Sprite, then it has to call the base constructor and pass in a texture to use, which just defaults to the first index in the texture atlas.

For right now, you should be able to fix this by instead doing the following workaround until a fix is pushed"
```csharp
_dog = new AnimatedSprite(_spriteSheet);
_dog.SetAnimation("sleep");
```

dev-ISO:
I ran into the same issue where the first region of the Texture2DAtlas was displaying before the desired animation and even with your workaround I still was running into this issue. I ended up having to fork and modify the source code to implement a working fix.

What I did to temporarily fix this was adding a custom method to AnimatedSprite
`public AnimatedSprite(SpriteSheet spriteSheet, int textureAtlasRegionIndex)
    : base(spriteSheet.TextureAtlas[textureAtlasRegionIndex])
{
    ArgumentNullException.ThrowIfNull(spriteSheet);
    _spriteSheet = spriteSheet;
}`

and adding two custom methods to SpriteSheet that I use to try and get the Index of the First Texture2DRegion for my targeted Animation Frame:
`public bool TryGetAnimation(string name, out SpriteSheetAnimation animation)
{
    return _animations.TryGetValue(name, out animation);
}

public void GetAnimationFirstRegion(string name, out Texture2DRegion firstTextureRegion)
{
    TryGetAnimation(name, out SpriteSheetAnimation animation);
    var firstFrame = animation.Frames[0];
    int frameIndex = firstFrame.FrameIndex;
    firstTextureRegion = TextureAtlas.GetRegion(frameIndex);
}`

[...]

PR #972, opened 2025-01-11
Fix incorrect first animation frame

DavidFidge:
Fix for first animation frame always showing the first sprite sheet texture region rather than the first animation's texture region - issue #929


DavidFidge:
If you want to see the issue before taking the fix, a demo project is here:

https://github.com/DavidFidge/MonoGameExtendedAnimatedSpriteFix

(replace MonoGame.Extended project reference to point to your local source)

AristurtleDev (approved):
Thanks @DavidFidge

Issue #933, opened 2024-08-18
First Frame Duration being applied before first frame is shown


dev-ISO:
The issue occurs when defining an animation where the first frame has a duration of a few seconds. The AnimationController applies this duration before the first frame is even displayed, causing a noticeable lag when switching from another animation, like "WalkLeft," to the "Idle" animation. This delay results in the first frame of the "Idle" animation not appearing immediately, creating an undesirable pause during the transition.

```csharp
//Define the "WalkLeft" animation
_spriteSheet.DefineAnimation("WalkLeft", builder =>
{
    builder.IsLooping(true)
        .AddFrame("WalkLeft1", TimeSpan.FromSeconds(0.2))
        .AddFrame("WalkLeft2", TimeSpan.FromSeconds(0.2))
        .AddFrame("WalkLeft3", TimeSpan.FromSeconds(0.2))
        .AddFrame("WalkLeft4", TimeSpan.FromSeconds(0.2));
});

// Define the "Idle" animation
_spriteSheet.DefineAnimation("Idle", builder =>
{
    builder.IsLooping(true)
        .AddFrame("Idle1", TimeSpan.FromSeconds(4))
        .AddFrame("Idle2", TimeSpan.FromSeconds(0.25));
});
```
I believe the issue stems from the AnimationController's Play() Method 
```csharp
public bool Play(int startingFrame)
{
    if (startingFrame < 0 || startingFrame >= _definition.FrameCount)
    {
        throw new ArgumentOutOfRangeException(nameof(startingFrame), $"{nameof(startingFrame)} cannot be less than zero or greater than or equal to the total number of frames in this {nameof(AnimationController)}");
    }

    //  Cannot play something that is already playing
    if (IsAnimating)
    {
        return false;
    }

    IsAnimating = true;
    _internalFrame = startingFrame;
    CurrentFrameTimeRemaining = _definition.Frames[_internalFrame].Duration;
    return true;
}
```

Because the CurrentFrameTimeRemaining is applied immediately the first frame's duration is being applied before the animation is shown
```csharp
CurrentFrameTimeRemaining = _definition.Frames[_internalFrame].Duration;
```

To fix this issue I commented out the CurrentFrameTimeRemaining line and set the CurrentFrameTimeRemaining to a TimeSpan of 0 so that the delay happens only after the first frame is rendered to the screen.
```csharp
//CurrentFrameTimeRemaining = _definition.Frames[_internalFrame].Duration;
[...]

PR #976, opened 2025-03-01
Set initial frame when animation is changed

wolf81:
I wanted to have single-frame animations (just to keep the logic simple in my game). E.g. while a move state might have multiple frames, an idle state just shows a single frame.

However, I noticed in such cases the animation would not be updated. Initially, I thought I did something wrong but couldn't figure it out. Then I looked into the codebase and noticed a small issue. When an animation is set, the initial frame is not yet configured. It will only change after an update, which would only happen for multi-frame animations.

This small change ensures the initial frame is configured as soon as the animation is set. This way, the correct sprite is also immediately shown for the new animation.

Perhaps these issues are related:
 
- https://github.com/craftworkgames/MonoGame.Extended/issues/933
- https://github.com/craftworkgames/MonoGame.Extended/issues/929


AristurtleDev (approved):
Thansk @wolf81

PR #978, opened 2025-03-19
Change CollisionEventArgs settter to init require

### LINKED ISSUE #977: Change CollisionEventArgs setters to init require
I propose to change the properties setters from `internal` to `init`. Optionally, I'd also add `required`

Original:
```C#
    public class CollisionEventArgs : EventArgs
    {
        public ICollisionActor Other { get; internal set; }
        public Vector2 PenetrationVector { get; internal set; }
    }
```

Change:
```C#
    public class CollisionEventArgs : EventArgs
    {
        public required ICollisionActor Other { get; init; }
        public required Vector2 PenetrationVector { get; init; }
    }
```

Reasoning:

I believe this is how it was supposed to be implemented according to the documentation. Since `Other` and `PenetrationVector` are not marked as `nullable`, we'd never expect them to be `null`. By forcing `required` `init`, we force users of this class to always set these values.

A second advantage is that it allows you to extend this class in a sane way. `internal` setters were forcing us to implement custom properties for custom collisions that would otherwise be handled the same way as any other collision, simply because we can't set the properties due to the accessibility modifier. Making the setters as `init` and leaving them `public` means that we can inherit from this class and set those properties during construction, while maintaining the current contract of not allowing consumers to further change these properties.

AlexMacocian:
Closes #977

### DIFF
--- a/source/MonoGame.Extended/Collisions/CollisionEventArgs.cs
+++ b/source/MonoGame.Extended/Collisions/CollisionEventArgs.cs
@@ -12,7 +12,7 @@ namespace MonoGame.Extended.Collisions
         /// <summary>
         /// Gets the object being collided with.
         /// </summary>
-        public ICollisionActor Other { get; internal set; }
+        public required ICollisionActor Other { get; init; }
 
         /// <summary>
         /// Gets a vector representing the overlap between the two objects.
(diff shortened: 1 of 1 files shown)


AristurtleDev (approved):

AristurtleDev:
Thanks @AlexMacocian

Issue #994, opened 2025-06-06
ClipRectangles does not work when clipping textures from the top down


BradBergstrom:
I noticed that when BitmapFontExtensions.DrawString and SpriteBatchExtensions.Draw and passing in a ClippingRectangle parameter, there are unexpected artifacts that are rendered when the ClippingRectangle passed in is clipping the BitmapFont or Texture from the Top. 

I don't notice any issues when the ClippingRectangle intersects the BitmapFont/Texture from the bottom. Here is a screenshot. 

![Image](https://github.com/user-attachments/assets/baea030b-741a-405b-a4d4-7d2e7bebb7d4)

BradBergstrom:
I have pushed a pull request that has a fix for this: https://github.com/craftworkgames/MonoGame.Extended/pull/995

PR #995, opened 2025-06-06
Fix Issue in ClipRectangles from SpriteBatchExtensions.cs

BradBergstrom:
Fix for issue described in: https://github.com/craftworkgames/MonoGame.Extended/issues/994

### DIFF
--- a/source/MonoGame.Extended/Graphics/SpriteBatch.Extensions.cs
+++ b/source/MonoGame.Extended/Graphics/SpriteBatch.Extensions.cs
@@ -3,6 +3,7 @@
 // See LICENSE file in the project root for full license information.
 
 using System;
+using System.Drawing;
 using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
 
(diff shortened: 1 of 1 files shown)


BradBergstrom:
After further testing, it appears this current solution only works when Scale = 1.

BradBergstrom:
After further testing, this solution only works when Scale == 1. Otherwise the results are still unexpected when trying to render BitmapFonts. I will open a new pull request when I get this solved.

BradBergstrom:
After testing this for 7+ hours, the solution I have presented is still the best one, as it fixes issues that the current ClipRectangles function has when clipping Textures/BitmapFonts from the top down. However, there are major drawbacks to doing this kind of CPU bound clipping. Both my formula, and the original formula do not work 100% when the scale is anything other than 1. 

The position needs to cleanly divide by the scale in order for the ClipRectangles function to put the object in the right place on the screen. This completely falls apart when you try lerping an object into a position. The need to find the scale by dividing causes funny rounding errors all over the place, resulting in the lerp not looking smooth and the Texture/BitmapFont being shifted up or down by a few pixels. 

There really isn't a way to fix this, the only way to make this function work properly at all scales would be to return the DestinationRectangle and ClippingRectangle to its original 1x scale, then do the math, then scale it back up. But again, that only works so long as the DistinationRectangle and ClippingRectangle is cleanly divisible by the scale. 

So from my point of view, there really isn't a way to fix this 100%. I'm going to switch over to using ScissorRectangles for my Graphical User Interface code that I'm writing. But I'm going to leave the pull request up because I do think it's an improvement.

AristurtleDev:
Thanks for the time put into this @BradBergstrom.  I'm currently focused on the particle system as it's a priority for a user and getting the new particle editor out with it.  

I'll leave this PR open as well, but will convert it to a draft until such time that either you, I, or someone can revisit it.

PR #998, opened 2025-07-22
fixed broken Rectangle.Clip() extensions method

CodeAndWeb:
rectangle.Right/Bottom depends on rectangle.X/Y, which has been changed two lines before;
unit test will fail with old implementation


AristurtleDev (approved):
Thanks @CodeAndWeb

PR #1002, opened 2025-07-25
Removed boxing in TweenMemberss getters and setters

DiegoG1019:
## Description

* Usually, in both tween member classes, both the runtime compiled setter and the getter box their values, despite being constrained to structs
* This PR fixes this by adjusting the typing of the delegates used, as well as modifying the expressions that are compiled so they are assignable to the delegates.
* Finally, it does so in a way that requires no boxing.

Testing of this should be no different from the tests that currently exists, and I tested this for a couple hours with a couple different tweens and it seems to both work and show up far less allocations on my end

I hope it's useful!

## Related Issues/Tickets

* I couldn't find any issues pertaining this, I probably should have opened one before delving into this, my bad.

## Changes Made

* Changed the typing of the TweenMember's delegates to be `Func<object, T>`
* Changed the compiled expressions to instead expect a value (known to be a struct) of the property/field's type, and to convert the instance into the property/field's declaring type before accessing.

## Checklist

### DIFF
--- a/source/MonoGame.Extended/Tweening/TweenFieldMember.cs
+++ b/source/MonoGame.Extended/Tweening/TweenFieldMember.cs
@@ -2,7 +2,6 @@ using System;
 using System.Diagnostics;
 using System.Linq.Expressions;
 using System.Reflection;
-using Microsoft.VisualBasic;
 
 namespace MonoGame.Extended.Tweening
 {
--- a/source/MonoGame.Extended/Tweening/TweenFieldMember.cs
+++ b/source/MonoGame.Extended/Tweening/TweenFieldMember.cs
@@ -2,40 +2,43 @@ using System;
 using System.Diagnostics;
 using System.Linq.Expressions;
 using System.Reflection;
+using Microsoft.VisualBasic;
 
 namespace MonoGame.Extended.Tweening
 {
     public sealed class TweenFieldMember<T> : TweenMember<T>
-        where T : struct 
+        where T : struct
     {
         private readonly FieldInfo _fieldInfo;
 
-        public TweenFieldMember(object target, FieldInfo fieldInfo) 
+        public TweenFieldMember(object target, FieldInfo fieldInfo)
             : base(target, CompileGetMethod(fieldInfo), CompileSetMethod(fieldInfo))
         {
             _fieldInfo = fieldInfo;
         }
 
-        private static Func<object, object> CompileGetMethod(FieldInfo fieldInfo)
+        private static Func<object, T> CompileGetMethod(FieldInfo fieldInfo)
         {
-            var self = Expression.Parameter(typeof(object));
-            var instance = Expression.Convert(self, fieldInfo.DeclaringType);
-            var field = Expression.Field(instance, fieldInfo);
-            var convert = Expression.TypeAs(field, typeof(object));
-
-            return Expression.Lambda<Func<object, object>>(convert, self).Compile();
+            var entityType = fieldInfo.DeclaringType!;
+            var parameter = Expression.Parameter(typeof(object), "entity");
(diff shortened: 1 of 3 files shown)


source/MonoGame.Extended/Tweening/TweenMember.cs
```diff
@@ -1,6 +1,8 @@
 using System;
 using System.Linq.Expressions;
 
+// TODO: THIS BITCH RIGHT HERE DOES A LOT OF BOXING. FIX IT!
```
AristurtleDev: I lol'd but please remove this comment to keep the code base professional.

source/MonoGame.Extended/Tweening/TweenFieldMember.cs
```diff
@@ -2,43 +2,46 @@
 using System.Diagnostics;
 using System.Linq.Expressions;
 using System.Reflection;
+using Microsoft.VisualBasic;
```
AristurtleDev, line 5: This may have been added by accident, don't think it's needed

AristurtleDev (changes requested):

source/MonoGame.Extended/Tweening/TweenMember.cs
DiegoG1019: I'm so sorry. I thought I had removed it before submitting!

source/MonoGame.Extended/Tweening/TweenFieldMember.cs
DiegoG1019: Yes, I missed that -- It's usually just me hitting tab by accident and my IDE promptly completing a type I don't need and including its namespace

AristurtleDev:
Thanks for the contribution on this @DiegoG1019.

A few follow ups on the PR if you want to add them in or try them out.

In **TweenPropertyMember.cs**

```cs
private static Func<object, T> CompileGetMethod(PropertyInfo propertyInfo)
{
    var entityType = propertyInfo.DeclaringType!;
    var parameter = Expression.Parameter(typeof(object), "entity");
    var property = Expression.Property(Expression.Convert(parameter, entityType), propertyInfo);
    var funcType = typeof(Func<,>).MakeGenericType(typeof(object), propertyInfo.PropertyType);
    return (Func<object, T>)Expression.Lambda(funcType, property, parameter).Compile();
}
```

This could probably be simplified now that we are using `T` generics and not boxing with `object`.  Since `T` is known at compile time for each generic, we should be able to just use the generic `Expression.Lambda<T>()` method instead of constructing the delegate type.  Something like this

```cs
private static Func<object, T> CompileGetMethod(PropertyInfo propertyInfo)
{
    var entityType = propertyInfo.DeclaringType!;
    var parameter = Expression.Parameter(typeof(object), "entity");
    var property = Expression.Property(Expression.Convert(parameter, entityType), propertyInfo);
    return Expression.Lambda<Func<object, T>>(property, parameter).Compile();
}
```

This would remove the need for the `MakeGenericType()` reflection call, removes the explict cast in the `return` statement that won't be needed now, and is more readable.

Might be able to do the same in **TweenFieldMember.cs**

DiegoG1019:
Just added in your suggestions!

source/MonoGame.Extended/Tweening/TweenFieldMember.cs
AristurtleDev: Not sure why, but this is still showing up. I can see in your commit [d292bb0](https://github.com/MonoGame-Extended/Monogame-Extended/pull/1002/commits/d292bb030fd37361cc1709917d063181b4ffc8c3) that it was removed.  Maybe GitHub is just lagging, going to approve it

AristurtleDev (approved):

AristurtleDev:
Thanks again for the fixes and performance improvements on this.

PR #1004, opened 2025-07-28
Texture2DRegion: support trimmed and rotated sprites optionally load origin from atlas data file

CodeAndWeb:
## Description

To improve sprite sheet packing efficiency the following features have been added to Texture2DRegion :
- Regions on the texture can now be rotated.
- Transparent borders around sprites can be trimmed.

The SpriteBatch.Draw method will automatically rotate the region back to its original orientation and restore the transparent borders during rendering.

Additionally, since sprite sheet data files can define sprite origins, TextureRegion2D can now optionally store this origin. Sprites created from a Texture2DRegion will use this stored origin as their default.

## Related Issues/Tickets

Discussion on Discord

## Changes Made

- Extended Texture2DRegion, adapted Texture2DAtlas, Sprite and SpriteBatch
- Adapted nine-patch sprite implementation for support rotated / trimmed sprites, too
- Extended TextureAtlas content pipeline to support trimmed and rotated sprites, as well as reading anchor points from the data file. The JSON format has been cleaned up, unnecessary keys have been removed. The old format still remains supported.

## Checklist

Note: We found no way to test the SpriteBatch.Draw() calls automatically. A test app which renders trimmed, rotated sprites is available here: [https://github.com/CodeAndWeb/MonoGame.Extended-TestApp](https://github.com/CodeAndWeb/MonoGame.Extended-TestApp)


source/MonoGame.Extended.Content.Pipeline/TextureAtlases/TexturePackerJsonImporter.cs
```diff
@@ -13,7 +13,20 @@ public class TexturePackerJsonImporter : ContentImporter<TexturePackerFileConten
...
+            if (tpFile.Meta.Image != null)
+            {
+                context.AddDependency(tpFile.Meta.Image);
+            }
+            else if (tpFile.Meta.DataFormat == "monogame-extended")
+            {
+                // new format: textures are in the textures array
+                foreach (var texture in tpFile.Textures)
+                {
+                    context.AddDependency(texture.FileName);
```
AristurtleDev, line 26: **Note on `AddDependency` behavior:** There's a common misconception about MonoGame's `AddDependency` method. While the name suggests it enforces required dependencies, it actually only creates **change tracking hints** for incremental builds. If a dependency file doesn't exist, `AddDependency` silently ignores it rather than failing the build.

> [!NOTE]
> After reviewing this PR, i opened a discussion in the MonoGame repository about this.  The term `AddDependency` is confusing in what it actually does.  https://github.com/MonoGame/MonoGame/issues/8943

## Validation-Only Approach

To **validate** that dependencies exist at build time without creating additional outputs, use `BuildAndLoadAsset` in the processor:

```cs
public override TexturePackerProcessorResult Process(TexturePackerFileContent input, ContentProcessorContext context)
{
    if (input.Meta.Image != null)
    {
        // Validates the texture exists and can be processed (fails build if missing)
        var externalRef = new ExternalReference<Texture2DContent>(input.Meta.Image);
        context.BuildAndLoadAsset<Texture2DContent, Texture2DContent>(externalRef, "TextureProcessor");
    }

    return new TexturePackerProcessorResult(input);
}
```

> [!IMPORTANT]
> This validates dependencies but doesn't create separate .xnb files, instead it returns the `Texture2DContent` item, basically loading the build into memory.  For runtime loading, texture files must still be added to your .mgcb project as individual content items.

## Build and Output Approach

Alternatively, you can use `BuildAsset` to both validate and create .xnb output files:

```cs
public override TexturePackerProcessorResult Process(TexturePackerFileContent input, ContentProcessorContext context)
{
    if (input.Meta.Image != null)
    {
        var externalRef = new ExternalReference<Texture2DContent>(input.Meta.Image);
        context.BuildAsset<Texture2DContent, Texture2DContent>(externalRef, nameof(TextureProcessor));
    }
    else if (input.Meta.DataFormat == "monogame-extended")
    {
        foreach (var texture in input.Textures)
        {
            var externalRef = new ExternalReference<Texture2DContent>(texture.FileName);
            context.BuildAsset<Texture2DContent, Texture2DContent>(externalRef, nameof(TextureProcessor));
        }
    }
    return new TexturePackerProcessorResult(input);
}
```

**Caveat**: MonoGame appends `_0` to the generated .xnb filenames to avoid collisions (e.g., `texture_0.xnb` instead of `texture.xnb`). Your runtime reader would need to account for this naming convention.

## Recommendation

Given MonoGame's limitations around dependency validation, there isn't a perfectly clean solution. Consider these options:

1. **Build-time validation (preferred)**: Use `BuildAndLoadAsset `for validation, but require developers to manually add textures to their .mgcb project
2. **Runtime validation (simpler)**: Use `try/catch` in your runtime reader with clear error messages when dependencies are missing
3. **Documentation**: Clearly document the dependency requirements and setup steps

Currently we use approach #1 in the Tiled processor, though the way Dylan originally set it up is a bit overkill, if you want to look at it as an example.  I think just usin the `BuildAndLoadAsset` for validation during build time would be preferred so that there isn't a clean build that then throws a runtime error later.

AristurtleDev (changes requested):
See note about validation for the importer.  Some method of validation that the image file is also added needs to be implemented.

CodeAndWeb:
Ok, we've implemented approach #1. If the image file is missing, the build fails with an error like this:

`demo/Content/spritesheet.json : error : The source file 'textures/spritesheet.png' does not exist!`

If the image file exists but has not been added to the mgcb file, we get an exception at runtime:

`Microsoft.Xna.Framework.Content.ContentLoadException: Could not load textures/spritesheet asset!`

AristurtleDev (approved):

PR #1006, opened 2025-08-04
feat(ComponentMapper): Add 'TryGet' methods

varelen:
## Description

I am writing a multiplayer Pong game as an example game for my network library and I have some places where I am not sure if an entity has some component. I could use 'ComponentMapper.Has' and 'ComponentMapper.Get', but these are two lookups and it's not that clean. Or use 'ComponentMapper.Get' with a null check.

I personally like the "Try" pattern for this and it is easier to read, so I made this PR.

Example of a network system which uses the new `TryGet` method:

```c#
    protected override void OnEntityRemoved(int entityId)
    {
        if (!this.networkMapper.TryGet(entityId, out NetworkComponent? networkComponent))
        {
            return;
        }

        var destroyEntityPacket = Packet.FromPool(MessageType.DestroyEntity);
        destroyEntityPacket.WriteInt(networkComponent.Id);

        this.server.BroadcastReliable(destroyEntityPacket);

        base.OnEntityRemoved(entityId);
    }
```

Feel free to reject the PR if these changes are not wanted/needed.

## Related Issues/Tickets

- no related issues/tickets

## Changes Made

- added `bool TryGet(Entity entity, [NotNullWhen(true)] out T result)` method to `ComponentMapper<T>`
- added `bool TryGet(int entityId, [NotNullWhen(true)] out T result)` method to `ComponentMapper<T>`
- added a unit test for it

## Checklist

* [X] I have verified that there are no existing pull requests that would overlap with this pull request.
* [X] I have verified that I am following the guidelines as outlined in this project's contribution policy
* [X] I have verified that this pull request adheres to this project's code of conduct.
* [X] I have written a descriptive title for this pull request.
* [X] I have provided appropriate test coverage were applicable.

### DIFF
--- a/source/MonoGame.Extended/ECS/ComponentMapper.cs
+++ b/source/MonoGame.Extended/ECS/ComponentMapper.cs
@@ -59,17 +59,8 @@ namespace MonoGame.Extended.ECS
 
         public bool TryGet(int entityId, [NotNullWhen(true)] out T result)
         {
-            var component = Get(entityId);
-
-            if (component == null)
-            {
-                result = default(T);
-                return false;
-            }
-
-            result = component;
-
-            return true;
+            result = Get(entityId);
+            return result != null;
         }
 
         public override bool Has(int entityId)
--- a/source/MonoGame.Extended/ECS/ComponentMapper.cs
+++ b/source/MonoGame.Extended/ECS/ComponentMapper.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics.CodeAnalysis;
 using MonoGame.Extended.Collections;
 
 namespace MonoGame.Extended.ECS
@@ -49,6 +50,26 @@ namespace MonoGame.Extended.ECS
         public T Get(int entityId)
         {
             return Components[entityId];
+        }
+
+        public bool TryGet(Entity entity, [NotNullWhen(true)] out T result)
+        {
+            return TryGet(entity.Id, out result);
+        }
+
+        public bool TryGet(int entityId, [NotNullWhen(true)] out T result)
+        {
+            var component = Get(entityId);
+
+            if (component == null)
+            {
+                result = default(T);
+                return false;
+            }
+
+            result = component;
+
+            return true;
         }
 
         public override bool Has(int entityId)
--- a/tests/MonoGame.Extended.Tests/ECS/ComponentMapperTests.cs
+++ b/tests/MonoGame.Extended.Tests/ECS/ComponentMapperTests.cs
@@ -46,6 +46,31 @@ namespace MonoGame.Extended.ECS.Tests
         }
 
         [Fact]
+        public void PutAndTryGetComponent()
+        {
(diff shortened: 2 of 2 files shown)


AristurtleDev:
Hey @varelen,  thanks for the PR. I wanted to discuss the addition of the TryGet pattern first before accepting this.

### Current Behavior

The `ComponentMapper.Get()` method delegates to the `Bag<T>` indexer:

https://github.com/MonoGame-Extended/Monogame-Extended/blob/de157c44e8e7f135cb3e60c871d365a71962296b/source/MonoGame.Extended/ECS/ComponentMapper.cs#L49-L52

https://github.com/MonoGame-Extended/Monogame-Extended/blob/de157c44e8e7f135cb3e60c871d365a71962296b/source/MonoGame.Extended/Collections/Bag.cs#L58-L60

This indexer already returns `default(T)` when the index is out of bounds, so `ComponentMapper.Get()` naturally returns `null` for invalid entity IDs (since `ComponentMapper<T>` is constrained to reference types).

### Thoughts on TryGet Implementation

The `TryGet` pattern is idiomatic in .NET (e.g. `TryGetValue`, `TryParse`, etc) and it's definitely more readable than `Get` then `null` check or doing `Has() + Get()` with two method calls.

However given the current behavior shown above, in that the `Bag<T>` indexer already returns `default(T)`, I think the method could be simplified as just:

```cs
public bool TryGet(int entityId, [NotNullWhen(true)] out T result)
{
    result = Get(entityId);
    return result == null;
}
```

(Also, `Entity` overload can call into the int version as proposed.)

Open to feedback, if this improves clarity for end users, I’m not opposed to including it.

varelen:
Hey @AristurtleDev, thank you for the suggestion. I have updated the 'TryGet' method to the simplified version.

By the way, isn't the out of bounds check in the `Has` method redundant, because it is also checked in `Bag`?

https://github.com/MonoGame-Extended/Monogame-Extended/blob/de157c44e8e7f135cb3e60c871d365a71962296b/source/MonoGame.Extended/ECS/ComponentMapper.cs#L54-L60

AristurtleDev:
> Hey @AristurtleDev, thank you for the suggestion. I have updated the 'TryGet' method to the simplified version.
> 
> By the way, isn't the out of bounds check in the `Has` method redundant, because it is also checked in `Bag`?
> 
> https://github.com/MonoGame-Extended/Monogame-Extended/blob/de157c44e8e7f135cb3e60c871d365a71962296b/source/MonoGame.Extended/ECS/ComponentMapper.cs#L54-L60

Good catch, yes, that would be redundant as well.  I'm guessing it was originally added as an early exit condition, but it's doing nothing different than what the actual indexer call is doing below it.

varelen:
Should or can I change it in this PR?

AristurtleDev:
> Should or can I change it in this PR?

Let's do this in a separate PR to keep the PR's focused since this one is just for adding the TryGet

PR #1167, opened 2026-07-14
Honor particle's initial opacity in OpacityFastFadeModifier

### LINKED ISSUE #1166: Ember OpacityFastFadeModifier should honor the particle's initial Opactiy
### Description

The OpacityFastFadeModifier in the particle implementation always starts from a hard-coded 1f instead of honoring the particle's initial opacity. 

### Steps to Reproduce

- Set Opacity on an ember particle to anything between 0f and 1f.
- Add an OpacityFastFadeModifier
- See the particle start at 1f Opacity

### Expected Behavior

OpacityFastFadeModifier should honor the initial Opacity and fade to 0f over he lifespan of the particle

### Actual Behavior

OpacityFastFadeModifier sets initial Opacity at 1f regardless of what it is set to

### Configuration

MGEX 6.0.1

### Other Information

_No response_

walteryoung:
## Description

OpacityFastFadeModifier previously always faded from a hard-coded 1.0 opacity regardless of the particle's release-time opacity. Particle now tracks InitialOpacity at release, and the modifier scales by it so particles released with Opacity < 1.0 fade from their actual starting value instead of jumping to full opacity.

## Related Issues/Tickets

Closes #1166

## Changes Made

Particles now track the InitialOpacity and use that value when activating the OpacityFastFadeModifier.

## Checklist

### DIFF
--- a/source/MonoGame.Extended/Particles/Data/Particle.cs
+++ b/source/MonoGame.Extended/Particles/Data/Particle.cs
@@ -57,6 +57,11 @@ public unsafe struct Particle
     public float Opacity;
 
     /// <summary>
+    /// The opacity (alpha) value assigned to this particle at release time.
+    /// </summary>
+    public float InitialOpacity;
+
+    /// <summary>
     /// The rotation of this particle in radians.
     /// </summary>
     public float Rotation;
--- a/source/MonoGame.Extended/Particles/Modifiers/OpacityFastFadeModifier.cs
+++ b/source/MonoGame.Extended/Particles/Modifiers/OpacityFastFadeModifier.cs
@@ -16,20 +16,19 @@ namespace MonoGame.Extended.Particles.Modifiers;
 /// Important notes:
 /// <list type="bullet">
 ///   <item>
-///     This modifier assumes particles have a standard lifespan of 1.0 second. For particles
-///     with different lifespans, the fade effect may not complete before the particle is removed,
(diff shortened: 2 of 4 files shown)


AristurtleDev:
My only concern here is that the `InitialOpacity` value is only applied during a `ParticleEmitter.Release`, which is how this is going to be used for the majority of the time.

However, users can still create and use a `ParticleBuffer` manually if they wanted to for more control or advanced scenarios.  In those situations, they would have to to write code similar to this

```
ParticleBuffer buffer = new ParticleBuffer(1);
ParticleIterator iterator = buffer.Release(1);
Particle* particle = iterator.Next();

particle->Age = 0.25f;
particle->Opacity = 0.5f;
// ...
```

The problem here is that they may do `partice->Opacity = ...` and think that's all that's needed.  If they do not also set the `particle-InitialOpacity` here then it will receive whatever stale value is in the buffer since its a ring buffer. No other fields on the particle start as a copy or mirror of an existing field like this, so it could be easily overlooked that they need to also set it. For instance, they may see the name `InitialOpacity` and think that it internally gets set some how.

Should probably add some documentation somewhere about this so it doesn't become a point of confusion in the future; that `particle->Opacity` and `particle-InitialOpacity` are two separate fields, one is a state, and one is a cached value

I'm going to merge for now and open an issue about how best to document this or make it obvious

AristurtleDev (approved):

PR #1178, opened 2026-07-23
Release/6.1.0 monogame 3.8.5

AristurtleDev:
## Description

Updates for MonoGame 3.8.5

## Checklist


AristurtleDev:
Shaders were originally compiled using mgfx 3.8.4, now are using 3.8.5 with the MonoGame update.

These shaders are failing in the tests for the PR due to KNI tests. They are stating

```
System.Exception : This effect seems to be for a newer version of KNI
```

@nkast i don't see a new package release in KNI for this, what would you recommend here? We could keep separate shader compiles, don't exactly like that, but wasn't sure if this was something you were aware of already in KNI

nkast:
The message is wrong and it's fixed for the next release.
https://github.com/kniEngine/kni/blob/31e620a3f22c5177fdb913935bc1807e22367f1f/src/Xna.Framework.Graphics/Graphics/Effect/Effect.cs#L218

What happened is that KNI moved to a [new effect format](https://blog.nkast.gr/post/what-s-new-in-kni-v4-02) in 4.02. It still can load MGFX v10 effects to help with easy migration from older KNI versions. Since MG moved to a newer MGFX v11, it can no longer load those shaders.

I recoment building the effect with KNIFX, because the support for MGFX v10 and the runtime patching of GLes shaders, are still there to allow transition from previous KNI versions, not cross compatibility with MonoGame.

nkast:
One recomentation is to remove reflection and instead check against  _graphicsDevice.Adapter.Backend_ at runtime.
See: https://github.com/nkast/Aether.Extras/blob/883bd789c3ac1f50370a5d5001ae7af9cfe3d5ca/Shaders/Tilemap/TilemapEffect.cs#L87

Note: In 4.02, GLes shaders are seperate from openGL shaders. You are going to need an additional embeded resource to support mobile and WebGL.

Another alternative is to use something I 've been expirimenting with in 4.02. _**Universal shaders**_.
To use universal shaders, you build a shader targeting multiple backends. You don't need any checks in your runtime. You have a single effectBytecode to load, and KNI does the rest.
See: https://github.com/kniEngine/kni/pull/2557/changes

You can generate them with something like:
`KNIFX MyEffect.fx MyEffect.fxo /Backend:DirectX11 /Backend:OpenGL /Backend:GLES`

PR #1188, opened 2026-07-28
Using `PlatformInfo.GraphicsBackend` for Platform Detection on MonoGame

### LINKED ISSUE #1187: Clean-up Platform Detection
Platform detection currently uses a bytecode test approach.
It should use `PlatformInfo.GraphicsBackend` instead.

uygary:
Also updating Directory.Build.props to refer to MonoGame.Framework.Native instead of MonoGame.Framework.DesktopGL.

## Description

* We now use `PlatformInfo.GraphicsBackend` to detect the graphics backend instead of testing possible bytecodes the backend.

## Related Issues/Tickets

* Closes #1187

## Changes Made

* Removed `EffectResource.DetermineShaderExtension()` and replaced the switch statement in `EffectResource.GetDefaultEffect()` with a switch on `PlatformInfo.GraphicsBackend`.
* Also updated `Directory.Build.props` to refer to the new core framework package instead of the DesktopGL package.

## Checklist

* [ ] I have provided appropriate test coverage were applicable.


source/MonoGame.Extended/Graphics/Effects/EffectResource.cs
```diff
@@ -25,86 +25,34 @@ namespace MonoGame.Extended.Graphics.Effects
     /// </remarks>
     public class EffectResource
     {
-        private static EffectResource _defaultEffectFna;
```
uygary, line 28: Was there an actual reason that I miss for us having these separate fields?

uygary: Just pinging @AristurtleDev.

nkast: The only reason I can think of, is if it was possible to switch backends at runtime.
Non of the frameworks support that at the moment.

Actually , I don't know why the effects are cached in the first place. Isn't that a problem with overriding effect parameters when it returns the same effect instance?

AristurtleDev: I'll need to look into this a bit more, just noting I haven't forgotten about it

AristurtleDev: Okay sorry for the delay, looked into this. @nkast brought up two separate concerns so I'll address them both here.

On the first point, i agree the separate cached fields look unecessary in the current design. NKast is right, XNA-likes do no support switching backends at runtime. So we can bring this down to a single cached default resource, which would be much simpler. 

On the second point, overriding effect parameters shouldn't be an issue.  What's cached here is the `EffectResource` and it's byte code, not the actual `DefaultEffect`/`Effect` object.  The `DefaultEffect` constructor still creates its own `Effect` instance from that bytecode, so paramter changes on one should not affect other.

https://github.com/MonoGame-Extended/Monogame-Extended/blob/6658f4e18a66e4c154f6b63b9cf87b9d45b0f89b/source/MonoGame.Extended/Graphics/Effects/DefaultEffect.cs#L111-L119

AristurtleDev (approved):
Thanks @uygary

Issue #1170, opened 2026-07-16
Native AOT Compatibility


uygary:
**Goal**

Refactoring MGE to be Native AOT compatible.
This will be an umbrella issue to keep track of child issues that cover specific areas of MGE.

---
**Confirmed Areas**

- [x] Platform Detection: #1171
- [x] Tiled Tilemap Support: #1172
- [x] LDtk Tilemap Support: #1173
- [x] Ogmo Tilemap Support: #1174
- [ ] Json Serialization: #1198

**Outstanding Areas to Confirm**

- [ ] `BitMapFonts`
- [ ] `ContentReaders`
- [ ] `TexturePackers`
- [ ] `ECS`
- [ ] `InputListeners`
- [ ] `Tweening`
- [ ] ...*possibly more to come*...

PR #1189, opened 2026-08-07
Add NativeAOT registration for tilemap readers

tigurx:
## Description

MonoGame.Extended’s binary tilemap content readers rely on reflection-based reader resolution. In trimmed or NativeAOT builds, this can cause tilemaps to fail at runtime with:

`Could not find ContentTypeReader Type: MonoGame.Extended.Tilemaps.Content.TilemapReader`

This PR adds explicit registration methods following the existing `BitmapFontContentReader.Register()` pattern.

## Related Issues/Tickets

* Related to #1170 (Native AOT Compatibility)

## Changes Made

* Added `Register()` to `TilemapReader`
* Added `Register()` to `TilemapTilesetReader`
* Added `Register()` to `TilemapWorldReader`
* Used `ContentTypeReaderManager.AddTypeCreator(...)`

## Checklist

* [x] I have verified that I am following the guidelines as outlined in this project's contribution policy.
* [ ] I have provided appropriate test coverage where applicable.

I’ve verified the library builds successfully. Would maintainers prefer a dedicated test that loads minimal tilemap XNB fixtures after calling the new Register() methods, or is the implementation pattern and build validation sufficient here?

### DIFF
--- a/source/MonoGame.Extended/Tilemaps/Content/TilemapReader.cs
+++ b/source/MonoGame.Extended/Tilemaps/Content/TilemapReader.cs
@@ -8,6 +8,21 @@ namespace MonoGame.Extended.Tilemaps.Content;
 /// </summary>
 public sealed class TilemapReader : ContentTypeReader<Tilemap>
 {
+#if !FNA && !KNI
+    /// <summary>
+    /// Registers this <see cref="ContentTypeReader"/> with the <see cref="ContentTypeReaderManager"/>
+    /// so it is resolved without reflection.
+    /// </summary>
+    /// <remarks>
+    /// Call this method once during application startup when publishing with
+    /// <c>PublishAot</c> or <c>PublishTrimmed</c>.
+    /// </remarks>
+    public static void Register() =>
+        ContentTypeReaderManager.AddTypeCreator(
+            typeof(TilemapReader).AssemblyQualifiedName,
+            () => new TilemapReader());
+#endif
+
     /// <inheritdoc/>
     protected override Tilemap Read(ContentReader reader, Tilemap existingInstance)
     {
--- a/source/MonoGame.Extended/Tilemaps/Content/TilemapTilesetReader.cs
+++ b/source/MonoGame.Extended/Tilemaps/Content/TilemapTilesetReader.cs
@@ -11,6 +11,21 @@ namespace MonoGame.Extended.Tilemaps.Content
     /// </summary>
     public sealed class TilemapTilesetReader : ContentTypeReader<TilemapTileset>
     {
+#if !FNA && !KNI
+        /// <summary>
+        /// Registers this <see cref="ContentTypeReader"/> with the <see cref="ContentTypeReaderManager"/>
+        /// so it is resolved without reflection.
+        /// </summary>
+        /// <remarks>
+        /// Call this method once during application startup when publishing with
+        /// <c>PublishAot</c> or <c>PublishTrimmed</c>.
+        /// </remarks>
+        public static void Register() =>
+            ContentTypeReaderManager.AddTypeCreator(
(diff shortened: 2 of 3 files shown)


AristurtleDev (approved):

AristurtleDev:
Thanks @tigurx

source/MonoGame.Extended/Tilemaps/Content/TilemapReader.cs
```diff
@@ -8,6 +8,21 @@ namespace MonoGame.Extended.Tilemaps.Content;
...
+    /// Registers this <see cref="ContentTypeReader"/> with the <see cref="ContentTypeReaderManager"/>
+    /// so it is resolved without reflection.
+    /// </summary>
+    /// <remarks>
+    /// Call this method once during application startup when publishing with
+    /// <c>PublishAot</c> or <c>PublishTrimmed</c>.
+    /// </remarks>
+    public static void Register() =>
+        ContentTypeReaderManager.AddTypeCreator(
+            typeof(TilemapReader).AssemblyQualifiedName,
```
uygary, line 22: I have a silly question:
Did we actually **test** this and **confirm that it works** when the game is published with Native AOT?

The reason I'm asking is, these type registrations seem to have a standard that uses shorthand type names, not assembly qualified names.

For example when I register `TileMapReader` with assembly qualified name like this:
`MonoGame.Extended.Tilemaps.Content.TilemapReader, MonoGame.Extended, Version=6.1.0.0, Culture=neutral, PublicKeyToken=null`
It doesn't work. When I register it with a shortened name:
`MonoGame.Extended.Tilemaps.Content.TilemapReader, MonoGame.Extended`
This time it works. Which is why I have these registrations set up like this in my own code:

```cs
public static void AddTypeCreator<T>(Func<ContentTypeReader> createFunction) where T: ContentTypeReader
{
    var type = typeof(T);
    var assembly = type.Assembly;
    var typeName = type.FullName;
    var assemblyName = assembly.GetName().Name;
    var typeString = $@"{typeName}, {assemblyName}";
    ContentTypeReaderManager.AddTypeCreator(typeString, createFunction);
    
    Trace.TraceInformation($"Registered ContentReader for AOT: {typeString}");
}
```

So I don't think these registrations will actually work.

uygary: Yeah, this won't work.
```
ApocalypseNow Error: 0 : Fatal error in ApocalypseNow GL: System.NotSupportedException: It seems that you are using PublishAot and trying to load assets with a reflection-based serializer (which is not natively supported). To work around this error, call ContentTypeReaderManager.AddTypeCreator() in your Game constructor with the following type: MonoGame.Extended.Tilemaps.Content.TilemapReader, MonoGame.Extended
at Microsoft.Xna.Framework.Content.ContentTypeReaderManager.LoadAssetReaders(ContentReader) + 0x54c
...
at Entropik.Framework.MonoGame.Host.HostHelper.<RunGame>d__1`1.MoveNext() + 0x7ad
```
[See how the header is generated for the XNB](https://github.com/MonoGame-Extended/Monogame-Extended/blob/6658f4e18a66e4c154f6b63b9cf87b9d45b0f89b/source/MonoGame.Extended.Content.Pipeline/Tilemaps/TilemapWriter.cs#L33).

tigurx: Good catch, and thank you for following up with a fix! 

I initially believed my NativeAOT test was exercising the new Register() methods, but my game’s existing compatibility helper was still registering the shortened reader names. That masked the mismatch and explains why tilemap loading passed in my test. I appreciate you catching this.

